PR #1, opened 2021-04-08
Feature/initial walletconnect provider

kevinhcolgan:
Initial WalletConnect provider implementation.

The WalletConnect implementation doesn't extend the existing BaseWallet class as the paradigm for WalletConnect is a bit different: there is the concept of a requester and a signer. The library thus provides implementations for each of these use-cases. The detailed flows and code-snippets are available in packages/walletconnect/README.md


alice-the-sender:
Any plans to publish to npm?

PR #7, opened 2021-08-03
Solflare Web

vsakos:


jordansexton:
Very nice! When will the wallet extension be available so I can test this?

vsakos:
@jordansexton We decided to add only the web wallet (a `sol-wallet-adapter` compatible provider) for now and will create a separate pull request when the browser extension is approved in the Chrome Web Store.

PR #13, opened 2021-08-18
Move to `sendTransaction` interface for online signers

jordansexton:


yanglin5689446:
For Blocto integration, there're still some tricky parts: we convert a transaction from dApp in our end, and a transaction may be converted to a main tx and multiple appended txs.
Appended txs serve the following purposes:  1. create buffer  2. append data   3. execute. all these steps are dependent so we need to send them to cluster at once, the reason why we doing so is because of the transaction MTU limits. 
Due to this implementation, we really want to make sure we send the transaction in our end, we certainly can returns a bunch of pre-signed txs to dApp side and ask them to sign these txs, but it'll be very confusing.

Please refer to the following architecture diagram of Blocto:
solana-tx3

jordansexton:
@yanglin5689446 Thanks for this context.

Screen Shot 2021-08-18 at 12 29 30 PM

It seems like the first flow can be handled by the abstraction in this PR, without the dapp needing to know how Blocto's wallet works. We could have an adapter that wraps the Blocto SDK, the dapp will just blindly call `sendTransaction` on the WalletAdapter, and the adapter will call the SKD, which will go back and forth with your backend, and ultimately just return a transaction hash.

Screen Shot 2021-08-18 at 12 31 02 PM

What does the second flow require them to partially sign with? How would a dapp know this was a requirement? Can this part also just be moved into the adapter that communicates with the SDK and backend?

yanglin5689446:
Hi @jordansexton,
Indeed first flow is easy to adapt to wallet-adapter with current `sendTransaction` interface. 
As for second flow, consider following transaction containing 2 instructions:
1. A user create an account, assuming that this created account is owned by dApp side.
2. The user transfer specific amount of lamports to this just-created account.
In this case, dApp own the keypair of newly created account, it needs to `transaction.partialSign([newAccount])` for this transaction to work, thus requires the `convertToProgramTx` step.

we have a demo app to show the 2 flows we mentioned above:
demo link: https://demo-sdk.blocto.app/solana
source code here: https://github.com/portto/blocto-sdk-demo-app/blob/c007ed4588291bdc606a37a10eb618c4ce0905af/src/pages/Solana.js#L82

we certainly want to integrate this part of logic with adapter, but haven't come up with a good way to solve this

jordansexton:
> A user create an account, assuming that this created account is owned by dApp side. The user transfer specific amount of lamports to this just-created account.

So they would need another wallet to do this right? Or, if they used the Blocto wallet, could they use the first flow, provided they have some way of getting lamports sent to it, and then the second flow can be used?

I suppose what I'm wondering is, does the dapp need to own / know about the keypair, or can this be handled by the adapter too?

scottphc:
@jordansexton, give a real-world example of how Raydium uses partialSign: https://github.com/raydium-io/raydium-ui/blob/176d616982a562d5c3b7b668dac74417d5b38a7e/src/utils/swap.ts#L384

From wallet's perspective, we can not know what instructions are from dApps. But we know if dApp wants to do partialSign, they must call it after the transaction is converted, in Blocto's flow. That's why we suggest to add `additionalSigners` param.

jordansexton:
@yanglin5689446 @scottphc I made this change by adding it to the options passed to the function. IMO this will let us keep the API simple and extensible without adding more arguments. Let me know if this works for you!

scottphc:
@jordansexton awesome! It should be able to work for us. We're going to make a PR for adding `wallet-adapter-blocto`.

jordansexton:
@scottphc sounds good. I'll hold on merging this for now to see if the design needs to be revised after that implementation.

PR #16, opened 2021-08-23
Added Coin98 wallet.

leonardgee:


packages/coin98/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+  WalletAdapter,
+  WalletAdapterEvents,
+  WalletNotConnectedError,
+  WalletNotFoundError,
+  WalletPublicKeyError,
+  WalletSignatureError,
+  WalletNotInstalledError
+} from '@solana/wallet-adapter-base';
+import { PublicKey, Transaction } from '@solana/web3.js';
+import bs58 from 'bs58'
```
jordansexton, line 14: I see there's a peer/dev dependency on `@types/bs58`, but this suggests that there should be a dependency on both `bs58` and `@types/bs58`, and remove the peer and dev deps (we publish both compiled `lib` and uncompiled TS `src` dirs to npm)

packages/coin98/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+  signTransaction: (transaction: Transaction) => Promise<Transaction>;
+  isConnected: () => boolean,
+  connect: ()=> Promise<string[]>,
+  disconnect: ()=> Promise<void>,
+  request: (params: RequestParams) => Promise<SignTransactionResponse>
+}
+
+interface Coin98Wallet{
+  sol?: Coin98SolProvider,
+  provider: unknown
```
jordansexton, line 38: `provider` isn't used anywhere here.

packages/coin98/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+  private _wallet: Coin98SolProvider | null;
+  private _publicKey: PublicKey | null;
+
+  constructor(config: Coin98WalletAdapterConfig = {}) {
+      super();
+      this._connecting = false;
+      this._wallet = null;
+      this._publicKey = null;
+      this.signTransaction = this.signTransaction.bind(this);
+      this.signAllTransactions = this.signAllTransactions.bind(this);
```
jordansexton, line 63: It shouldn't be necessary to bind these methods -- `signAllTransactions` uses an arrow function to invoke `signTransaction`

packages/coin98/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+          this._connecting = true;
+
+          const wallet = window.coin98?.sol;
+          if (!wallet) throw new WalletNotFoundError();
+          if (!wallet.isCoin98) throw new WalletNotInstalledError();
+
+          let account: string;
+          try {
+              const [address] = await wallet.connect();
+              account = address
```
jordansexton, line 99:
```suggestion
          let account: string;
          try {
              ([account] = await wallet.connect());
```

packages/wallets/src/wallets.ts
```diff
@@ -27,6 +29,13 @@ export interface Wallet {
 
 export const ICONS_URL = 'https://raw.githubusercontent.com/solana-labs/wallet-adapter/master/packages/wallets/icons';
 
+export const getBitpieWallet = (config?: Coin98WalletAdapterConfig): Wallet => ({
+    name: WalletName.Coin98,
+    url: 'https://coin98.com/',
```
jordansexton, line 34:
```suggestion
    url: 'https://coin98.com',
```

Just removing the trailing slash for consistency between the wallets.

jordansexton (commented):
Looks good! Can you run `yarn fmt` from the package?

packages/coin98/src/adapter.ts
leonardgee: removed.

packages/coin98/src/adapter.ts
leonardgee: removed

packages/coin98/src/adapter.ts
leonardgee: fixed.

jordansexton:
Merged, thanks!

PR #19, opened 2021-08-25
Blocto adapter

yanglin5689446:
Add [blocto wallet](https://blocto.portto.com/) adapter package.


packages/blocto/src/adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+    ): Promise<TransactionSignature> {
+        try {
+            if (!this._wallet?.connected) {
+                throw new WalletNotConnectedError();
+            }
+
+            // @TODO: remove this when blocto-sdk implements connection injection
+            if (this._wallet.rpc !== (connection as any)?._rpcEndpoint || connection.commitment) {
+                console.warn("@solana/wallet-adapter-blocto: Blocto SDK has not yet supported commitment nor switching network dynamically, your settings on connection object will be ignored.")
+            }
```
jordansexton, line 121: This is similar to throwing an error because the adapter can't handle the connection as expected, which the dapp is expected to handle at runtime. This is a compromise I don't think is acceptable. I'm also not comfortable with the compromise of printing a console warning, which is worse because the dapp can't see or do anything about it.

The purpose of the wallet-adapter library is to allow wallets to be used interchangeably by dapps and most importantly their users. Not using the connection object provided to send the transaction is a violation of the principle of least surprise. I will wait until this change can be fully implemented to accept this adapter.

jordansexton:
I wasn't able to push to this branch, so I've pushed to https://github.com/solana-labs/wallet-adapter/tree/blocto-adapter

PR #21, opened 2021-08-25
Feature/material UI collapse

piotrjanosz:
- Added a collapse component to make the wallet list shorter
- Added featuredWalletsNumber prop to make wallets number adjustable
- Made font weight on the list items slightly bolder (300 => 400)
- Updated Readme with image sof new functionality


sreekv143 (commented):
Thanks

PR #26, opened 2021-08-27
Feature/phantom UI

piotrjanosz:


jordansexton:
I merged and resolved conflicts from master and made small changes to the readme. I think it's probably overkill to have two full code samples for React UI and Material UI with the code being so similar, so I just kept your React UI sample. Perhaps we could just move the Material UI and Ant Design versions of the setup code to the readme files for those packages, and at the top of the root readme, just show a single image with 3 columns with each of the wallet modals side by side.

At a glance, the new components look really good. I'm a little concerned about how much padding is on them though. Try scrolling over the expanded list of wallets, for example -- they feel far enough apart that there's a lot of mouse travel just to get to one of them. The length of the list also seems like it might have a tough time fitting on a phone.

You might have some ideas about this, and I don't think it's anything we should hold up the PR over, but that's just my initial sense of it. I'll do the actual review of the react-ui and react-ui-starter packages tomorrow. I did notice this z-index issue when building the example:

Screen Shot 2021-08-30 at 9 25 19 PM

Anyway, great work. I'm excited to review and merge this.

chriskalani:
I can take a look at the mobile styles tomorrow. I forgot to include any mocks in Figma. My bad!

piotrjanosz:
> we could just move the Material UI and Ant Design versions of the setup code to the readme files for those packages, and at the top of the root readme, just show a single image with 3 columns with each of the wallet modals side by side.

I think we could also provide clearly visible links to the other repos - so that if people use those packages they will be easily accessible.

If we make the new components the default package I will take a harder look today / tomorrow at a11y and maybe styling the notifications in the example a bit. 

Also, I will fix the ui issues after a check with @chriskalani

piotrjanosz:
As an add-on, I will propose a PR to replace `notistack` with `react-hot-toast` as notistack has material-ui as a dep :)

jordansexton:
> As an add-on, I will propose a PR to replace `notistack` with `react-hot-toast` as notistack has material-ui as a dep :)

In the react-ui-starter package? I'm thinking for the Demo we should probably start to think about splitting it up by UI framework and add a select list or some navigation to view them, rather than put them all together on one page. Then each can use whatever notification system that fits.

piotrjanosz:
I agree, that would be a cleaner solution.
Would you then get rid of starter packages and display all of them in one demo or keep the starters too?

jordansexton:
I'm wondering if the starter packages will become a trap to maintain here. They don't really have anything in them, and realistically if you are already going to use React/Material-UI/Angular/whatever, you probably have a project going before adding this lib.

We don't necessarily have to get rid of them, but maybe we should work on better defining what they are for?

piotrjanosz:
Hey @jordansexton 

Made some changes:
- Modified readme to include the different component next to each other, added props passable to modal provider, links to material-ui and antd repos
- Fixed ui issues
- Added mobile styles
- Improved a11y by adding focus trap to modal an focusable buttons (users can navigate with tab)

Please take a look and let me know what you think.

I am all for moving the other examples (installation, usage, etc.) to material-ui and antd repos.

One thing I'd like to add tomorrow is that notification library replacement, so that there are no peer dependencies (for example, on material-ui).

piotrjanosz:
Forgot to comment here:
> I'm wondering if the starter packages will become a trap to maintain here

I would probably consolidate the example in one place (e.g. only leave example) and allow to change between components with some sort of navigation :)

I think it will become a maintenance drag otherwise.

One thing that I think could be an improvement is sharing the business logic (for example in send transaction or connect) and only importing pure ui components to display for different ui (no) frameworks - that's for another PR though ;)

packages/react-ui-starter/README.md
```diff
@@ -0,0 +1,52 @@
+# `@solana/wallet-adapter-material-ui-starter`
```
jordansexton, line 1:
```suggestion
# `@solana/wallet-adapter-react-ui-starter`
```

jordansexton:
Nice work!

chriskalani:
Looks great!!

PR #29, opened 2021-08-28
add slope wallet

SlopeFinance:
Hello, please review, thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,6 @@ export const Wallet: FC = () => {
 ### Usage
 
 ```tsx
-import { Button } from '@material-ui/core';
 import { WalletNotConnectedError } from "@solana/wallet-adapter-base";
 import { useConnection, useWallet } from '@solana/wallet-adapter-react';
 import { Keypair, SystemProgram, Transaction } from '@solana/web3.js';


packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+
+        if (!this.ready) pollUntilReady(this, config.pollInterval || 1000, config.pollCount || 3);
+    }
+
+    get publicKey(): PublicKey | null {
+        return this._publicKey;
+    }
+
+    get ready(): boolean {
+        return !!window.Slope;
```
jordansexton, line 70:
```suggestion
        return typeof window !== 'undefined' && !!window.Slope;
```

I added some checks in https://github.com/solana-labs/wallet-adapter/pull/32 to handle server side rendering, where the window object isn't available.

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    async connect(): Promise<void> {
+        try {
+            if (this.connected || this.connecting) return;
+            this._connecting = true;
+
+            if (!window.Slope) throw new WalletNotFoundError();
+
+            const wallet = this._wallet || new window.Slope();
```
jordansexton, line 92: I'm not sure how Slope is implemented, but just checking -- are there any possible complications from instantiating `new window.Slope()` multiple times?

The adapter cannot rely on there being only one instantiation of `window.slope`, because the adapter itself may be instantiated multiple times, and therefore `this._wallet` will not be defined for subsequent instances of it.

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+            if (!wallet.connect) throw new WalletNotInstalledError()
+
+            let account: string;
+            try {
+               const { msg, data } = await wallet.connect();
+
+               if (msg !== 'ok') {
+                   throw new WalletAccountError(msg);
+               }
```
jordansexton, line 101:
```suggestion
               if (msg !== 'ok') throw new WalletAccountError(msg);
```

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+            }
+
+            let publicKey: PublicKey;
+            try {
+                publicKey = new PublicKey(account);
+            } catch (error) {
+                throw new WalletPublicKeyError(error?.message, error);
+            }
+
+            wallet.isConnected = true;
```
jordansexton, line 115: `wallet` doesn't define (in the TypeScript above, at least) an `isConnected` property, so this should fail TS compilation. What's the reason to set `isConnected` on the Slope object? Wouldn't Slope already know whether it's connected or not?

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+            this.emit('error', error);
+            throw error;
+        } finally {
+            this._connecting = false;
+        }
+    }
+
+    async disconnect(): Promise<void> {
+        if (this._wallet) {
+            this._wallet.disconnect();
```
jordansexton, line 130: `_wallet.disconnect` returns a Promise and this function is async so it should await that Promise and catch any errors. Please see a reference implementation @ https://github.com/solana-labs/wallet-adapter/blob/f14d5b30f7bdd2f47ee03c0744298f44b8803210/packages/phantom/src/adapter.ts#L157-L161

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const message = bs58.encode(transaction.serializeMessage());
+                const { msg, data } = await wallet.signTransaction(message);
+
+                if (msg !== 'ok') {
+                    throw new WalletSignTransactionError(msg);
+                }
```
jordansexton, line 148:
```suggestion
                if (msg !== 'ok') throw new WalletSignTransactionError(msg);
```

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+} from '@solana/wallet-adapter-base';
+import { PublicKey, Transaction } from '@solana/web3.js';
+import bs58 from 'bs58';
+
+interface SlopeWallet {
+    connect(): Promise<{
+        msg: string;
+        data: {
+            publicKey: string;
+        };
```
jordansexton, line 19: Will `data` be defined even if `msg != 'ok'`?

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+        msg: string;
+        data: {
+            publicKey: string;
+            signatures: string[];
+        }
+    }>;
+}
+
+interface SlopeWindow extends Window {
+    Slope?: any;
```
jordansexton, line 41: This doesn't seem like it should be typed as `any`? The interface is provided.

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,191 @@
...
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const messages = transactions.map(tx => bs58.encode(tx.serializeMessage()));
+                const { msg, data } = await wallet.signAllTransactions(messages);
+
+                if (msg !== 'ok') {
+                    throw new WalletSignTransactionError(msg);
+                }
```
jordansexton, line 175:
```suggestion
                if (msg !== 'ok') throw new WalletSignTransactionError(msg);
```

jordansexton (changes requested):
Looks good, I have a few questions around the `window.Slope` object.

packages/slope/src/adapter.ts
SlopeFinance: Don’t worry, there are no complications, thank you for your checked.

packages/slope/src/adapter.ts
SlopeFinance: I'm not sure how to modify this, changing `any` to another value will cause `new window.Slope()` to report an error.

packages/slope/src/adapter.ts
SlopeFinance: When `msg != 'ok'`, `data = {}`

packages/slope/src/adapter.ts
SlopeFinance: Modified

packages/slope/src/adapter.ts
SlopeFinance: Removed

packages/slope/src/adapter.ts
jordansexton: Since `Slope` has a constructor function, this might do what you want.
```suggestion
    Slope?: {
        new (): SlopeWallet;
    };
```

packages/slope/src/adapter.ts
jordansexton: As long as it's not a problem to instantiate more than once, can we do this? The issue here is that `connect` should ideally start from a fresh state every time.
```suggestion
            const wallet = new window.Slope();
```

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,194 @@
...
+    }>;
+    disconnect(): Promise<{
+        msg: string;
+    }>;
+    signTransaction(message: string): Promise<{
+        msg: string;
+        data: {
+            publicKey?: string;
+            signature?: string;
+        }
```
jordansexton, line 30: Great, I like that you added these as possibly undefined and then check for that.

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,194 @@
...
+            if (!window.Slope) throw new WalletNotFoundError();
+
+            const wallet = this._wallet || new window.Slope();
+            if (!wallet.connect) throw new WalletNotInstalledError()
+
+            let account: string;
+            try {
+               const { msg, data } = await wallet.connect();
+
+               if (msg !== 'ok') throw new WalletAccountError(msg);
```
jordansexton, line 100:
```suggestion
               if (!data.publicKey) throw new WalletAccountError(msg);
```

packages/slope/src/adapter.ts
```diff
@@ -0,0 +1,194 @@
...
+            this._connecting = true;
+
+            if (!window.Slope) throw new WalletNotFoundError();
+
+            const wallet = this._wallet || new window.Slope();
+            if (!wallet.connect) throw new WalletNotInstalledError()
+
+            let account: string;
+            try {
+               const { msg, data } = await wallet.connect();
```
jordansexton, line 98:
```suggestion
               const { data } = await wallet.connect();
```

jordansexton: We can remove `msg` in general since `data` is now neatly typed.

packages/slope/src/adapter.ts
jordansexton: Got it. Thanks!

jordansexton:
I'm not sure what happened exactly, but pushing to this PR caused it to be closed. I made a few small changes and will open a new PR. Sorry about that!

packages/slope/src/adapter.ts
jordansexton: I ended up undoing this change -- I realized `msg` is used by the error.

packages/slope/src/adapter.ts
SlopeFinance: 😊

Issue #30, opened 2021-08-28
Incompatibility with NextJS (11.0.0)


grumpyoldman-io:
When using the latest version of NextJS (with TS), I am unable to use the @solana/wallet-adapter packages.

```
***/node_modules/@solana/wallet-adapter-react/lib/index.js:1
export * from './ConnectionProvider';
^^^^^^

SyntaxError: Unexpected token 'export'
```

I'm quite positive it has to do with how the packages are transpiled, and wouldve hoped adding them to the list of transpiled code in my nextjs project (using [next-transpile-modules]( https://www.npmjs.com/package/next-transpile-modules)) would work, but unfortunately I was not able to get them working (edit: see update bellow). Any tips? If I get it working I'd be happy to submit a PR for either a readme update or change.

tsconfig:
```
{
  "compilerOptions": {
    "target": "es6",
    "lib": [
      "dom",
      "dom.iterable",
      "esnext"
    ],
    "allowJs": false,
    "skipLibCheck": true,
    "strict": false,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve",
  },
  "include": [
    "next-env.d.ts",
    "**/*.ts",
    "**/*.tsx"
  ],
  "exclude": [
    "node_modules"
  ]
}
```

nextjs config:
```
module.exports = {
  webpack5: true
}
```

## Update:

I was finally able to get it working taking quite a detour, I've had to change my nextjs config to the following:
```
/* eslint-disable @typescript-eslint/no-var-requires */
const withTM = require('next-transpile-modules')([
  '@solana/wallet-adapter-base',
  '@solana/wallet-adapter-bitpie',
  '@solana/wallet-adapter-coin98',
  '@solana/wallet-adapter-ledger',
  '@solana/wallet-adapter-mathwallet',
  '@solana/wallet-adapter-phantom',
  '@solana/wallet-adapter-react',
  '@solana/wallet-adapter-solflare',
  '@solana/wallet-adapter-sollet',
  '@solana/wallet-adapter-solong',
  '@solana/wallet-adapter-torus',
  '@solana/wallet-adapter-wallets',
  '@project-serum/sol-wallet-adapter',
])

module.exports = withTM({
  webpack5: true
})
```

[...]

PR #32, opened 2021-08-30
Changes for Next.js support

jordansexton:
re: https://github.com/solana-labs/wallet-adapter/issues/30


grumpyoldman-io (commented):
I think this solves a lot of the issues, but I also noticed external dependencies relying on window (Torus labs one for instant). Which is automatically imported because it is used in the `wallets` package. But lets first validate these changes.

grumpyoldman-io:
I've tested these packages in my nextjs project:

@solana/wallet-adapter-base
@solana/wallet-adapter-ledger
@solana/wallet-adapter-mathwallet
@solana/wallet-adapter-phantom
@solana/wallet-adapter-react
@solana/wallet-adapter-solflare
@solana/wallet-adapter-sollet
@solana/wallet-adapter-solong

and can confirm the import works, but the third party dependencies indeed still break:

```
xxx/node_modules/@project-serum/sol-wallet-adapter/dist/cjs/index.js:79
import EventEmitter from 'eventemitter3';

SyntaxError: Cannot use import statement outside a module
```

jordansexton:
Interesting. I was trying not to add a build tool for this library besides tsc, but it may be unavoidable.

jordansexton:
I've merged this, but will leave #30 open until I can integrate webpack/rollup/something this week.

grumpyoldman-io:
@jordansexton Can I suggest Microbundle? I think that could hit the spot just right.

jovincroninwilesmith:
Getting this same issue 

```
/node_modules/@project-serum/sol-wallet-adapter/dist/cjs/index.js:79
import EventEmitter from 'eventemitter3';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

Hossman333:
As am i. 😊 I'll take a look at #30 issue to see what I can do.

jordansexton:
Yeah, #30 has some possible workarounds. This library includes some external dependencies but doesn't transpile them, so it may come down to adding some config to Next / Webpack / Babel. Please let me know what you find out!

PR #40, opened 2021-08-30
Add a React hook for the Wallet type expected by Anchor

jordansexton:


JuliaMerz:
Heya! Sorry for the necro, but I was curious what the reasoning behind this was given that useWallet appears to be a superset of useAnchorWallet. Wouldn't using useWallet be fine, since Anchor would only call the functions that it understands?

I'm working with an API that tries to use signMessage, and if I can use regular useWallet inside anchor on the web that would be a huge boon, so I wanted to check in case I missed something..

jordansexton:
You can use both! `useAnchorWallet` is not a pure subset of `useWallet`. Use the former to get an interface compatible with what Anchor expects (or did... that function was added early on, and Anchor has changed a lot). Use the latter to get a lower level interface if you need it for message signing.

PR #42, opened 2021-09-01
feat: Angular support for wallet-adapter

### LINKED ISSUE #37: Angular support
# Overview

I'm trying to replicate the [react library](https://github.com/solana-labs/wallet-adapter/tree/master/packages/react) in Angular, I'm getting to a point where things start to make sense. I was wondering if this is something you'd like to include in the library set.

I think the pain point of doing this would be the UI part, from the top of my head I can think of two alternatives:

- Make UI libs framework agnostic (hard to code IMO)
- Create different sets of UI libs per framework (easier to code, harder to maintain)

Given the React-ish approach of everything Solana-related, I'm thinking of doing this on my own in a different monorepo. Not having Angular-support held me back for a while, would love to make it easy for other Angular fellas like me.

danmt:
# Overview

As an Angular dev there's no straightforward way of using this package without having to code everything from scratch. This PR intention is to give support for Angular devs trying to integrate `@solana/wallet-adapter` into their dapps.

## Features

In order to ensure this package is analogous to `@solana/wallet-adapter-react` a list of features is exposed below to make sure nothing is missed.

- [x] Connect to a wallet
- [x] Disconnect from a wallet
- [x] Change selected wallet (disconnecting current wallet and persist in localStorage)
- [x] Send transaction
- [x] Sign transaction
- [x] Sign all transactions
- [x] Sign message
- [x] Allow auto connect
- [x] Listen to wallet adapter events
- [x] Customizable error handler
- [x] Mapping from wallet to anchor wallet
- [x] Managing the connection

## NOTES

The state of the wallet and connection is handled by @ngrx/component-store which is the recommended tool, the component-store is very similar to the Providers in the React library. There are some `providers` that are similar to the Context from React, it just takes a Provider and makes it available for a component and its children.

closes #37


danmt:
CRA depends on `babel-loader@8.1.0` and `webpack@4.4.0`, while Angular depends on `babel-loader@8.2.2` and `webpack@5.50.0`. Angular deps are hoisted to the root by `yarn`, leading to a warning. I had to do [this commit](https://github.com/danmt/wallet-adapter/commit/eb6c45fd4ec3c3e7f656719f9b599c93d72b4642) in order to bypass the error. 

Might worth taking a second look at it. I'm open to questions as needed!

packages/angular/src/wallet/wallet.errors.ts
```diff
@@ -0,0 +1,29 @@
...
+        this.name = 'SignAllTransactionsNotFoundError';
+    }
+}
+
+export class SignMessageNotFoundError extends WalletError {
+    constructor() {
+        super();
+        this.name = 'SignMessageNotFoundError';
+    }
+}
```
jordansexton, line 29: It would be ideal if there was a way to not throw runtime errors when these are used, which would mean this package wouldn't need these errors. The way the React package for this works is that feature detection is done in a TypeScript compatible way, so if the method is defined by the adapter, it's callable. Is there a way we can do that here?

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,290 @@
...
+    readonly disconnecting$ = this.select((state) => state.disconnecting);
+    readonly wallet$ = this.select((state) => state.wallet);
+    readonly adapter$ = this.select((state) => state.adapter);
+    readonly publicKey$ = this.select((state) => state.publicKey);
+    readonly ready$ = this.select((state) => state.ready);
+    readonly anchorWallet$ = this.select(this.publicKey$.pipe(isNotNull), (publicKey) => ({
+        publicKey,
+        signTransaction: (transaction: Transaction) => this.signTransaction(transaction).toPromise(),
+        signAllTransactions: (transactions: Transaction[]) => this.signAllTransactions(transactions).toPromise(),
+    }));
```
jordansexton, line 36: Anchor expects this interface: https://github.com/project-serum/anchor/blob/2604a442fda443f333cf7d3340892b95bc5f776b/ts/src/provider.ts#L220-L224

I'm thinking this should also feature detect if the signTransaction and signAllTransaction methods are defined. This way it's all or nothing, as in the React package.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,290 @@
...
+                }
+
+                return from(defer(() => adapter.signAllTransactions(transactions))).pipe(
+                    map((transactions) => transactions as Transaction[])
+                );
+            })
+        );
+    }
+
+    signMessage(message: Uint8Array): Observable<Uint8Array> {
```
jordansexton, line 270: Touched on this already, but my thinking is that these methods should be conditionally defined. Right now, Phantom is the only wallet that supports arbitrary message signing. It's a rough UX if users get hit with a runtime error because it's not supported by their wallet and the dapp doesn't have a reliably way of detecting this.

packages/angular/src/wallet/wallet.errors.ts
danmt: Let's see if I get this right. We want the `signMessage` to be of type `((message: Uint8Array) => Observable<Uint8Array>) | undefined`?

From my understanding, that means consumers would have to do something like:

```ts
if (signMessage) {
    signMessage(message)
}
```

And that goes for all the sign-methods, right? If that's the case it seems like a doable thing.

Aside from that, in RxJs, if an error is thrown inside a stream like I'm doing there it won't terminate the **app**, instead, it would terminate the **stream**. This way consumers can have a handler for the error to inspect if something went wrong, but if they don't add the handler, the app won't break because the error is not treated as an unhandled exception at the application level.

packages/angular/src/wallet/wallet.store.ts
danmt: Gotcha

packages/angular/src/wallet/wallet.errors.ts
jordansexton: Yeah, exactly. In the React package, if they do `const { signTransaction } = useWallet()` the function can be undefined, so they'll get a tsc error if they try to use it without checking. This also encourages dapp developers to not rely on all wallets having all features, since using the common `sendTransaction` interface is the golden path

danmt: I hear ya.

I'll try to figure that out, given that the store is a class and the sign-methods are instance methods, it might not be trivial.

On the other hand, the sign-methods I use atm return an observable, in order for the observable to fire, the consumer needs to call .subscribe, if my method returns undefined they would get a compilation error, forcing them to check before subscribing.

I think that if the sign-methods have a type of `(input) => Observable | undefined`, it will behave as we expect, the function definitions won't be undefined but at least the library won't allow consumers to subscribe without checking. I know this isn't exactly analogous but would have the behavior intended.

danmt: Just for clarity, for example if the `signMessage` is of type `(message: Uint8Array) => Observable<Uint8Array> | undefined`, a consumer trying to sign a message would have to:

```typescript
const signMessage$ = signMessage(message);

if (!signMessage$) {
     // notify error and return
}

signMessage$.subscribe()
```

The lazy nature of an observable makes it that just calling the function doesn't trigger `adapter.<method>`, but we have to subscribe. If the "factory method" creating the observable returns undefined calling `.subscribe` will fail at compilation.

jordansexton: Huh, interesting. Yeah, I think as long as it's reasonably idiomatic/expected behavior for Angular/Rxjs users, and will produce TS errors instead of runtime, that's fine.

danmt: IMHO, the error is what I'd expect, but the main goal is to make it as similar to the react package as possible. If at some point in time it's decided to switch to an error in React, I'll gladly update this. 

I'll have that fixed by next week and will include the rest of the feedback, currently buidling something cool for Ignition so I'm juggling with a lot atm.

jordansexton: The core issue as I see it is that dapps can't be expected handle different wallets differently, and they need a way to feature detect. Otherwise the error has to be thrown and handled at runtime, or bubble up to the user, which is bad for everyone because they aren't able to do anything about it.

Take your time and good luck in the hackathon :)

danmt: I totally get the reasoning behind it and I think it's a cool design, is just that I find errors a bit more expressive to know what's going on, as a dev I'm used to try/catch every promise/observable as a rule of thumb. Don't get wrong, I'm a bit Angular-biased, and since I use a lot of classes and Object-oriented stuff, having a "conditional method" makes some noise to me. 

Thanks a lot man!!

dubin-s:
If there is anything I can do to help get this merged in please let me know, would be happy to work with you @danmt to resolve anything.

danmt:
Right on @dubin-s. I'm currently working on my hackathon project so I haven't been able to wrap this one up. 

If you want to have early access check this out [@danmt/wallet-adapter-angular](https://www.npmjs.com/package/@danmt/wallet-adapter-angular), it's basically a fork I have for development purposes. I'm using it in a project and it's looking fine. It's literally the same thing you see in the PR plus some minor adjustments.

You can give it a try and drop feedback here, I'd love some extra pair of eyes on it.

jordansexton:
Might be good to merge master in, and check the simplifying state changes that have been made to the `react` package since this is based on that.

danmt:
I can do the sign-methods fix we talked about and the changes you mentioned. I've been using this for a little while and it's looking good, we can push it even if it's not ready to start gathering feedback. 

I can have this later today.

jordansexton:
Awesome, sounds good. Also, hit me up on Discord -- I'd be happy to take a look at your project if you like.

danmt:
I changed the sign-methods to be `Observable<type> | undefined` for all sign-methods. Also, made the anchorWallet thing so it returns `undefined` when the selected wallet cannot sign. 

BTW - The only way I could bypass the `babel-loader` error was by adding `SKIP_PREFLIGHT_CHECK=true` to every CRA package (example, material-ui-starter, react-ui-starter).

I successfully built every package. It wasn't as easy as I wanted with Lerna but that should work.

danmt:
Once this gets merged and published, will update my example app and will raise another PR for that.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+                }
+
+                return from(defer(() => adapter.sendTransaction(transaction, connection, options))).pipe(
+                    map((txId) => txId as string)
+                );
+            })
+        );
+    }
+
+    signTransaction(transaction: Transaction): Observable<Transaction> | undefined {
```
jordansexton, line 239: How would this be called by a dapp, and how does the dapp detect whether the adapter allows it? Just trying to understand the type safety this provides.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+export const WALLET_DEFAULT_CONFIG: WalletConfig = {
+    wallets: [],
+    autoConnect: false,
+    localStorageKey: 'walletName',
+    onError: (error: unknown) => console.error(error),
+};
+
+@Injectable()
+export class WalletStore extends ComponentStore<WalletState> {
+    private readonly _autoConnect = this._config?.autoConnect || false;
```
jordansexton, line 23: Should this be readonly? I can't necessarily think of a case where the dapp would want to change this value, but I'm hesitant to enforce that at the type level.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+                          this.signAllTransactions(transactions)?.toPromise(),
+                  }
+                : undefined;
+        }
+    );
+
+    constructor(
+        @Optional()
+        @Inject(WALLET_CONFIG)
+        private _config: WalletConfig
```
jordansexton, line 53: Can this only be provided in the constructor, or is there a way to change the config values afterward?

packages/angular/src/operators/from-adapter-event.ts
```diff
@@ -0,0 +1,14 @@
+import { SignerWalletAdapter, WalletAdapter, WalletAdapterEvents } from '@solana/wallet-adapter-base';
+import { fromEventPattern, Observable } from 'rxjs';
+import { switchMap } from 'rxjs/operators';
+
+export const fromAdapterEvent =
+    (eventName: keyof WalletAdapterEvents) => (source: Observable<WalletAdapter | SignerWalletAdapter>) =>
```
jordansexton, line 6: What does `SignerWalletAdapter` do here?

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+    readonly wallets$ = this.select((state) => state.wallets);
+    readonly selectedWallet$ = this.select((state) => state.selectedWallet);
+    readonly connected$ = this.select((state) => state.connected);
+    readonly connecting$ = this.select((state) => state.connecting);
+    readonly disconnecting$ = this.select((state) => state.disconnecting);
+    readonly wallet$ = this.select((state) => state.wallet);
+    readonly adapter$ = this.select((state) => state.adapter);
+    readonly publicKey$ = this.select((state) => state.publicKey);
+    readonly ready$ = this.select((state) => state.ready);
+    readonly anchorWallet$ = this.select(
```
jordansexton, line 35: How does a dapp reference this?

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+export class WalletStore extends ComponentStore<WalletState> {
+    private readonly _autoConnect = this._config?.autoConnect || false;
+    private readonly _localStorageKey = this._config?.localStorageKey || 'walletName';
+    private logError = this._config?.onError || ((error: unknown) => console.error(error));
+    readonly wallets$ = this.select((state) => state.wallets);
+    readonly selectedWallet$ = this.select((state) => state.selectedWallet);
+    readonly connected$ = this.select((state) => state.connected);
+    readonly connecting$ = this.select((state) => state.connecting);
+    readonly disconnecting$ = this.select((state) => state.disconnecting);
+    readonly wallet$ = this.select((state) => state.wallet);
```
jordansexton, line 31: Can we change `selectedWallet` to `name` or `walletName`? `selectedWallet` implies that it's the same type as `wallet`, and keeping some parity with the names in the react package will probably help us maintain this easier when either changes.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+        if (!this._config) {
+            this._config = WALLET_DEFAULT_CONFIG;
+        } else {
+            this._config = {
+                ...WALLET_DEFAULT_CONFIG,
+                ...this._config,
+            };
+        }
+
+        const walletName = localStorage.getItem(this._localStorageKey);
```
jordansexton, line 77: In the react package we JSON encode this value, so the literal value is e.g. `"WalletName"`.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+                    return of(null);
+                }
+
+                if (!ready) {
+                    window.open(wallet.url, '_blank');
+                    this.logError(new WalletNotReadyError());
+                    return of(null);
+                }
+
+                return from(defer(() => adapter.connect())).pipe(catchError(() => of(null)));
```
jordansexton, line 112: My reading of this is that it's doing something like
```ts
try {
    await adapter.connect();
} catch (error) {
    return null;
}
```
Is that accurate? It doesn't seem to be logging or rethrowing the error.

packages/angular/.prettierignore
```diff
@@ -0,0 +1 @@
+/lib/**
```
jordansexton, line 1: nit: I think just `/lib` should work

jordansexton (commented):
Thanks for this! Mostly I have a bunch of questions with a few small change requests.

packages/angular/package.json
```diff
@@ -0,0 +1,46 @@
+{
+    "name": "@solana/wallet-adapter-angular",
+    "version": "0.0.2",
```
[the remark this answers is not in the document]
danmt: Easy peasy

packages/example/.env
```diff
@@ -1,2 +1,3 @@
 HTTPS=true
 PUBLIC_URL=https://solana-labs.github.io/wallet-adapter/example/
+SKIP_PREFLIGHT_CHECK=true
```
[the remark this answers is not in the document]
danmt: The CRA packages have their own node_modules with `babel-loader@8.1.0` included, while Angular uses `babel-loader@8.2.2`. Not sure why but CRA errors during the build if it finds another version of `babel-loader` up in the chain, basically the node_modules hoisting makes it fail. 

They show an ugly message with some instructions I tried to follow but couldn't make it work, even trying with resolutions. I figured you could help me knock this one.

packages/angular/tsconfig.lib.json
```diff
@@ -0,0 +1,31 @@
...
+        "lib": ["dom", "es2018"],
+        "allowSyntheticDefaultImports": true
+    },
+    "angularCompilerOptions": {
+        "enableI18nLegacyMessageIdFormat": false,
+        "strictInjectionParameters": true,
+        "strictInputAccessModifiers": true,
+        "strictTemplates": true
+    }
+}
```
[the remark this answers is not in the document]
danmt: In Nx, which is what I usually use for my workspaces, there's the concept of app (a buildable and deployable artifact) and libs (a package that's symlinked so you can use it as if it was an external thing). In a common Nx workspace, there's a "base" tsconfig at the root path of the workspace, but also there's multiple tsconfigs per app/lib. As a convention you suffix your lib-related tsconfigs with "lib" and app-related tsconfigs with "app".

Also you may have noticed the fact that there's a `tsconfig.lib.json` and `tsconfig.lib.prod.json`, the reason is that you may have different rules based on the configuration you're looking for. There's also a `tsconfig.lib.test.json` that I removed on purpose.

Leaving all that aside (I just wanted to explain the reason behind it), I can easily change the names so they match the standard and make it easy for everyone. We can always re-visit this later on.

packages/angular/src/wallet/wallet.store.ts
danmt: Angular uses DI for accessing singletons, which is the case of the WalletStore. Consumers can do:

```typescript
class WildComponent {
    constructor(private walletStore: WalletStore) {}
}
``` 

With that in place, they have access to the walletStore singleton instance and they can do something like `this.walletStore.anchorWallet$` to access the observable. What I do in my dapp is:

```typescript
const anchorProvider$ = combineLatest([
  this.connectionStore.connection$,
  this.walletStore.anchorWallet$
]).pipe(
  map(
    ([connection, anchorWallet]) => 
      anchorWallet ? 
        new Provider(connection, anchorWallet, Provider.defaultOptions()) : 
        undefined
))
```

A responsible developer would also add an error notification of some kind which would teardown the observable and wouldn't continue.

packages/angular/src/wallet/wallet.store.ts
danmt: As it is, you can't change the values afterward. 

I can add a `setConfig` method to allow consumers to change it later, no biggie. The reasoning behind putting this in the constructor is because it's what you usually see in Angular maintained libraries, I'd keep it while also allowing consumers to change it on the fly.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+            connecting: false,
+            disconnecting: false,
+            ready: false,
+            publicKey: null,
+            autoApprove: false,
+        });
+
+        if (!this._config) {
+            this._config = WALLET_DEFAULT_CONFIG;
+        } else {
```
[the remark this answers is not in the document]
danmt: Yeah, this was a bad judgement call LOL.

packages/angular/src/wallet/wallet.store.ts
danmt: Unlike `Promises` that are eager, `Observables` are lazy, they do nothing until you `subscribe`. Even if you call `signTransaction`, nothing happens until you subscribe. Given that this can return undefined, you can't simply subscribe without checking first.

A common dapp usage would have to look like:

```typescript
const signTransaction$ = this.walletStore.signTransaction(tx);

if (!signTransaction$) {
  notifyError()
}

signTransaction$.subscribe()
```

Without the check, the signTransaction$ can be undefined and wouldn't have the subscribe method. Given the `adapter && 'signTransaction' in adapter` check, the adapter type inferred is `SignerWalletAdapter`, and for  `signMessage` it's `MessageSignerWalletAdapter`.

packages/angular/src/operators/from-adapter-event.ts
danmt: This is a custom RxJs Observable that maps an adapter event listener into an observable, which is pretty handy. This operator would receive an adapter which can be `WalletAdapter | SignerWalletAdapter | MessageSignerWalletAdapter`, I believe the compiler was crying at me for some reason and I had to add this, let me double-check because it's weird that I didn't include `MessageSignerWalletAdapter` and yet it works

packages/angular/src/wallet/wallet.store.ts
danmt: TBH I wouldn't expect consumers to change the config on runtime but maybe there's a use-case I'm not seeing. I just add readonly unless I'm sure I'll be changing it. At the time of writing I assumed consumers wouldn't be changing config afterward.

It's an easy change though.

packages/angular/src/wallet/wallet.store.ts
danmt: You're definitely right. I think `name` is the way to go so it matches the react package

packages/angular/src/wallet/wallet.store.ts
danmt: Yes, this is accurate. The thing is that I have the adapter.on('error') also running, if there's an error during connection I would expect that to catch it, am I wrong here? 

Let me guess, not all wallets work the same and some do fail during connect but the error is not emitted through the error event.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+            };
+        }
+
+        const walletName = localStorage.getItem(this._localStorageKey);
+        const wallet = this._config.wallets.find(({ name }) => name === walletName);
+
+        if (wallet) {
+            this.selectWallet(walletName as WalletName);
+        } else if (this._config.wallets.length > 0) {
+            this.selectWallet(this._config.wallets[0].name);
```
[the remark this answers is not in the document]
danmt: The reason behind this is that I wanted to **always** have a selected wallet, thinking it through, this is probably something we want consumers to choose instead of imposing it.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+
+    readonly onDisconnect = this.effect(() => {
+        return this.adapter$.pipe(
+            isNotNull,
+            fromAdapterEvent('disconnect'),
+            tap(() =>
+                this.patchState({
+                    connected: false,
+                    connecting: false,
+                    disconnecting: false,
```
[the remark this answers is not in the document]
danmt: You can think of `tap` as a way to trigger side-effects, in this case, to update the store state.

The connect/disconnect methods handle the connecting/disconnecting stuff. But I thought it was possible for the adapter.on('disconnect') to be fired for an external reason and I would still want the connecting/disconnecting to be set to false. 

I can't think of a good reason for that to happen ATM, so I'm OK removing those two from there.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+
+    readonly connect = this.effect((action$: Observable<void>) => {
+        return action$.pipe(
+            concatMap(() => of(null).pipe(withLatestFrom(this.state$, (_, state) => state))),
+            filter(({ connected, connecting, disconnecting }) => !connected && !connecting && !disconnecting),
+            tap(() => this.patchState({ connecting: true })),
+            concatMap(({ adapter, wallet, ready }) => {
+                if (!wallet || !adapter) {
+                    this.logError(new WalletNotSelectedError());
+                    return of(null);
```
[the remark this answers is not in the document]
danmt: The reality is that throwing the error there was killing the entire stream and calling connect afterward wasn't working, so I decided to hack it that way, I can give it a second try.

packages/angular/src/wallet/wallet.store.ts
danmt: Would you mind elaborating on this? I think don't get it

packages/angular/src/wallet/wallet.store.ts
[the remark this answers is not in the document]
jordansexton: Yeah, what if the first wallet on the list doesn't work on the user's browser/device, or they just don't use it? Also, a UI would then show that wallet as selected when they didn't do anything.

packages/example/.env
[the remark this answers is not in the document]
jordansexton: Cool. Yeah, I'll try to build it and see. I think the preflight checks also do TypeScript and lint checks, so it's probably not great to silence them.

packages/angular/tsconfig.lib.json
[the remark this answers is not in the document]
jordansexton: Cool, thanks for the detailed explanation. I'm inclined to stick with `tsconfig.json` for consistency and not commit to Nx until we sort out https://github.com/solana-labs/wallet-adapter/issues/66 but I don't have a strong opinion as long as this works out of the box for most users.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+                if (!adapter) {
+                    return throwError(new WalletNotSelectedError());
+                }
+
+                if (!connected) {
+                    return throwError(new WalletNotConnectedError());
+                }
+
+                return from(defer(() => adapter.sendTransaction(transaction, connection, options))).pipe(
+                    map((txId) => txId as string)
```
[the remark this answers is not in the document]
danmt: I'm 100% sure this wasn't inferring the type but I just checked and it does. I'll fix it right away

packages/angular/src/wallet/wallet.store.ts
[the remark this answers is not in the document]
jordansexton: `adapter.on('disconnect')` can be fired by the adapter itself, but in that case, we will not be connecting or disconnecting anyway. That's to indicate that the adapter has already disconnected and so should we.

packages/angular/src/wallet/wallet.store.ts
jordansexton: Got it, makes sense. Thanks.

packages/angular/tsconfig.lib.json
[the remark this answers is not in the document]
danmt: I'm there with you.

packages/angular/src/wallet/wallet.store.ts
jordansexton: Yeah, hard for me to say. It's private anyway so it's nonbreaking to change this later. However it's worth noting that the react package exposes the value of `autoConnect` in the context so that descendants can detect whether it's enabled or not.

packages/example/.env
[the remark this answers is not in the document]
danmt: Indeed, we need a better way. Skipping preflight checks is almost never the way to go.

packages/angular/src/wallet/wallet.store.ts
```diff
@@ -0,0 +1,295 @@
...
+    autoConnect: false,
+    localStorageKey: 'walletName',
+    onError: (error: unknown) => console.error(error),
+};
+
+@Injectable()
+export class WalletStore extends ComponentStore<WalletState> {
+    private readonly _autoConnect = this._config?.autoConnect || false;
+    private readonly _localStorageKey = this._config?.localStorageKey || 'walletName';
+    private logError = this._config?.onError || ((error: unknown) => console.error(error));
```
jordansexton, line 25: Should this be `_logError`?

packages/angular/src/wallet/wallet.store.ts
jordansexton: Yeah, it's probably fine to leave it as is. If we see a need for adding the ability to change it, we can always add later.

packages/angular/src/wallet/wallet.store.ts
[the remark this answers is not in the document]
danmt: I was thinking more of a race condition, you try to connect, and somehow the adapter disconnects and you get in this weird mid-stage. It shouldn't happen though, and even if it did, the connect/disconnect call should handle it

packages/angular/src/wallet/wallet.store.ts
jordansexton: LocalStorage values need to be strings, so to encode any kind of data, it's a common pattern to use a wrapper around the API to `JSON.stringify` in and then `JSON.parse` out, which has the effect of quoting strings.

This probably doesn't matter (it's not like you're just going to switch your app from using React to Angular and care if wallets autoConnect) but I'm thinking we should do it for consistency.

packages/angular/src/wallet/wallet.store.ts
danmt: Nice catch

packages/angular/src/wallet/wallet.store.ts
jordansexton: The adapter will always both emit an event and throw an error if the connect call fails, so I think you're right that we don't need to log that, since it will already get logged.

 I'm wondering how it works when you call `connect` in your app. The global error handler is really just there to provide end-user notifications, so you can still catch errors from your `connect` call (or ignore them like https://github.com/solana-labs/wallet-adapter/blob/e94577e73b483c2425006a28a37f9f2662ee12f5/packages/react-ui/src/WalletConnectButton.tsx#L13).

But maybe there's a more idiomatic way of doing this in this framework?

packages/angular/src/wallet/wallet.store.ts
danmt: Ohh I see, my train of thought was: If the walletName is a string, I don't need to encode/decode. But I just noticed in the React package it's a WalletName | null, not a plain string. I think I got it.

packages/angular/src/wallet/wallet.store.ts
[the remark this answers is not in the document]
jordansexton: Hmm, okay. Yeah, I guess I just don't understand the expected behavior for these methods. In the React code they are really just vanilla JS, so they throw errors if there are errors.

packages/angular/src/wallet/wallet.store.ts
jordansexton: Interesting, okay. Makes sense then.

packages/angular/src/wallet/wallet.store.ts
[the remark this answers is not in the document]
jordansexton: Cool. There are a few more of these constructors in the other methods, but I figured I'd check first.

packages/angular/src/wallet/wallet.store.ts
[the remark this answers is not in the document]
danmt: I don't like that hack, it gives me the creeps lol. I mean the method works, I have that working in a dapp but I don't like the fact that it's still a hack. 

Let me see if I can crack it down.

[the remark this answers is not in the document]
jordansexton: My main issue is that if you return null, the code that wraps this doesn't know how to handle it.

For example, in the React code, you can do:
```ts
const { connect } = useWallet();
//...
try {
    await connect();
} catch (error) {
    if (error instanceof WalletNotSelectedError) {
        // do something specific
    }
}
```

This general pattern derives from web3-react: https://github.com/NoahZinsmeister/web3-react/blob/f5c91a368a7cc15d0ae801288212c384449eb464/packages/injected-connector/src/index.ts#L11-L17

Each wallet can throw errors with different levels of specificity and apps can handle them differently.

packages/angular/src/wallet/wallet.store.ts
jordansexton: Ah, WalletName is an enum of a string without the quotes. But if you look at the `useLocalStorage` hook, that's where the wrap/unwrap and typecasting happens.

packages/angular/src/wallet/wallet.store.ts
danmt: This is something peculiar of how the @ngrx/component-store works, when you trigger an effect, you **can't** wait for a response, you just fire and forget. If you take a look at the send and sign methods, I couldn't use an effect because otherwise, consumers wouldn't be able to wait for the response to do something else. 

The idea is to make effects more declarative but that sometimes hurts ergonomics IMHO. My rule of thumb is, do I really need to wait for the response? If I do, I return an observable and let consumers handle the lifecycle of the observable, otherwise, I encapsulate it inside an effect.

packages/angular/src/operators/from-adapter-event.ts
jordansexton: Huh, okay. I would think that just `WalletAdapter` would work since the others inherit.

packages/angular/src/wallet/wallet.store.ts
jordansexton: I see. Can we at least return the type of error that was caught instead of null, or is there another way to access it for consumers?

packages/angular/src/wallet/wallet.store.ts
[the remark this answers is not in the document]
danmt: This relates to [this comment](https://github.com/solana-labs/wallet-adapter/pull/42#discussion_r706651931), you can't actually call this.walletStore.connect() and *wait* for an error/response, that's handled by the component-store library. When you call `connect()`, you're just triggering the effect, then the store tries to connect, if it does connect it updates the state if not it handles the error internally, it's common to add the error to an internal state so consumers can "listen" to an errors stream, that sends every error.

packages/angular/src/wallet/wallet.store.ts
jordansexton: Oh, an important new behavior I missed in review -- https://github.com/solana-labs/wallet-adapter/blob/e94577e73b483c2425006a28a37f9f2662ee12f5/packages/react/src/WalletProvider.tsx#L127-L143

If the connection fails, we reset the state, including the selected wallet. Many dapps just show a single select/connect button, and this prevents the user from getting locked into a state where the wallet is selected but connecting fails (for example, if the wallet isn't viable on their platform). Especially since the state is in localStorage, this produced a really bad UX where the user had now way to recover except to know to clear their cookies.

danmt: Funnily this method returns void and that's how the library works LOL. But what you usually do, is create an error stream and push everything there, that way consumers can do:

```typescript
error$.subscribe((error) => notifyError)

// or this
error$
  .pipe(filter((error) => error instanceof WhatEverError))
  .subscribe((error) => notifyError)
```

packages/angular/src/wallet/wallet.store.ts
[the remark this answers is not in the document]
jordansexton: Ah, I see. 

For example, one common user flow in EVM apps is to try to connect, and if the user has Metamask installed but they don't have it configured to use the chain your app is on (say, BSC), you can actually catch the error, and then send a request to Metamask to use BSC instead of the chain they are on. The user accepts, they never see an error, the app state just shows them as connecting the whole time, the connection goes through, and the whole thing is basically seamless.

As long as there's an idiomatic way to handle the error like you showed in https://github.com/solana-labs/wallet-adapter/pull/42#discussion_r706657394 I think that makes sense then. It definitely weirds me out that you can't handle the error locally though.

danmt:
Yo @jordansexton, I decided to go the extra mile and made some refactoring so the package maps almost 1:1 with the react package. As you would expect, there are some places I couldn't do the exact same thing but I tried to keep the same behavior.

Things to have in mind:

- I removed `_logError ` which is analogous to `onError` from the react package. The reason behind it is that I created the error stream I mentioned in our discussion, if the consumer wants to "tap" into the errors being thrown all it takes is to subscribe to the exposed error$.
- Given that the WalletStore is a class, making the sign methods *conditional* was a real challenge, I ended up creating a higher-order function per method, each returns `undefined` if the adapter cannot do that or a function that returns takes whatever you want to sign and returns an observable that on subscription does the actual signing.
- I don't have to set up the event listeners and tearing them down because `switchMap`  + the custom operator `fromAdapterEvent` takes care of that for me.
- I turned the connect/disconnect methods from effects into methods returning observables, that way the consumers have full control of it. I figured that if it was confusing for both of us, it would be too for a newcomer.

Things left to do:

- The SKIP_PREFLIGHT thingy, I'm not sure how to deal with it but I'm open to suggestions.

Besides that, I re-ordered values and methods just like the React package for easier follow-up.

dubin-s:
Going to take a look at this today @danmt. I see there have been lots of updates since I looked at this last. Was MIA over the weekend.

danmt:
Essentially it works the same way, I made adjustments so it was more similar to the react package. As well as adding a bunch of suggestions from @jordansexton. If you're familiar with the react package, this should be easy to follow, if it isn't feel free to drop a comment.

jordansexton:
Awesome, thanks @danmt. I'll try to review this today.

jordansexton:
Awesome work, thank you @danmt 🚀

danmt:
Wohoo! 🎉

PR #45, opened 2021-09-02
Clear selected wallet on fatal connection failures

jordansexton:
This should fix https://github.com/solana-labs/wallet-adapter/issues/44 by clearing the selected wallet and adapter when a user initiates a connection that fails because the wallet isn't found, or when autoConnect reports the wallet is ready but the connection still fails.

It should also make the connection UX slightly smoother by using synchronous state changes for the selected wallet and adapter rather than using an effect to do it.


jordansexton:
This is a breaking change because it changes the optional type of `wallet` and `adapter` in the context state to `null` instead of `undefined`.

PR #47, opened 2021-09-02
Fixed width on modal

piotrjanosz:
Fixed centering on react-ui modal so that it now appears in the middle of the screen. 
Added a wrapper div around the modal wrapper - that's why the change looks big :)

Made modal title smaller on smallest mobile phones too.


jordansexton:
Thanks!

PR #60, opened 2021-09-09
Fixed ui render bug

TheGrado:
Fixed a render bug caused by short circuit evaluation on the modal collapse.
A "0" would render after the wallets list if you included every wallet in the featured wallets section.

The bug happened when trying to render every wallet without showing the "More options" button.

Fixed by replacing the short circuit evaluation with a ternary operator inside the render functions.


jordansexton:
Ah, good call. Thanks!

jordansexton:
These packages and their dependents have been published to NPM.

PR #63, opened 2021-09-10
blocto: update deps

yanglin5689446:
Update blocto SDK version in order to fix mobile browser injected provider issue


jordansexton:
Looks fine, though technically anyone who installs this as a dependency can `yarn update` Blocto's SDK already since the fix is a patch release.

PR #78, opened 2021-09-15
Add signMessage to Sollet

richard-fisher:


packages/wallets/sollet/src/adapter.ts
```diff
@@ -149,6 +149,23 @@ export class SolletWalletAdapter extends BaseSignerWalletAdapter {
         }
     }
 
+    async signMessage(message: Uint8Array): Promise<Uint8Array> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const { signature } = await wallet.sign(message, 'utf8');
```
jordansexton, line 158: Overall looks good, but I'm concerned about the `utf8` param here. Looking at the implementation of `sign` --

```tsx
export default class Wallet extends EventEmitter {
  // ...
  async sign(
    data: Uint8Array,
    display: unknown,
  ): Promise<{
    signature: Buffer;
    publicKey: PublicKey;
  }> {
    if (!(data instanceof Uint8Array)) {
      throw new Error('Data must be an instance of Uint8Array');
    }

    const response = (await this.sendRequest('sign', {
      data,
      display,
    })) as { publicKey: string; signature: string };
    const signature = bs58.decode(response.signature);
    const publicKey = new PublicKey(response.publicKey);
    return {
      signature,
      publicKey,
    };
  }
  // ...
}
```
It's not safe to assume the bytes represent a utf8 string, they may be binary or anything else. What would it do to pass no argument for `display`?

richard-fisher: If you don't pass any argument then your message gets displayed as hex but message signing still works.

Phantom docs say they support utf8 and hex.  By passing no param there is it defaulting to utf8?  If so then passing utf8 here might achieve the same behaviour.

jordansexton: Ah, I see. That's tricky. I guess utf8 is probably ideal then.

jordansexton:
Thank you!

PR #79, opened 2021-09-16
Update naming conventions and using native signAllTransactions from clover wallet

werlandy:
1. Update naming conventions
2. Using native signAllTransactions from clover wallet


jordansexton:
Looks good, thanks!

PR #86, opened 2021-09-18
[FIX] Yarn dependency command updated

GageBachik:
Yarn add command in the readme was missing a "\" causing it not to copy/paste right.


jordansexton:
Thanks!

PR #95, opened 2021-09-24
Loosen `@types/react` dependency

jnwng:
these packages ultimately are consumed by applications which will determine what version of `@types/react` to use. when the consumer has an incompatible version of `@types/react`, _both_ end up getting installed by `yarn` which causes some duplication of types, causing `tsc` to fail to build the consumer. see more information about the failure mode here: https://stackoverflow.com/questions/52399839/duplicate-identifier-librarymanagedattributes

i chose not to loosen the types in the starter packages, since those are technically the "consumers" of these packages. however, the packages here (`@solana/wallet-adapter-react`, `@solana/wallet-adapter-material-ui`, `@solana/wallet-adapter-ant-design`, `@solana/wallet-adapter-react-ui`) are not standalone packages, so packages that have dependencies here would be the ones establishing the proper version for `@types/react`. 

see this [PR](https://github.com/metaplex-foundation/metaplex/pull/473#issuecomment-925456319) for the sort of fix that has to be put in place by consumers otherwise

### DIFF
--- a/packages/core/react/package.json
+++ b/packages/core/react/package.json
@@ -25,13 +25,11 @@
         "@solana/web3.js": "^1.20.0",
         "react": "^17.0.2"
     },
-    "dependencies": {
-        "@types/react": "*"
-    },
     "devDependencies": {
         "@solana/wallet-adapter-base": "^0.6.0",
         "@solana/wallet-adapter-wallets": "^0.10.0",
         "@solana/web3.js": "^1.20.0",
+        "@types/react": "^17.0.24",
         "react": "^17.0.2"
     }
 }
--- a/packages/core/react/package.json
+++ b/packages/core/react/package.json
@@ -26,7 +26,7 @@
         "react": "^17.0.2"
     },
     "dependencies": {
-        "@types/react": "^17.0.16"
+        "@types/react": "*"
     },
     "devDependencies": {
         "@solana/wallet-adapter-base": "^0.6.0",
--- a/packages/ui/ant-design/package.json
+++ b/packages/ui/ant-design/package.json
@@ -30,7 +30,7 @@
         "react-dom": "^17.0.2"
     },
     "dependencies": {
-        "@types/react": "^17.0.16"
+        "@types/react": "*"
     },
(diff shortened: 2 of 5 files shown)


jordansexton:
Thanks for this PR and raising the issue.

The problem I see with using `*` for the version here is that these packages can then break on any breaking changes to `@types/react`. Do we even need a dependency for this? Since they're just types, can we use a dev dependency and then there's no conflict?

jordansexton:
Basically, the only benefit I can think of making this a dependency is that `@types/react` gets included, but if they are using React and TypeScript, they should already have that, so it's moot. And the only reason to even use those types is if they import as e.g. `@solana/wallet-adapter-react/src`, and it should just resolve to whatever version of the types they have.

jnwng:
yeah, that's a good point too — there's no reason a types package should be a direct dependency at all, it should definitely go into `devDependencies`. looking at a couple of other packages that similarly get "consumed" by other applications ([@material-ui/core](https://github.com/mui-org/material-ui/blob/686ec23b59a60105ad7b6ce51c29d4bc0aa9696e/packages/mui-core/package.json#L39-L48), [styled-components](https://github.com/styled-components/styled-components/blob/061736aa6dbaf8e21168db1c07bb2d061d9f2546/packages/styled-components/package.json#L90-L91)) they either use a combination of `peerDependencies` (which also makes sense) or just a `devDependency`.

i think removing this as a direct dependency makes the most sense though! i can update this PR later this evening.

jordansexton:
Yeah, peer dependencies that are mapped to `"*"` don't make _too_ much sense to me. The end user only actually has that dependency if they are using React with TypeScript, in which case they have that dependency regardless. If you want to make these dev deps, I'll merge and publish!

jnwng:
@jordansexton all ready to go! when this is published, i'll update https://github.com/metaplex-foundation/metaplex/pull/473 to update the `yarn.lock` on that end, too

jordansexton:
Published. Thanks!

jnwng:
thank you!

PR #99, opened 2021-09-27
Add TokenPocket wallet adapter

chendatony31:
TokenPocket will support solana in the next release after two days.

### DIFF
--- a/packages/wallets/tokenpocket/src/adapter.ts
+++ b/packages/wallets/tokenpocket/src/adapter.ts
@@ -154,7 +154,7 @@ export class TokenPocketWalletAdapter extends BaseMessageSignerWalletAdapter {
 
             try {
                 const { signature } = await wallet.signMessage(message);
-                return Uint8Array.from(signature);
+                return signature;
             } catch (error: any) {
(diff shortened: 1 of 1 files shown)


jordansexton:
Additionally, please allow commits from maintainers on your fork's master branch so I can update the PR as well. Thanks!

chendatony31:
Yes, as lots of dapps using Phantom as a login option. We are trying to be compatible with it at an early stage. 

I've added you into our fork's master, thank you Jordan.

jordansexton:
The problem isn't that you want to use the same API as Phantom, that's fine. The problem is that the Phantom adapter in this library contains workarounds for issues with Phantom, so you don't necessarily want to copy those parts. I think these things can be removed from the TokenPocket adapter if the TokenPocket wallet doesn't require those workarounds for connection problems.

chendatony31:
Got it, and have removed those code.

chendatony31:
@jordansexton  Hi we've released our new version which supports Solana dapp. You can log in dapps using Phantom option for now, if you want to try. https://tokenpocket.pro/

PR #105, opened 2021-09-30
Feature: Connection switch

kevinrodriguez-io:
This feature enables switching the connection.

Screen Shot 2021-09-30 at 01 15 57
Screen Shot 2021-09-30 at 01 16 03

Draft PR. Additional implementations on the way.

### DIFF
--- a/packages/core/react/src/ConnectionProvider.tsx
+++ b/packages/core/react/src/ConnectionProvider.tsx
@@ -1,6 +1,6 @@
 import { Connection, ConnectionConfig } from '@solana/web3.js';
-import React, { FC, ReactNode, useMemo } from 'react';
-import { ConnectionContext } from './useConnection';
+import React, { FC, ReactNode, useEffect, useMemo, useReducer } from 'react';
+import { ConnectionContext, ConnectionDispatchContext } from './useConnection';
 
 export interface ConnectionProviderProps {
     children: ReactNode;
@@ -13,7 +13,16 @@ export const ConnectionProvider: FC<ConnectionProviderProps> = ({
     endpoint,
     config = { commitment: 'confirmed' },
 }) => {
-    const connection = useMemo(() => new Connection(endpoint, config), [endpoint, config]);
-
-    return <ConnectionContext.Provider value={{ connection }}>{children}</ConnectionContext.Provider>;
+    const [connection, setConnection] = useReducer(
+        (_: Connection, newConnection: Connection) => newConnection,
+        new Connection(endpoint, config)
+    );
+    useEffect(() => {
+        setConnection(new Connection(endpoint, config));
+    }, [endpoint, config]);
+    return (
+        <ConnectionDispatchContext.Provider value={setConnection}>
+            <ConnectionContext.Provider value={{ connection }}>{children}</ConnectionContext.Provider>
+        </ConnectionDispatchContext.Provider>
+    );
 };
--- a/packages/core/react/src/useConnection.tsx
+++ b/packages/core/react/src/useConnection.tsx
@@ -6,7 +6,13 @@ export interface ConnectionContextState {
 }
 
 export const ConnectionContext = createContext<ConnectionContextState>({} as ConnectionContextState);
+export const ConnectionDispatchContext = createContext<React.Dispatch<Connection>>({} as React.Dispatch<Connection>);
 
 export function useConnection(): ConnectionContextState {
     return useContext(ConnectionContext);
+}
+
+export function useSetConnection(): (connection: Connection) => void {
+    const dispatch = useContext(ConnectionDispatchContext);
+    return (connection: Connection) => dispatch(connection);
 }
--- a/packages/starter/example/pages/index.tsx
+++ b/packages/starter/example/pages/index.tsx
@@ -5,6 +5,7 @@ import {
     WalletModalButton as AntDesignWalletModalButton,
     WalletModalProvider as AntDesignWalletModalProvider,
     WalletMultiButton as AntDesignWalletMultiButton,
+    ConnectionModalProvider as AntDesignConnectionModalProvider,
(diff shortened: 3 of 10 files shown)


packages/core/react/src/ConnectionProvider.tsx
```diff
@@ -13,7 +13,16 @@ export const ConnectionProvider: FC<ConnectionProviderProps> = ({
     endpoint,
     config = { commitment: 'confirmed' },
 }) => {
-    const connection = useMemo(() => new Connection(endpoint, config), [endpoint, config]);
-
-    return <ConnectionContext.Provider value={{ connection }}>{children}</ConnectionContext.Provider>;
+    const [connection, setConnection] = useReducer(
+        (_: Connection, newConnection: Connection) => newConnection,
+        new Connection(endpoint, config)
+    );
```
jordansexton, line 19: From https://reactjs.org/docs/hooks-reference.html#usereducer --

> useReducer is usually preferable to useState when you have complex state logic that involves multiple sub-values or when the next state depends on the previous one. useReducer also lets you optimize performance for components that trigger deep updates because you can pass dispatch down instead of callbacks.

What about this use case makes this a fit?

packages/core/react/src/useConnection.tsx
```diff
@@ -6,7 +6,13 @@ export interface ConnectionContextState {
...
+export const ConnectionDispatchContext = createContext<React.Dispatch<Connection>>({} as React.Dispatch<Connection>);
 
 export function useConnection(): ConnectionContextState {
     return useContext(ConnectionContext);
 }
+
+export function useSetConnection(): (connection: Connection) => void {
+    const dispatch = useContext(ConnectionDispatchContext);
+    return (connection: Connection) => dispatch(connection);
+}
```
jordansexton, line 18: This seems overly complicated. Why not just `const [connection, setConnection] = useState(() => new Connection(...))` in the provider and then pass `setConnection` into the context?

packages/starter/example/pages/index.tsx
```diff
@@ -26,107 +27,126 @@ import { useAutoConnect } from '../components/AutoConnectProvider';
...
+                            endpoint: clusterApiUrl(WalletAdapterNetwork.Mainnet),
+                            name: 'Mainnet',
+                        },
+                        {
+                            endpoint: clusterApiUrl(WalletAdapterNetwork.Devnet),
+                            name: 'Devnet',
+                        },
+                        {
+                            endpoint: clusterApiUrl(WalletAdapterNetwork.Testnet),
+                            name: 'Testnet',
```
jordansexton, line 51: Names like this should be an enum or some other typesafe structure so that child components are sure the input the correct values. The `connections` array could be part of the `ConnectionProvider` since it's not dependent upon UI, and the `connection` value can be an item in that array.

packages/ui/ant-design/src/ConnectionMenuItem.tsx
```diff
@@ -0,0 +1,19 @@
+import { Button, Menu, MenuItemProps } from 'antd';
+import React, { FC, MouseEventHandler } from 'react';
+
+interface ConnectionMenuItemProps extends Omit<MenuItemProps, 'onClick'> {
+    onClick: MouseEventHandler<HTMLButtonElement>;
+    connection: {
+        name: string;
+    };
```
jordansexton, line 8: This should be a named interface provided by the context.

packages/ui/ant-design/src/ConnectionModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+import React, { FC, MouseEvent, useCallback, useMemo, useState } from 'react';
+import { useConnectionModal } from './useConnectionModal';
+import { ConnectionMenuItem } from './ConnectionMenuItem';
+import { Connection } from '@solana/web3.js';
+
+export interface ConnectionModalProps extends Omit<ModalProps, 'visible'> {
+    connections: {
+        name: string;
+        endpoint: string;
+    }[];
```
jordansexton, line 12: Always good to export named interfaces for nested props if they need to be consumed by descendants and downstream devs.

packages/ui/ant-design/src/ConnectionModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const handleConnectionClick = useCallback(
+        (
+            event: MouseEvent<HTMLElement>,
+            connection: {
+                name: string;
+                endpoint: string;
+            }
+        ) => {
+            // TODO: Add a way to replace commitment.
+            setConnection(new Connection(connection.endpoint, { commitment: 'confirmed' }));
```
jordansexton, line 37: I'm not sure why this doesn't use the connection options provided by the existing ConnectionProvider context.

packages/ui/ant-design/src/WalletMultiButton.tsx
```diff
@@ -7,13 +7,15 @@ import { useWallet } from '@solana/wallet-adapter-react';
...
 import { useWalletModal } from './useWalletModal';
+import { useConnectionModal } from './useConnectionModal';
 import { WalletConnectButton } from './WalletConnectButton';
 import { WalletIcon } from './WalletIcon';
 import { WalletModalButton } from './WalletModalButton';
 
 export const WalletMultiButton: FC<ButtonProps> = ({ type = 'primary', size = 'large', children, ...props }) => {
     const { publicKey, wallet, disconnect } = useWallet();
     const { setVisible } = useWalletModal();
+    const { setVisible: setConnetionModalVisible } = useConnectionModal();
```
jordansexton, line 18: Typo in `Connetion`

jordansexton (changes requested):
Thanks for this! I think the implementation can be simplified significantly and the type safety of handling connection endpoints can be improved.

jordansexton:
Closing for now since there's been no response. Please comment if you want to reopen!

Issue #67, opened 2021-09-11
Add Vue support


jordansexton:
I'm not a user of Vue, but if anyone wants to PR `vue` and `vue-ui` packages based on the `react` and `react-ui` packages, I'd be happy to accept them.

goldcoders:
there should be a vue and svelte version of this....

danmt:
PRs are welcome @goldcoders

scorpnode:
Would like a vue adapter as well please.

jordansexton:
I don't use Vue or Svelte, but I'm happy to accept contributions to support them. I opened this issue to encourage contribution, so please open a PR if this is something you'd like to see. 🙇‍♂️

SilverSeva:
We need that wallet-adapter-vue please

lorisleiva:
Hi there 👋 

I've managed to use the wallet adapter using a Vue3 application.

Here's the code: https://github.com/lorisleiva/solana-vue3-counter/tree/9321e7c584906b3844d1c3eb1b69321f3ba8343d/app/src

It's a simple counter application but I've gathered most of the logic in `useWallet` and `useWorkspace` files.

I hope this helps. 🍀

danmt:
That looks nice @lorisleiva, are you willing to put a PR for it?

lorisleiva:
I could certainly try. 🙂 

I've got two freelance jobs on the go at the moment so I might be a bit slow though.

Is it okay for me to use TailwindCSS for the UI?

danmt:
I'd recommend doing the adapter first and then discuss the UI part, just so the PRs are a bit easier to digest.

I don't think there's a hurry, if somebody really needs it, they can do it themselves. Thanks a lot for stepping up and offer the solution you're using.

Regarding TailwindCSS, I'm really a fan of it but that means more dependencies. Since I'm not the code owner we may want to wait for @jordansexton feedback.

SilverSeva:
Hey, thats awesome @lorisleiva! Great Job

lorisleiva:
That's a good shout, I'll get started with the core adapter first to provide a more digestible PR.

I see what you mean about the extra dependency but I don't think Vue has any official UI libraries so there would be a need for extra dependencies anyway. With Tailwind I can make sure to purge any unused classes to make the final minified CSS super tiny. Anyway, that's my case for Tailwind but I'm more than happy to integrate it with something else. 😅

danmt:
[...]

PR #107, opened 2021-09-30
Add Vue core package

lorisleiva:
Hi there 👋 

This PR adds a new core package that supports Vue 3.

It works by providing two functions:
- `initWallet` that initialises the wallet store and should be called only once.
- `useWallet` that returns that store by reference so we can use its content anywhere in our application.

### Basic usage

```js
import { initWallet, useWallet } from '@solana/wallet-adapter-vue';

// Initialise the wallet store once.

// Use its content anywhere.
const { wallet, select, connect, disconnect, connected, sendTransaction } = useWallet();
select('Phantom');
connect();
sendTransaction(...);
```

### References

Analogously to the React core package, here are the options you can provide to the `initWallet` method.

```ts
interface WalletStoreProps {
    wallets: Wallet[];
    autoConnect?: boolean;                  // Default: false
    onError?: (error: WalletError) => void; // Default: (error: WalletError) => console.error(error)
    localStorageKey?: string;               // Default: "walletName"
}
```

And here is the definition of the wallet store you can access from anywhere in your Vue application once initialised.

```ts
interface WalletStore {
    // Props.
    wallets: Wallet[];
    autoConnect: boolean;

    // Data.
    walletName: Ref<string | null>,
    walletsByName: Ref<WalletDictionary>,
    wallet: Ref<Wallet | null>;
    adapter: Ref<Adapter | null>;
    publicKey: Ref<PublicKey | null>;
    ready: Ref<boolean>;
    connected: Ref<boolean>;
    connecting: Ref<boolean>;
    disconnecting: Ref<boolean>;

    // Methods.
    select(walletName: WalletName): void;
    connect(): Promise<void>;
    disconnect(): Promise<void>;
    sendTransaction(transaction: Transaction, connection: Connection, options?: SendTransactionOptions): Promise<TransactionSignature>;
    signTransaction(transaction: Transaction): Promise<Transaction>;
    signAllTransactions(transaction: Transaction[]): Promise<Transaction[]>;
    signMessage(message: Uint8Array): Promise<Uint8Array>;
}
```

### Why not provide/inject?
In React, it is common to use Context components and retrieve their content within their descendants. Whilst Vue has a similar concept via the `provide` and `inject` API, they aren't commonly used in the Vue community. Instead, it is more common to create a global store using the composition API and access its content from anywhere when needed, which is why I went for that approach in this PR.

### The `isolatedModules` flag
Note that I had to override the following TS option `"isolatedModules": false` inside the vue package in order to import from `@vue/runtime-core` (See https://github.com/solana-labs/wallet-adapter/issues/67#issuecomment-929386709).

### Manual testing
I manually tested this package locally [in my solana test app](https://github.com/lorisleiva/solana-vue3-counter/tree/feature/test-local-vue-adapter) by installing it as a symlink — i.e. using `"@solana/wallet-adapter-vue": "file:../../wallet-adapter/packages/core/vue"`. Everything worked properly and I was able to sign and send some transactions using my Phantom wallet.

However, I noticed that components using the references (`Ref`) provided by the store did not update when the references changed. I can see the references updating in real time in the Vue dev tools but for some reason the component itself does not re-render itself. When I call `this.$forceUpdate` on the component, I can see all the latest values being updated properly. I believe this is just a quirk from installing the package locally using a symlink but I thought I'd mention it in case this issue ring a bell to someone.


packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (! adapter.value) throw newError(new WalletNotSelectedError());
+        if (! connected.value) throw newError(new WalletNotConnectedError());
+        return await adapter.value.sendTransaction(transaction, connection, options);
+    };
+
+    // Sign a transaction if the wallet supports it.
+    const signTransaction = async (transaction: Transaction) => {
+        if (! adapter.value) throw newError(new WalletNotSelectedError());
+        if (! connected.value) throw newError(new WalletNotConnectedError());
+        if (!('signTransaction' in adapter.value)) throw newError(new OperationNotSupportedByWalletError());
```
jordansexton, line 174: Throwing a runtime error here is something the React API design intentionally avoids. The optional methods will be undefined if the wallet doesn't support them. This produces type safety for the developer that calls them and avoids an error the user sees.

packages/core/vue/src/errors.ts
```diff
@@ -0,0 +1,9 @@
+import { WalletError } from '@solana/wallet-adapter-base';
+
+export class WalletNotSelectedError extends WalletError {
+    name = 'WalletNotSelectedError';
+}
+
+export class OperationNotSupportedByWalletError extends WalletError {
+    name = 'OperationNotSupportedByWalletError';
+}
```
jordansexton, line 9: This shouldn't be needed if the operations are typesafe.

packages/core/vue/src/useLocalStorage.ts
```diff
@@ -0,0 +1,19 @@
+import { customRef, Ref } from '@vue/reactivity';
+
+export function useLocalStorage<T> (key: string, defaultValue: T | null = null): Ref<T | null> {
+    return customRef<T | null>((track, trigger) => ({
+        get: () => {
+            track();
+            const value = localStorage.getItem(key);
+            return value ? JSON.parse(value) : defaultValue;
```
jordansexton, line 8: We don't do this yet in the React design, but we should -- wrap `JSON.parse` in try/catch and return the default value if it fails. This way users won't get an uncaught runtime error.

packages/core/vue/src/useLocalStorage.ts
```diff
@@ -0,0 +1,19 @@
...
+        },
+        set: value => {
+            if (value === null) {
+                localStorage.removeItem(key);
+            } else {
+                localStorage.setItem(key, JSON.stringify(value));
```
jordansexton, line 14: Should probably do the same here, since `JSON.stringify` will throw for circular references.

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+type WalletDictionary = { [key: string]: Wallet };
+
+export interface WalletStore {
+    // Props.
+    wallets: Wallet[];
+    autoConnect: boolean;
+
+    // Data.
+    walletName: Ref<string | null>,
+    walletsByName: Ref<WalletDictionary>,
```
jordansexton, line 23: Let's continue to use the available `WalletName` enum here for type safety.

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+    localStorageKey = 'walletName',
+}: WalletStoreProps): void => {
+    const walletName = useLocalStorage<string>(localStorageKey);
+    const wallet = ref<Wallet | null>(null);
+    const adapter = ref<Adapter | null>(null);
+    const publicKey = ref<PublicKey | null>(null);
+    const ready = ref<boolean>(false);
+    const connected = ref<boolean>(false);
+    const connecting = ref<boolean>(false);
+    const disconnecting = ref<boolean>(false);
```
jordansexton, line 66: I'm not sure how this applies to Vue, but the way this state is handled has changed in the React package:
https://github.com/solana-labs/wallet-adapter/blob/581cb52a41ad7a255542de39115d051b0833e57c/packages/core/react/src/WalletProvider.tsx#L42

This allows all related state updates to happen at once, which prevents inconsistent state, rerenders, and race conditions. I see you are updating the state below in one place which is great.

It may be worth looking at how the `connecting`/`disconnecting`/`working` values are handled now also to avoid rerenders or `useEffect` invocations (again with the caveat that I don't know how relevant to Vue this is).

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+
+export const useWallet = (): WalletStore => walletStore;
+
+export const initWallet = ({
+    wallets,
+    autoConnect = false,
+    onError = (error: WalletError) => console.error(error),
+    localStorageKey = 'walletName',
+}: WalletStoreProps): void => {
+    const walletName = useLocalStorage<string>(localStorageKey);
```
jordansexton, line 59: Can't this be `null`?

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+    const walletsByName = computed<WalletDictionary>(() => {
+        return wallets.reduce((walletsByName, wallet) => {
+            walletsByName[wallet.name] = wallet;
+            return walletsByName;
+        }, {} as WalletDictionary);
+    });
+
+    // Update the wallet and adapter based on the wallet provider.
+    watch(walletName, (): void => {
+        wallet.value = walletsByName.value?.[walletName.value as string] ?? null;
```
jordansexton, line 77: What does `walletName.value as string` do? Isn't the type of `walletName` known?

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (adapter.value) {
+            ready.value = adapter.value.ready;
+            publicKey.value = adapter.value.publicKey;
+            connected.value = adapter.value.connected;
+        } else {
+            ready.value = false;
+            publicKey.value = null;
+            connected.value = false;
+        }
+    }, { immediate:true });
```
jordansexton, line 88: Will you explain what this flag does for me?

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (! adapter.value) return;
+        ready.value = adapter.value.ready;
+        publicKey.value = adapter.value.publicKey;
+        connected.value = adapter.value.connected;
+    };
+    watchEffect(onInvalidate => {
+        if (! adapter.value) return;
+        adapter.value.on('ready', onReady);
+        adapter.value.on('connect', onConnect);
+        adapter.value.on('error', onError);
```
jordansexton, line 109: What about `disconnect`?

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+        onInvalidate(() => {
+            if (! adapter.value) return;
+            adapter.value.off('ready', onReady);
+            adapter.value.off('connect', onConnect);
+            adapter.value.off('error', onError);
+        });
+    });
+
+    // Helper method to return an error whilst using the onError callback.
```
jordansexton, line 118: Smart! We should probably do this elsewhere.

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (! connected.value) throw newError(new WalletNotConnectedError());
+        if (!('signMessage' in adapter.value)) throw newError(new OperationNotSupportedByWalletError());
+        return await adapter.value.signMessage(message);
+    };
+
+    // If autoConnect is enabled, try to connect when the adapter changes and is ready.
+    watchEffect(async (): Promise<void> => {
+        if (! autoConnect || ! adapter.value || ! ready.value || connected.value || connecting.value) return;
+        await connect();
+    });
```
jordansexton, line 198: The implementation has changed significantly in the React package:
https://github.com/solana-labs/wallet-adapter/blob/581cb52a41ad7a255542de39115d051b0833e57c/packages/core/react/src/WalletProvider.tsx#L70-L87

Several types of bugs were eliminated. This also shouldn't call the `connect` method because that can open a popup if the wallet isn't available.

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    walletStore = {
+        // Props.
+        wallets,
+        autoConnect,
+
+        // Data.
+        walletName,
+        walletsByName,
```
jordansexton, line 207: These shouldn't need to be exposed.

jordansexton (commented):
Overall looks good so far! I'll try to run it in your example package with a UI to see how it feels.

lorisleiva:
Thanks Jordan! I'll go through each of these points gradually. 🙂

packages/core/vue/src/useWallet.ts
lorisleiva: In Vue:

- `watchEffect(fn)` is the equivalent of `useEffect(fn)` but you cannot provide a list of variables to listen for, instead it will automatically recognise any reactive variables inside the function and call that method every time one or more of these variables change during a given tick. Most of the time, this is exactly what we need and so we let Vue handle the list of vars for us.
- If you want to explicitly tell Vue what you're watching, you can use `watch(var, fn)` or `watch([vars], fn)`.
- However, by default, the `watch` method will only be called on future updates and not immediately after defining it (as `watchEffect` and `useEffect` do).
- Therefore, you need to provide that flag to tell it to trigger immediately.
- In conclusion, `watch([vars], fn, { immediate:true })` is equivalent to `useEffect(fn, [vars])`.

packages/core/vue/src/useWallet.ts
lorisleiva: I had issues with the `onDisconnect` callback when defined as `() => walletName.value = null`.

When using the `autoconnect` flag, it would cause the whole thing to disconnect on refresh.

The flow was:
- Connect your wallet using the `autoconnect` flag.
- Refreshing the page would call `onDisconnect` on the adapter.
- On reload, the walletName would be empty and so we wouldn't be able to autoconnect.

Additionally, we explicitly set `walletName.value = null` in the `finally` scope of the `disconnect` method so I thought this was perhaps redundant.

I'm happy to give it another go though, did you come across that issue with React?

packages/core/vue/src/useWallet.ts
lorisleiva: It can, the `null` value is defined in the return type of the signature:

```ts
useLocalStorage<T> (key: string, defaultValue: T | null = null): Ref<T | null>;
```

I wasn't sure how to make this more explicit whilst making sure we allow for null values in the signature.

packages/core/vue/src/useWallet.ts
lorisleiva: The type of this variable

```ts
const walletName = useLocalStorage<string>(localStorageKey);
```

Resolves to `Ref<string | null>` so `walletName.value` resolves to `string | null` and TS was complaining.

I should probably mark the `walletName` explicitly like 


```ts
const walletName:Ref<string | null> = useLocalStorage<string>(localStorageKey);
```

packages/core/vue/src/useWallet.ts
lorisleiva: I can do something similar but I don't think this will change anything in Vue because the code will be executed within the same DOM update cycle (a.k.a a tick).

But I can of course group it in a state object so that it is at least consistent with the logic of the other packages.

---

On a similar note, I had issues with reactivity if I didn't provide the `Refs` directly in the `WalletStore` but I'll try again to give the real values instead of the `Ref` variables.

To give more context, in Vue `Refs` are equivalent to `useState` but the API is quite different:

```js
// In React.
const [foo, setFoo] = useState('defaultFoo'); 
foo; // Access
setFoo('newFoo'); // Edit

// In Vue.
const foo = ref('defaultFoo');
foo.value; // Access
foo.value = 'newFoo'; // Edit
```

So ultimately, by returning the `Ref` we are also giving the equivalent of the `setFoo` method which I don't think is what we want. I'll see what I can do. 🙂

lorisleiva: Okay I've done some tests and If I don't return `Refs` reactivity simply does not work in Vue for primitive values.

An alternative would be to wrap the entire store in the `reactive` function, but that would still allow us to update any value inside it (including functions) and would differ from what most people are doing when creating Vue 3 libraries.

Also, the nice thing about returning `Refs` is that they are perfectly understood by Vue's components and their template knows how to handle them even without using `.value`.

packages/core/vue/src/useWallet.ts
lorisleiva: I've explicitly defined it as:

```js
const walletName: Ref<WalletName | null> = useLocalStorage<WalletName>(localStorageKey);
```

packages/core/vue/src/useLocalStorage.ts
lorisleiva: All done. Am I right to assume we want to fail silently if that happens?

Btw, there is already an implementation of `useLocalStorage` in the `@vueuse/core` npm package but I assumed we didn't want an extra dependency just for that. Let me know, if you'd rather I use it instead.

https://vueuse.org/core/useLocalStorage/

jordansexton: I think it's okay to just fail to add the item. We might want to `console.error` though, wdyt?

packages/core/vue/src/useWallet.ts
jordansexton: Ah, okay. The behavior you had doesn't necessarily need to be changed, as long as nothing can react to those individual value changes before they are all changed.

packages/core/vue/src/useWallet.ts
jordansexton: Can you help me understand what a Ref does in Vue?

packages/core/vue/src/useWallet.ts
jordansexton: > Refreshing the page would call onDisconnect on the adapter.

This one is interesting, and seems likely to be the cause. I wonder why that is, as I don't think React does this?

jordansexton: A _possible_ workaround could be to add an `onbeforeunload` event listener to unregister the `disconnect` event handler so this doesn't happen. But we do need a `disconnect` event listener, since adapters can disconnect for various reasons outside the control of the app.

packages/core/vue/src/useLocalStorage.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function useLocalStorage<T> (key: string, defaultValue: T | null = null): Ref<T | null> {
+    return customRef<T | null>((track, trigger) => ({
+        get: () => {
+            track();
+            const value = localStorage.getItem(key);
+            try {
+                return value ? JSON.parse(value) : defaultValue;
+            } catch (error) {
+                return defaultValue;
```
jordansexton, line 11: I'm thinking we should `console.warn` here, and `console.error` in the setter

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,264 @@
...
+    WalletNotConnectedError,
+    WalletNotReadyError,
+    WalletError,
+} from '@solana/wallet-adapter-base';
+import { Wallet, WalletName } from '@solana/wallet-adapter-wallets';
+import { useLocalStorage } from './useLocalStorage';
+import { WalletNotSelectedError } from './errors';
+
+type Adapter = ReturnType<Wallet['adapter']>;
+type WalletDictionary = { [key: string]: Wallet };
```
jordansexton, line 14: The key here should be a `[name in WalletName]` if possible

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,264 @@
...
+    // Send a transaction using the provided connection.
+    const sendTransaction = async (transaction: Transaction, connection: Connection, options?: SendTransactionOptions) => {
+        if (! adapter.value) throw newError(new WalletNotSelectedError());
+        if (! connected.value) throw newError(new WalletNotConnectedError());
+        return await adapter.value.sendTransaction(transaction, connection, options);
+    };
+
+    // Sign a transaction if the wallet supports it.
+    const signTransaction = computed(() => {
+        if (! (adapter.value && 'signTransaction' in adapter.value)) return undefined;
```
jordansexton, line 198: Nice, this is pretty elegant

packages/core/vue/src/useWallet.ts
```diff
@@ -0,0 +1,264 @@
...
+        }
+    });
+
+    // Sign multiple transactions if the wallet supports it
+    const signAllTransactions = computed(() => {
+        if (! (adapter.value && 'signAllTransactions' in adapter.value)) return undefined;
+        return async (transactions: Transaction[]) => {
+            if (! connected.value) throw newError(new WalletNotConnectedError());
+            // @ts-ignore
+            return await adapter.value.signAllTransactions(transactions);
```
jordansexton, line 212: Would `adapter.value.signAllTransactions!` work instead of ts-ignore?

packages/core/vue/src/useWallet.ts
jordansexton: Thanks for the detailed explanation!

jordansexton (commented):
Looks great, this seems nearly ready to merge

packages/core/vue/src/useLocalStorage.ts
lorisleiva: Good shout! 👍

packages/core/vue/src/useWallet.ts
lorisleiva: Of course! The following code in React:

```js
const [foo, setFoo] = useState('defaultFoo');
foo; // Access
setFoo('newFoo'); // Edit
```

Can be achieved in Vue this way:

```js
const foo = ref('defaultFoo')
foo.value // Access
foo.value = 'newFoo'; // Edit
```

This ensures that reactivity is kept even when dealing with primitive values as they are wrapped in a reactive object.

Alternatively, one can use the `reactive` method provided by Vue to make an object and its properties reactive. However, if you then destructure that object, reactivity will be lost on primitive properties.

```js
const state = reactive({
    foo: 'defaultFoo',
    bar: { some: 'object' },
})
state.foo // Access
state.foo = 'newFoo'; // Edit

const { foo, bar } = state;
// `foo` is no longer reactive
// `bar` is now a `Reactive` object

state.foo = 'newFoo -- edited after destructuring';
state.bar.some = 'object -- edited after destructuring';

console.log(foo); // Outputs: newFoo
console.log(bar); // Outputs: { some: 'object -- edited after destructuring' }
```

When dealing with a `reactive` object, you can also destructure all of its properties into `Refs` using the `toRefs` method. This will ensure even primitive properties are reactive but they will be wrapped in a `Ref` object and so you'll need to access them using the `.value` attribute.

```js
const state = reactive({
    foo: 'defaultFoo',
    bar: { some: 'object' },
})

const { foo, bar } = toRefs(state);
state.foo = 'newFoo -- edited after destructuring';

console.log(foo.value); // Outputs: newFoo -- edited after destructuring
```

So in the end, when dealing with primitive values (or any values to stay consistent within our code), using `Refs` in Vue is the go-to approach. Additionally, when handling `Refs` inside templates (the equivalent of JSX in Vue), Vue will properly recognised `Refs` variables so you can access and edit them directly without using the `.value` attribute.

packages/core/vue/src/useWallet.ts
lorisleiva: It worked! I understand what was happening now.

Just like in React, the `onInvalidate` callback (that disables all adapter listeners) defined in the `watchEffect` method will be called when the component is unmounted. From the [documentation](https://v3.vuejs.org/guide/reactivity-computed-watchers.html#side-effect-invalidation):

> This invalidation callback is called when:
>
> - the effect is about to re-run
> - the watcher is stopped (i.e. when the component is unmounted if watchEffect is used inside setup() or lifecycle hooks)

However, in the `useWallet.ts` file, we're not inside a particular component as we're defining a global store. In React, it is more common to define these in `ContextProviders` which are components themselves giving data to their descendants. In Vue, we have something similar with `provide` and `inject` but it's not quite the same and is rarely used in the community.

So we can do two things to go around this:
- Since a component will ultimately be calling the `initWallet` method, we could ask for a component as a parameter and then give it to the `watchEffect` method so it hooks itself to whenever that component unmounts. The only issue with this approach is we're dirtying a little bit the API by having to ask the component to add itself to the parameters.
- Do what you suggested and add a `beforeunload` window listener that invalidates the adapter listeners.

Both of these make sure adapter listeners are invalidated before the `onDisconnect` callback is triggered by the wallet itself upon refresh.

I went for the second approach to keep the API cleaner and less coupled to components. Do let me know if you'd rather I went for the first approach (or any other approach).

Thanks for your help in this it was really bugging me and I feel much better now that I understand what's happening haha.

lorisleiva:
Thanks! I've been through all your points. Almost there! 💪 😊

Bigandy4climate:
Hey @lorisleiva, it says your package is not yet published on npm :(

packages/core/vue/src/useWallet.ts
jordansexton: My concern about the `beforeunload` event is that I don't know if it could result in a race condition. For example, if the wallet's `disconnect` is triggered by `beforeunload` or something, and that event handler is called before the one here. I'm not saying we shouldn't do this, but it's something to be aware of, and could be nasty to debug if reported by users.

jordansexton: We could maybe add a `cleanup`/`teardown` method to the store, and then a top level component unmounting could call it? It's guaranteed that devs will forget to do this though...

packages/core/vue/src/useWallet.ts
jordansexton: Thanks a lot, that makes perfect thanks.

jordansexton: It seems like the way Vue works is kinda like MobX then. It tracks access of reactive values when they are dereferenced by getting/setting `ref.value`, so it knows what functions to call and what component trees to render.

Bigandy4climate:
Hey unsure where to post this comment but I got 'unexpected token "export"' upon running this line :(


import { initWallet, useWallet } from '@solana/wallet-adapter-vue'

Bigandy4climate:
Oh yeah I am on Nuxt :( Is that relevant info?

lorisleiva:
I've not played with Nuxt before but yeah I think you might need to tweak your webpack settings.

Bigandy4climate:
In tsconfig, target was ES2018, switched it to ESNext with no result :cry:

lorisleiva:
Sorry I'm not sure I'm very qualified to help here I struggle so much when configuring bundlers/compilers.

All I can say is `unexpected token "export"` usually means your compiler wants modules to use `module.export = {}` and does not accept modules that export this way:

```js
export default { foo: 'bar' };
export const myAdditionalVariable = 42;
```

What I usually do when struggling with configuring compilers is start with a brand new application and only add the dependency that's causing me an issue to understand what I need to change to make it work. Then I bring my findings to my current application. I hope this helps somehow. 🍀

Bigandy4climate:
https://stackoverflow.com/a/69516471/10620200

Bigandy4climate:
This is weird actually because Nuxt does support ESModules in the src so why not node modules too?

Bigandy4climate:
Hi @lorisleiva, 

What does the select method do? 
Also, intellisense for type Wallet is very inconsistent on my install (randomly changing), while I get errors claiming that Wallet objects are self-referring objects 😕

lorisleiva:
Hi 👋 `select` selects a type of wallet (such as Phantom) by providing a `WalletName` (e.g. `Phantom`).

I'm not sure about the intellisense issues though.

Issue #49, opened 2021-09-03
account changed event


TimonPeng:
Should we implement a new event to make it easier for users to switch wallet?

jordansexton:
Yeah, I think this would be a really nice feature. Maybe we can start with making a list of which wallets support switching accounts, and what interfaces and events they provide to do it?

TimonPeng:
I think almost all extensions and web wallets can support that, except apps and ledger(hardware)? under my observation.

jordansexton:
Ledger supports multiple accounts too. The issue (and the work to implement this) is adapting each of their different interfaces and behaviors to support this in a generalized way, which requires an in-depth understanding of what each one does.

jordansexton:
Basically, if someone wants to implement this, the first step is to go through the wallet adapters we have and do a writeup of

1. Is account switching supported?
2. How is it supported (is it a method we call, is it an array of accounts, what format are they in, etc.)
3. What internal events and interfaces do we need to adapt?

And then, what's the common denominator of all these?

jordansexton:
@TimonPeng is this something you want to try to make a PR for?

TimonPeng:
- Coin98 https://docs.coin98.com/products/developer-guide/solana-dapps-integration#list-of-events

TimonPeng:
I'm not sure which one is best

```
const onChange = (accounts) => setAccount(accounts.length > 0 ? accounts[0] : null);
```

```
const onChange = (account) => setAccount(account);
```

```
const onChange = () => {
  const { publicKey } = adapter;
  setAccount(publicKey);
};
```

cryptogosu:
@TimonPeng Are you working on this at this time?

spencergray96:
Any update on this?

jordansexton:
Just got back from vacation! I'll be reviewing the work in #109 soon.

jordansexton:
Iceboxing this issue, but I'll follow up with the work in #109

samia-invozone:
How can i detect in my React App if user switches his wallet from Phantom extension?

Mahdibayat:
same issue, need switch account event

glitch-txs:
have this been resolved somewhere? We would like to add this event to the WalletConnect v2 protocol but would like to follow an already established standard.

PR #109, opened 2021-10-01
onAccountChange

jacobcreech:
Adding onAccountChange base function. 

Still testing and adding more implementation. Related to https://github.com/solana-labs/wallet-adapter/issues/49


cryptogosu:
@jordansexton Would love to know what you think so far, as well if there are any obvious misses in implementation

Still debugging an issue with the event listener not getting the event when expected, but getting there.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -235,11 +240,13 @@ export const WalletProvider: FC<WalletProviderProps> = ({
             adapter.on('ready', onReady);
             adapter.on('connect', onConnect);
             adapter.on('disconnect', onDisconnect);
+            adapter.on('onAccountChange', onAccountChanged);
```
jordansexton, line 243: This naming convention should probably be `change` for consistency. How to implement the change or what changes can be up to the adapter, and then it's more general -- this component will update it's state from anything in the adapter that changed.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -112,6 +112,11 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     // Handle the adapter's disconnect event
     const onDisconnect = useCallback(() => setName(null), [setName]);
 
+    const onAccountChanged = useCallback(async () => {
+        await disconnect();
+        await connect();
+    }, []);
```
jordansexton, line 118: I don't think this will produce a very good UX. Disconnecting and reconnecting the wallet isn't necessary if the change has already occurred (for example, indicating that a new `publicKey` is available).

packages/wallets/coin98/src/adapter.ts
```diff
@@ -140,4 +147,13 @@ export class Coin98WalletAdapter extends BaseSignerWalletAdapter {
         }
         return signedTransactions;
     }
+
+    private _accountChanged = () => {
+        const wallet = this._wallet;
+        console.log(wallet);
+        if (wallet) {
+            wallet.off('accountsChanged', this._accountChanged);
```
jordansexton, line 155: Why does this remove the event listener? If it's expecting a listener to disconnect and reconnect, that's prescribing too much behavior, and it's going to produce an inconsistent state in the adapter (no listener) if that listener doesn't disconnect.

jordansexton (changes requested):
I think this can be generalized to give the adapter more flexibility over how to handle different types of changes that can come from wallets, and to produce a more consistent UX/devx.

Arrowana:
Any help needed for that thing? Feels like a great to have.
Also for discrepancies between wallets, should the wallet adapter create proposals as to how things should work and wallet must align?
Seems quite convoluted to have to handled peculiar differences for each wallet

jordansexton:
> should the wallet adapter repo have proposals as to how things should work and wallet must align?

I think this is a good idea. It's just hard to get wallets to prioritize features unless users really care. But if we had notable dapps that lobbied for the feature, that might get them interested.

jacobcreech:
Closing for now

clou-neon:
hi, neon.game here 🖖
would love to subscribe for address inside of phantom wallet change, and update UI accordingly without forcing user to hit **reload page**

PR #111, opened 2021-10-04
add support for slope wallet adapter to sign messages

Jeffrieh:
Slope wallet now supports the signing of messages, i've updated the slope adapter accordingly, please have a look, and let me know if some extra changes have to be made!


jordansexton:
Looks good, thanks!

PR #114, opened 2021-10-07
Add Svelte core package

silvestrevivo:
This PR contains the core package for the Svelte support. 

Initially, the approach is done using different Svelte stores, being `useWallet` the one which is exposed to be used among the components. Initially `useWallet` has been already tested using SvelteKit. Later on we will need to create a component assuming the same function than `ConnectionProvider and WalletProvider` for React. At that point would be ideal to discuss different solutions for SSR and SPA.

## Usage with SvelteKit
```javascript
import { initWallet, useWallet, useWalletName, useWalletAdapter, watchWalletName, watchAdapter, destroyAdapter, autoConnectWallet } from '@solana/wallet-adapter-svelte';

// Initialize the wallet store once.

//add reactivity to listen to the watchers
$: watchWalletName($useWalletName.walletName, localStorageKey);
$: watchAdapterClient($useWalletAdapter.adapter);
$: autoConnectWalletClient({
	adapter: $useWalletAdapter.adapter,
	walletName: $useWalletName.walletName
    });
onDestroy(() => {
   if ($useWalletAdapterStore?.adapter) {
	destroyAdapterClient($useWalletAdapterStore.adapter);
	}
  });

// Use its content anywhere.
const { wallet, select, disconnect, connected, sendTransaction } = useWallet();
select('Phantom');
sendTransaction(...);
```


danmt:
Here's a screenshot of the compilation error I got.

error screenshot

silvestrevivo:
@jordansexton @danmt @lorisleiva thanks so much for the review, I really appreciate the time invested checking out the code. I will try to explain all the concerns about the amount of stores and exported functions. You can follow this example with an Anchor implementation https://github.com/silvestrevivo/solana-svelte-counter to understand the code better. 

Svelte has a state management based on stores like React, but instead to have 1 store per app, we create so many stores as we need. The reason to have 3 stores defining the data and a finally store called useWallet is as follow: 

- 3 Stores
My first attempt before to open a PR, was to include everything in one single store as the Vue implementation. The Vue composition API can have a store with reactive functions inside of it (watch), and in Svelte you need to listen to the reactivity in the components itself. As fas as I know, is not possible to create reactivity in a TS file. When I tried to apply reactivity listen just a property of the store (for example, `adapter`), it was impossible to isolate the property and instead it was listening to all the changes in the store. This was giving me problems in the way that the watch function for adapter, was being triggered in situations not desirable. 

Later on, I saw that `adapter` and `walletName` were not necessary in the client and I decided to create `3 writable stores` to isolate the reactivity for each case. Doing this like this, the watcher are being triggered when they have to.  That's why I need to export those functions to be used in a Svelte component. As you can see in the example attached, there is a component called `Solana.svele` which is doing the same than `ConnectionProvider ` and `WalletProvider` but just in one component, and there I solve all the logic supporting the `useWallet` store. This last store, is a `derived` store, and that means is only reading and reacting to the changes of the other 3. 

it sounds bit confusing maybe, but Stores in Svelte can not contain functions, the can be as much as you need, and reactivity is managed in a Svelte component. Never outside of it. 

- SSR
Just a thing to take in consideration if you look to the `Solana.svelte` component. Currently the `solana-wallet-adapter` library is giving me errors asking for the `Buffer` and `process` packages to run. To avoid those issues, following the documentation from SvelteKit, I need to import those packages inside an `onMount` function. Please think that all the tests are made in SvelteKit, and that means that this implementation works for SPA and SSR. If in the future SvelteKit find a solution to load node packages, the `Solana.svelte` component could be simplified. 

Please let me know what you think about this. Here below you can find a video of the demo project attached to show that the adapter is working. :)

https://user-images.githubusercontent.com/10574041/137030854-e83b6912-1d8f-4b75-8024-1dff6848c383.mov

silvestrevivo:
> > Currently the `solana-wallet-adapter` library is giving me errors asking for the `Buffer` and `process` packages to run.
> 
> I think you may be able to fix this with different build-time configuration rather rely on loading at runtime. This may be of use: https://github.com/solana-labs/wallet-adapter/blob/cd92692b47dd4cd86db330a3ffe310a3e837c70f/FAQ.md#webpack--gatsby

SvelteKit works with Vite, not with Webpack. What I know till now, there is no a solution like that at this moment with Vite. Do you think I can use that code from Gatsby in this implementation?

jordansexton:
> there is no a solution like that at this moment with Vite

Vite has no way to handle polyfills or Node.js globals?

silvestrevivo:
> > there is no a solution like that at this moment with Vite
> 
> Vite has no way to handle polyfills or Node.js globals?

I didn't find them. This issue is nor affecting to the `useWallet` file, is just in the Svelte implementation. I was dealing with this hours and hours to find a solution, and the only thing I could is to follow the SvelteKit docs and apply those imports in the onMount() method. As I mention in the explanation, this is valid for SvelteKit, and that means we can use this for SPA or SSR.

jordansexton:
So looking at this PR overall, I think most of it is great and the functionality is all there. What I'd like to change is how the external API is treated, and that's going to imply some amount of refactoring. Internally, I don't have an opinion on whether if it uses multiple logical stores for behavior separation or performance improvement or whatever is idiomatic -- if it needs to.

But externally, it should present one API to do what developers need to do, with as small a surface as possible. To me, this means having all the data in one store, as the React, Angular, and Vue packages do, and it means not exposing internal functions, including `watchWalletName`, `watchAdapter`, `destroyAdapter`, and `autoConnectWallet`.

These are all internal implementation details. As a whole, just like the other core packages, this package should be self-contained -- a Svelte application, even one not using the Svelte UI components you've created, should be able to use this package with minimal setup code and minimal API knowledge.

Looking at https://github.com/silvestrevivo/solana-svelte-counter/blob/master/app/src/lib/Solana.svelte -- this has a lot of setup and teardown code. Could these things be encapsulated into this package to make them usable with fewer lines of code and less conceptual overhead?

silvestrevivo:
@jordansexton thank you so much for checking out the code and you are right. This should be 'isolated' in the way you are mentioning. That's why I refactored the code and now I am just exporting the next things:

-- useWallet()
-- initWallet()
-- destroyAdapter()

The last one we can consider it as a trade-off with the Svelte implementation. I was yesterday talking with one of the Svelte maintainers and is not possible to unsubscribe outside a Svelte component. You can checkout in the repo example the implementation and now is very simple. https://github.com/silvestrevivo/solana-svelte-counter/blob/master/app/src/lib/Solana.svelte As you can see, we have 40% less code and is readable and easy to maintain.

Regarding the node polyfills on SvelteKit, I've tried several options but they didn't work. I asume that importing thing inside of the onMount() function, is something that SvelteKit developers are used to do importing for example packages containing the `window` object in SSR applications. In any case, this is something inherit to Svelte and not to the wallet adapter. If in the future SvelteKit solves this issue, the implementation becomes easier and is not affecting to the wallet adapter. 

Please let me know what you think. Thanks!!!!

jordansexton:
Okay, definitely getting there, but here's what I'm saying -- if doing all the non-visual setup code in https://github.com/silvestrevivo/solana-svelte-counter/blob/master/app/src/lib/Solana.svelte is required for any Svelte app, then it should be part of this package.

> is not possible to unsubscribe outside a Svelte component

I can definitely understand that, we essentially do the same thing in React context components that can be mounted/unmounted but don't render anything. Can we do this with Svelte?

Basically, take away all this overhead for devs and give them the smallest encapsulated API. Once this package is so easy that so anyone can drop it into their Svelte application in just a few lines of code, it will be there.

jordansexton:
For example, this function: 
https://github.com/silvestrevivo/solana-svelte-counter/blob/6602bf7bd97f52c09f757d2b18f065f8d2e07aa2/app/src/lib/Solana.svelte#L12-L17

```
	onMount(async () => {
		const module = await import('$utils/useWallet');
		initWallet = module.initWallet;
		destroyAdapter = module.destroyAdapter;
		useWallet = module.useWallet;
	});
```

If any dapp needs to do that, then just pull it into this package and export it as a one-line function.

silvestrevivo:
@jordansexton I'll try to answer here to the last two comments. 

Definitely we need a component responsible of the injection of the wallet, and like in the React implementation, we will need a component responsible for the connection. Let's call them `<WalletProvider />` and `<ConnectionProvider />`. In the documentation we need to explain how to implement them in the project. In this case for SvelteKit, in the `__layout.svelte` file, developers need to implement it like this. This ConnectionProvider would be for Anchor. 

```
<script lang="ts">
	import { onMount } from 'svelte';
	import WalletProvider from '$lib/WalletProvider.svelte';
	import ConnectionProvider from '$lib/ConnectionProvider.svelte';
	import idl from '../../../target/idl/solana_svelte_counter.json';
	import { Buffer } from 'buffer';
	import process from 'process';

	globalThis.Buffer = Buffer;
	globalThis.process = process;

	const localStorageKey = 'walletAdapter';
	const network = 'http://127.0.0.1:8899';

	let wallets;

	onMount(async () => {
		const { getPhantomWallet } = await import('@solana/wallet-adapter-wallets');
		const walletsMap = [getPhantomWallet()];
		wallets = walletsMap;
	});
</script>

<WalletProvider {localStorageKey} {wallets} />
<ConnectionProvider {network} {idl} />
<div>
	<slot />
</div>
``` 

Regarding the second question, is not possible to write a Svelte life-cycle method outside a Svelte component. At this moment and without a solution for the node polyfills, we need to import the `useWallet` store like this. If in the future this is solved (I will open a Issue on SvelteKit soon), we can update the components. It will have no impact even after the update.  

Please, look at the demo example before to push here something and being sure that you like the solution. 

We are almost there. Thanks!!!

jordansexton:
You got it man! Let me know when you're satisfied and I'll merge :)

monomadic:
I'm getting this error after linking with lerna:
``` bash
 > node_modules/vite/dist/node/chunks/dep-713b45e1.js:30249:10: error: [plugin: vite:dep-scan] Failed to resolve entry for package "@solana/wallet-adapter-svelte". The package may have incorrect main/module/exports specified in its package.json: Failed to resolve entry for package "@solana/wallet-adapter-svelte". The package may have incorrect main/module/exports specified in its package.json.
```

Not sure if this is related, but package.json has these references:
``` json
  "main": "lib/index.js",
  "esnext": "lib/index.js",
  "types": "lib/index.d.ts",
  "files": [
    "lib",
    "src",
    "LICENSE"
  ],
```

And there is no `lib` directory in the project.

MellenIO:
Is there any word on when this will be published to NPM? It'd be really handy for me.

jordansexton:
There's some work left to do, it's not quite functional or API-stable yet. @silvestrevivo did you want to follow up?

silvestrevivo:
@jordansexton Sure! I am already working on it and I think I can open a new PR this week. I've solved some issues and the UI is almost done. Can I open a PR with the UI implementation and the fixes for the adapter?

jordansexton:
Ideally we can do it in 2 PRs just to keep the scope of changes needed for the core package small. If you have a demo with the UI code that shows how it works though, we can review that separately.

PR #118, opened 2021-10-09
Add Vue UI package

lorisleiva:
Hi there 👋 

This PR creates a Vue 3 version of the React UI package using the exact same `style.css` stylesheet.

It exposes the following components:

- `WalletConnectButton`: Connects the selected wallet.
- `WalletDisconnectButton`: Disconnects the current wallet/adapter.
- `WalletIcon`: Displays the icon in the Wallet object.
- `WalletModal`: Selects a wallet from all supported wallets within a Modal.
- `WalletModalButton`: Opens the `WalletModal`.
- `WalletModalProvider`: Exposes a context to child components (similar to React's Context Provider components) using provide and inject from Vue 3's API. This provider delegates to the `useWalletModal.ts` file which is responsible for encapsulating the `provide` and `inject` methods into `initWalletModal` and `useWalletModal` methods.
- `WalletMultiButton`: Aggregates most of the components above to create a simple and effective flow to select, connect and disconnect a wallet.

This package expects the `vue` core package to be installed and its global store `useWallet.ts` to be initialised before using any of these components.

### Basic usage

```vue
<script setup>
import { initWallet, useWallet } from '@solana/wallet-adapter-vue';

// Import the components and use the default styles.
import { WalletModalProvider, WalletMultiButton } from '@solana/wallet-adapter-vue-ui'
import '@solana/wallet-adapter-vue-ui/styles.css'

// Initialise the wallet store.
</script>

<template>
    <wallet-modal-provider>
        <wallet-multi-button></wallet-multi-button>
    </wallet-modal-provider>
</template>
```

### Compiling with Vite

Potentially one of the most important thing to notice here is that this package is not using the standard TypeScript compiler (`tsc`) to bundle the package.

This is because Vue components require their template to be compiled into a Vue-specific render function which the TypeScript compiler does not support. Vue also supports JSX but, even in that case, the JSX template would need to be compiled into a Vue render function. Another way around that would have been to write the template directly as a render function which is fine for small components but quickly becomes difficult to read and maintain for most components.

Therefore, I looked at what other Vue component libraries were doing and saw that [Vite](https://vitejs.dev/) was a popular option (a webpack alternative built by the Vue core team). Futhermore, I learnt that Vite has a [Library Mode](https://vitejs.dev/guide/build.html#library-mode) which fitted the need of the package perfectly.

A small caveat I found with Vite's library mode was that they would compile TypeScript into JavaScript but without emitting the `d.ts` files. Fortunately, there is a maintained [Vite dts plug-in](https://github.com/qmhc/vite-plugin-dts) that takes care of that for us.

### Manual testing

As with my [previous PR](https://github.com/solana-labs/wallet-adapter/pull/107), I've manually tested this package using a Vue 3 application which is [publicly available on GitHub](https://github.com/lorisleiva/solana-vue3-counter) if anyone wants to give it a try. Since the `vue-ui` package is not yet available, I've simply copy/pasted the latest compiled code into the Vue 3 application.

### Screenshots

Capture 2021-10-09T14 14 32@2x

Capture 2021-10-09T14 14 23@2x

---

See #67

### DIFF
--- a/packages/ui/vue-ui/tsconfig.json
+++ b/packages/ui/vue-ui/tsconfig.json
@@ -2,9 +2,7 @@
     "extends": "../../../tsconfig.json",
     "compilerOptions": {
         "noEmit": false,
-        "outDir": "lib",
-        "isolatedModules": false,
-        "typeRoots": []
+        "outDir": "lib"
     },
     "include": ["src"]
 }


jordansexton:
Merged! Thanks for this, I really appreciate you taking the time to talk through how Vue works.

PR #119, opened 2021-10-10
added missing `shx` before rm command for angular project

grmkris:
Building angular package from source on windows fails since `rm` is missing command. Other packages use `shx rm...` so I believe this should be the case here as well.


jordansexton:
Ah, good catch. Thanks!

PR #122, opened 2021-10-14
Use provide/inject for useWallet in vue core

lorisleiva:
Hi there 👋 

As discussed in https://github.com/solana-labs/wallet-adapter/pull/118, here's a PR that makes use of the `provide`/`inject` Vue 3 API to create a `WalletProvider` component analogous to Context components in React.

I've bumped up the version from `0.1.0` to `0.2.0` since this would be a breaking change for anyone currently using `0.1.0`.

### Basic usage

```vue
<script setup>
import { WalletAdapter } from '@solana/wallet-adapter-vue';
</script>

<template>
    <wallet-adapter :wallets="wallets" auto-connect>
        <!-- Now, `useWallet` is available for all descendants. -->
    </wallet-adapter>
</template>
```

### Invalidating adapter listeners

One big advantage of this approach is the user no longer need to manually set up a `onBeforeUnmount` listener that cleans up the adapter listeners. Instead these adapter listeners will be automatically invalidated when the `WalletAdapter` component is destroyed. Below is a little video that shows toggling the `WalletAdapter`'s existence does in fact invalidate adapter listeners (see console logs).

https://user-images.githubusercontent.com/3642397/137267497-d75052fe-0ff7-4507-ad80-e22c4ad9e596.mp4

⚠️ That being said, **this does not fix our issue about adapter listeners not being invalidating on page refresh**. I was curious to know why and I found out that [Vue components are not actually destroyed when the page is unloaded](https://stackoverflow.com/questions/34443187/do-something-before-reload-or-close-in-vue-js). This means I unfortunately had to keep the `beforeunload` listener we had before.

---

See #67

### DIFF
--- a/packages/ui/vue-ui/src/shims-vue.d.ts
+++ b/packages/ui/vue-ui/src/shims-vue.d.ts
@@ -1,9 +1,8 @@
 /// <reference types="vite/client" />
 
 declare module '*.vue' {
-    import { DefineComponent } from 'vue'
+    import { DefineComponent } from 'vue';
     // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/ban-types
-    const component: DefineComponent<{}, {}, any>
-    export default component
-  }
-  
+    const component: DefineComponent<{}, {}, any>;
+    export default component;
+}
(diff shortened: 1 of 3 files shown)


jordansexton:
Interesting, thank you for this. Does this change anything about how the UI components use this package?

PR #147, opened 2021-10-27
Message Signing for Solflare

kirksgithub:
Adding support for solflare message signing


kirksgithub:
Thank you Jackson for your feedback added the remaining code

PR #151, opened 2021-11-02
add message signing support for solflare

Jeffrieh:
Solflare supports signing of messages, made some changes to the wallet adapter to support this.
let me know if any other changes have to be made.

### DIFF
--- a/packages/wallets/solflare/src/adapter.ts
+++ b/packages/wallets/solflare/src/adapter.ts
@@ -1,5 +1,5 @@
 import {
-    BaseSignerWalletAdapter,
+    BaseMessageSignerWalletAdapter,
     EventEmitter,
     pollUntilReady,
     WalletConnectionError,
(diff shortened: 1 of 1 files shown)


Jeffrieh:
Saw there's already a PR active, closing.

Jeffrieh:
noticed there was already a similar PR open after creating this one, but this one should be ready to merge compared to the other one so i'll keep it open.

jordansexton:
Thanks!

jordansexton:
@Jeffrieh can you clarify whether the correct interface for this is `signMessage` (as implemented here) or `sign` (as in #147)? Also, is `Uint8Array.from(signature)` necessary if `signature` is already a `Uint8Array`?

Jeffrieh:
Hi Jordan,

I can confirm signMessage is actually the correct interface, as per the docs on : https://docs.solflare.com/technical/connecting-your-solana-dapp-with-solflare

As for your second question, i believe you're actually right, small oversight on my part, since signature is already a Uint8Array, this would be sufficient : 

```
const { signature } = await wallet.signMessage(message, 'utf8');
return signature
```

PR #156, opened 2021-11-04
Add Aldrin wallet adapter

rindeveloper:


jordansexton:
Sorry, I can't accept this PR. The core adapter code is just copied from https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/sollet/src/adapter.ts

The Sollet adapter already provides the functionality needed, and allows you to provide a different provider URL, so it's not necessary to add a new package to support this.

Check out how this code uses the Sollet adapter and just changes the provider URL and icon:
https://github.com/solana-labs/wallet-adapter/blob/master/packages/core/wallets/src/solflareWeb.ts

If you open a PR that does this, I'd be happy to review it! Please don't include changes to the usage, examples, and starter projects.

PR #161, opened 2021-11-10
feature/fix-and-complete-svelte-core-package

silvestrevivo:
### feature/fix-and-complete-svelte-core-package

The purpose of this PR is to fix and complete the previous [one](https://github.com/solana-labs/wallet-adapter/pull/114). Essentially there were two things to fix: 

- Smooth integration with SSR in SvelteKit
- How to destroy the wallet adapter

##### SSR in SvelteKit

Running __SvelteKit in SSR__, there were some issues loading the site in development and deploying in a node server (_this issue was not happening deploying as __SPA__, but in development it remains because by default SvelteKit needs a server to work on it_). Because of this, we needed to import packages and stores inside of `onMount()` methods to prevent errors, making the app not so readable with to much extra boilerplate. 

__Solution:__ Basically, includding `EventEmmiter3` as package in the implementation, the issue is solved, the server doesn't crash and development and build are working correctly. This means we can import the `storeWallet` as a usual Svelte store without `onMount()`.

From here:
```
let store;
onMount(async () => {
    const module = await import ('@solana/wallet-adapter-svelte');
    store = module.walletStore;
})
```
to here:
```
import { walletStore } from `@solana/wallet-adapter-svelte'
```

##### Destroy wallet adapter

We need to ensure that the adapter listerners should be 'killed' in this three situations:

- Page is refreshed
- Wallet is disconnected
- Wallet is replaced

To ensure this we can put back the `beforeunload` method listening for the `destroyAdapter()` function. With this we reset the listerens if the page is refreshed (_thus we don't depend of a Svelte component and a `onDestroy()` method to do this, keeping the `walletStore` isolated_). At the same time we can use this function inside `disconnect()`, at line __156__ previous the adapter is disconnected. To fix the issue when the wallet is replaced, in `select()` on line __131__ the `await adapter.disconnect()` is replaced for just the `disconnect()` function, doing all the steps at the time.

To show how this works, a [repo](https://github.com/silvestrevivo/solana-svelte-counter) is created as [demo](https://solana-svelte-counter.netlify.app/) which is deployed in Netlify connected to the DevNet.


silvestrevivo:
@jordansexton Hi there! 

This PR is ready for review. Looking forward for reading your feedback!

packages/core/svelte/src/workSpace.ts
```diff
@@ -0,0 +1,14 @@
+import { writable } from 'svelte/store';
+import type { Program, Provider, web3 } from '@project-serum/anchor';
```
cogoo, line 2: Unsure about the best way to handle this scenario. We only need the `@project-serum/anchor` package for typing, but seems unreasonable to list this as `devDependency`

silvestrevivo: I think that in this case, make totally sense for me to list this just as `devDependency`.

packages/core/svelte/package.json
```diff
@@ -24,12 +24,14 @@
         "@solana/wallet-adapter-base": "^0.6.0",
         "@solana/wallet-adapter-wallets": "^0.10.1",
         "@solana/web3.js": "^1.29.2",
+        "eventemitter3": "^4.0.7",
```
jordansexton, line 27: `@solana/wallet-adapter-base` exports the `EventEmitter` type and class, you may not need it as a dep. It's also not clear to me where this being used.

packages/core/svelte/src/workSpace.ts
```diff
@@ -0,0 +1,14 @@
...
+
+type WorkSpace = {
+	baseAccount?: Keypair;
+	connection: Connection;
+	provider?: Provider;
+	program?: Program<Idl>;
+	systemProgram?: typeof web3.SystemProgram;
+};
+
+export const workSpace = writable<WorkSpace>(undefined);
```
jordansexton, line 14: Not sure what this is, but it doesn't belong in the core package, not even as a devDependency. This is for wallets only, not Solana programs or Anchor.

packages/core/svelte/src/walletStore.ts
```diff
@@ -290,13 +295,10 @@ walletAdapterStore.subscribe(({ adapter }: { adapter: Adapter | null }) => {
...
 walletAdapterStore.subscribe(async ({ adapter }: { adapter: Adapter | null }) => {
-    if (!adapter) return;
-
     const { autoConnect } = get(walletConfigStore);
-    if (!autoConnect) return;
-
     const { ready, connected, connecting } = get(walletStore);
-    if (!ready || connected || connecting) return;
+
+    if (!adapter || !autoConnect || !ready || connected || connecting) return;
```
jordansexton, line 301: Unless there's a functional difference, please keep the early returns.

packages/core/svelte/src/walletStore.ts
```diff
@@ -151,6 +153,7 @@ async function disconnect(): Promise<void> {
             ...storeValues,
             disconnecting: true,
         }));
+        await destroyAdapter();
```
jordansexton, line 156: `destroyAdapter` is not an async function.

I think we may want to rename it as well (it doesn't destroy the adapter, or anything really). Maybe just a generic `cleanup` function would do.

packages/core/svelte/src/walletStore.ts
```diff
@@ -256,7 +259,9 @@ function onDisconnect() {
 
 walletNameStore.subscribe(({ walletName }: { walletName: WalletName | null }) => {
     const { localStorageKey } = get(walletConfigStore);
-    setLocalStorage(localStorageKey, walletName);
+    if (walletName){
+        setLocalStorage(localStorageKey, walletName);
+    }
```
jordansexton, line 264: We discussed this before, and I'm still not sure about this change. The idea here is that if `walletName` is `null` (which is explicitly allowed as an arg), then `setLocalStorage` should still be called with `null` in order to remove the item. This is what should be done any time the wallet is imperatively disconnected.

packages/core/svelte/package.json
cogoo: This prevents an issue that only arises on local development in SSR mode.

The error comes from the transpiled `adapter.js`

```javascript
import EventEmitter from 'eventemitter3';
export { EventEmitter };
export class BaseWalletAdapter extends EventEmitter {
}
export var WalletAdapterNetwork;
(function (WalletAdapterNetwork) {
    WalletAdapterNetwork["Mainnet"] = "mainnet-beta";
    WalletAdapterNetwork["Testnet"] = "testnet";
    WalletAdapterNetwork["Devnet"] = "devnet";
})(WalletAdapterNetwork || (WalletAdapterNetwork = {}));
//# sourceMappingURL=adapter.js.map
```

The import to `eventemitter3` doesn't exist in the bundle (probably because it's not actually used).

The error was something like `cannot extend value of undefined`, so we decided to include it as dep so other uses running `SvelteKit` in local development wouldn't run into this issue.

packages/core/svelte/src/workSpace.ts
cogoo: This is now moved to the `svelte-ui` package

packages/core/svelte/src/workSpace.ts
silvestrevivo: @cogoo `@project-serum/anchor` this will be removed since this file is removed too.

packages/core/svelte/src/workSpace.ts
silvestrevivo: I've removed this file and moved to #162 There will be necessary not only for Anchor, but even with it to share the connection in the whole application.

jordansexton: Sorry, no Anchor dependency, including in the UI package. I'm a big fan of Anchor, but it's not a dependency in this project, the goal of which is wallet support. Feel free to publish a library of helper utilities if you like!

jordansexton: Put another way, if I'm building a dapp on Solana with Svelte and I'm not using Anchor, I should be able to use all the packages I need without downloading, importing, or expecting Anchor as a dependency. Utilities that support Anchor without adding a dependency (e.g. the `useAnchorWallet` hook) are okay though.

packages/core/svelte/package.json
silvestrevivo: @jordansexton this solves the way to call the `walletStore`. If you remember in the first PR, I was importing the store inside an `onMount()` method because in SSR the adapter was crashing. Including `eventemitter3` the issue is solved and the app doesn't give any errors in development and in production.

packages/core/svelte/src/workSpace.ts
cogoo: ah ok, that makes sense! This was our take on the `useAnchorWallet`. So since we are just using it as a type. Would you be open to having wider typing? That's on the verge of `any` or `unknown`

jordansexton: I'm not sure I follow. How is defining a workspace with an IDL similar to `useAnchorWallet`?

cogoo: Sorry, there's some missing context. There's an `AnchorConnectionProvider.svelte` that goes with this in the `svelte-ui` package. Fully agree with your earlier point on the anchor dep. So we'll move that to a separate package all together

packages/core/svelte/src/walletStore.ts
silvestrevivo: Yes, we discussed this before and after you comment, I went to check it out again and I see what is happening. 

#### Doing your solution:

The subscribe method is listening since the very first value of `walletName`, and initially this is `null`. That's why autoConnect was not working: as soon I refreshed, the value was `null` and the localStorage removed. 

#### Doing my solution:
It didn't work either because there is not chance to remove the localStorage when we disconnect the wallet and the `walletName` is null. 

#### a new proposal
I've tested this several times and it is working as expected. What I want is that `setLocalStorage(localStorageKey, walletName);` is going to work when `initialize` is called, and the only way to know it is when `wallets.length > 0`. In this case we can check the changing value of `walletName` and save or remove the localStorage. 

```
if (wallets.length > 0) {
    setLocalStorage(localStorageKey, walletName);
}
```

jordansexton: Hmm. This third proposal looks much more hacky than I'm comfortable with. I think we need to go back to the drawing board and take cues from how the React and Vue packages handle more of the implementation.

silvestrevivo: @jordansexton investigating a bit further about this, the `subscribe` method reads the initial Store value and the changes related with it. The same that reactivity `$` in the Svelte components. In this case we need to ensure we are triggering `setLocalStorage` after the wallet is initialized. The wallets array in the only parameter which can ensure that it is happening.

jordansexton: Counting on the wallets array not to be populated doesn't feel like a future proof design (it's merely incidental to what the intent of this code is, which is to control localStorage).

What I meant above is that the React and Vue implementations are more straightforward to follow and maintain because the side effects of the code are controlled. We should try to follow them and bring this approach to the Svelte implementation, which right now is much less easy to follow logically, and there are several behaviors that are occurring that we don't have a complete understanding of yet.

silvestrevivo: There is another solution and is to call directly the function like this: 

on the `select()` function we can do `setLocalStorage(localStorageKey, newName);`
and in the `disconnect()`, `setLocalStorage(localStorageKey, null);`

The effect is the same and we prevent to call the function at the first load of the subscribe method. I think the Vue implementation the the cleanest, but Svelte ( React neither ) has not the Composition API. It is impossible to have the same solution for the 4 frameworks, since Svelte is the only one pure reactive. And the structure of the Svelte implementation is clear. We are used to work with many Stores. Basically to build this implementation, I brought the Vue one line by line and adapt it to the Svelte logic. 

Even we are making something totally isolated and not depending of Svelte Components as in the first commits of the first version. The React one needs the WalletProvider.ts for example. What I mean is that, each framework has it own logic and regarding complexity, the Angular impl is not very straightforward.

cogoo: > Counting on the wallets array not to be populated doesn't feel like a future proof design (it's merely incidental to what the intent of this code is, which is to control localStorage).
> 
> What I meant above is that the React and Vue implementations are more straightforward to follow and maintain because the side effects of the code are controlled. We should try to follow them and bring this approach to the Svelte implementation, which right now is much less easy to follow logically, and there are several behaviors that are occurring that we don't have a complete understanding of yet.

Just to chime in; I haven't followed this implementation or the react / vue implementations closely, but following this thread, this particular line of code is guarding against the standard behaviour of a svelte store. The moment you `.subsribe` on a store, the current values in that store are emitted. 

On an initial load, those vales will be `null` which will trigger the removal of the `localStorage` item (incorrectly). We could add a safeguard around that to disregard the first value emitted, which will solve this problem. Personally I think adding the safeguard is OK for now, as this is just how svelte stores behave naturally.

Some more general points;
1. It will be nice to write idiomatic svelte code, rather than mirror (which I don't think is what you're suggesting) the react / vue implementation
2. It's more important to get the overall API design of the core package correct. As its easier to fix/change implementation than to change the API once it's actually in use. The API seems to be inline with the other implementations

@silvestrevivo are my findings correct?

silvestrevivo: @cogoo Your explanation is very accurate and essentially that is what is happening. Regarding your points:

1. I think the solution is aligned with the Svelte style. It is true that is mirroring the Vue implementation since Vue and Svelte are quite similar. But as we have mentioned before, in the same way that the React iml has a component like `WalletProvider` or `WalletConnection` as part of the implementation, we could do something similar for this one. But in this case, as you requested in the previous PR, he have worked hard to solve it just in one file and in an isolated way. Like the Vue implementation. 
2. I think as well the API is aligned with the others.

packages/core/svelte/src/walletStore.ts
```diff
@@ -84,13 +83,54 @@ export const walletStore = writable<WalletStore>({
...
+    function updateWalletName(walletName: WalletName | null) {
+        const { localStorageKey } = get(walletConfigStore);
+
+        set({ walletName });
+        setLocalStorage(localStorageKey, walletName);
+    }
+
+    return {
+        subscribe,
+        updateName: (walletName: WalletName | null) => updateWalletName(walletName),
```
cogoo, line 100: With the new semantics, don't think you should call `updateName` with `null`

packages/core/svelte/src/walletStore.ts
```diff
@@ -84,13 +83,54 @@ export const walletStore = writable<WalletStore>({
...
+function createWalletAdapterStore() {
+    const { subscribe, set } = writable<WalletAdapterStore>({
+        adapter: null
+    });
+
+    return {
+        subscribe,
+        updateAdapter: (adapter: Adapter) => {
+            // clean up adapter event listeners
+            cleanup();
```
cogoo, line 116: maybe `cleanupAdapterEventListeners`?

packages/core/svelte/src/walletStore.ts
```diff
@@ -103,35 +143,32 @@ export async function initialize({
...
+        onError
+    });
 
-    walletNameStore.update((storeValues: WalletNameStore) => ({
-        ...storeValues,
-        walletName: getLocalStorage<WalletName>(localStorageKey),
-    }));
+    const walletName = getLocalStorage<WalletName>(localStorageKey);
+
+    if (walletName) {
```
cogoo, line 159: Prevents setting walletName to `null` if local storage is empty. Not having this check doesn't have any adverse effects, just prevents unnecessary side effects from running

packages/core/svelte/src/walletStore.ts
```diff
@@ -375,3 +371,8 @@ export function destroyAdapter() {
     adapter.off('disconnect', onDisconnect);
     adapter.off('error', onError);
 }
+
+if (typeof window !== 'undefined') {
```
cogoo, line 375: I don't think this is needed. I'll be surprised if the browser doesn't clean these up automatically

jordansexton: In this case the issue is timing and race condition prevention. When the window unloads, the adapter's `disconnect` event will usually fire. Removing the event listeners prevents this from causing the LocalStorage wallet name from being cleared, or errors due to the spontaneous disconnection from being caught.

Removing them explicitly is done in the React and Vue implementations as well.

packages/core/svelte/src/walletStore.ts
jordansexton: Not sure I understand this change, or the one below with `createWalletAdapterStore`. Why have an internal named setup function that is only called once? And why have `updateName` and `reset`? It seems like you could just do this if you're looking to decorate the `set` function and keep the API of the store.

```ts
export const walletNameStore = (function () {
    const { subscribe, set } = writable<WalletNameStore>({
        walletName: null
    });
    
    return {
        subscribe,
        set (walletName: WalletName | null) {
            const { localStorageKey } = get(walletConfigStore);

            set({ walletName });
            setLocalStorage(localStorageKey, walletName);
        }
    };
})();
```

Is it idiomatic to wrap stores like this? Or is there a way to react to changes to them?

cogoo: Your solution works, but it hides the intent. I wanted to make it clear that you're resetting the store, as when you're just reading the code, its not immediately clear that there's any side effects. It's not until I traced the code, I realised that updating the name to `null` triggers some side effects.

Its fairly common to create a custom store (wrap stores) to create a more clearly defined API, rather than just `set/update` and also a neat way to run side effect without having to listen to changes with a `.subscribe`

packages/core/svelte/src/walletStore.ts
cogoo: Thanks, that makes sense.

packages/core/svelte/src/walletStore.ts
cogoo: There's an example of this in the [svelte tutorial](https://svelte.dev/tutorial/custom-stores)

```ts
function createCount() {
	const { subscribe, set, update } = writable(0);

	return {
		subscribe,
		increment: () => update(n => n + 1),
		decrement: () => update(n => n - 1),
		reset: () => set(0)
	};
}
```

jordansexton: Got it, thanks for clarifying.

packages/core/svelte/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "private": true,
     "name": "@solana/wallet-adapter-svelte",
-    "version": "0.1.0",
+    "version": "0.2.0",
```
jordansexton, line 4:
```suggestion
    "version": "0.1.0",
```

Don't need to bump this version, the package is private and unpublished.

jordansexton:
Merging this now, thanks @silvestrevivo! @cogoo has indicated he's going to do some API refactoring as a followup.

PR #164, opened 2021-11-11
Assign type="button" attribute to button tags

smit1625:
Hi there! I was starting to toy around with the React component for selecting a wallet and I love it! One thing that is causing me some trouble, though, is that if it is rendered inside a `<form />` tag, it actually submits the form on click! The recommended way to deal with this is to add `type="button"` to your `<button />` tag. More info on that here: https://stackoverflow.com/questions/2825856/html-button-to-not-submit-form.

So I've added it for the React and the Vue components. I'm not familiar with ant-design or material-ui myself, but it seems from my uninformed perspective that both of those components allow the passing of arbitrary props, so a user could probably handle this themselves if they were using those components.

Please let me know if there's anything else I have missed or you would like for me to add, and thank you for your time!

### DIFF
--- a/packages/ui/ant-design/src/WalletConnectButton.tsx
+++ b/packages/ui/ant-design/src/WalletConnectButton.tsx
@@ -6,6 +6,7 @@ import { WalletIcon } from './WalletIcon';
 export const WalletConnectButton: FC<ButtonProps> = ({
     type = 'primary',
     size = 'large',
+    htmlType = 'button',
     children,
     disabled,
     onClick,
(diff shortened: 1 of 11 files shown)


jordansexton:
Good catch, thanks! I added the same for Material UI and Ant Design.

jordansexton:
Published:

 - @solana/wallet-adapter-ant-design@0.9.1
 - @solana/wallet-adapter-material-ui@0.13.1
 - @solana/wallet-adapter-react-ui@0.6.1
 - @solana/wallet-adapter-vue-ui@0.1.1

smit1625:
Thank you! This is awesome, your response time is amazing.

PR #166, opened 2021-11-12
Include wallets that provide signMessage

neverything:
- Changed sentence structure
- Added all wallet adapters that already support `signMessage`


jordansexton:
Thanks!

Issue #142, opened 2021-10-26
vue-ui: TypeScript doesn't compile


jordansexton:
@lorisleiva the package is throwing some strange errors when compiling the TypeScript source:

```
➜  vue-ui git:(master) tsc
../../../node_modules/@types/react/index.d.ts:120:51 - error TS2344: Type 'C' does not satisfy the constraint 'ElementType<any>'.
  Type 'string | number | ForwardRefExoticComponent<any> | (new (props: any) => Component<any, {}, any>) | ((props: any, context?: any) => ReactElement<any, string | JSXElementConstructor<...>> | null)' is not assignable to type 'ElementType<any>'.
    Type 'number' is not assignable to type 'ElementType<any>'.
      Type 'C' is not assignable to type 'FunctionComponent<any>'.
        Type 'string | number | ForwardRefExoticComponent<any> | (new (props: any) => Component<any, {}, any>) | ((props: any, context?: any) => ReactElement<any, string | JSXElementConstructor<...>> | null)' is not assignable to type 'FunctionComponent<any>'.
          Type 'string' is not assignable to type 'FunctionComponent<any>'.

120         "ref" extends keyof ComponentPropsWithRef<C>
                                                      ~

../../../node_modules/@types/react/index.d.ts:121:49 - error TS2344: Type 'C' does not satisfy the constraint 'ElementType<any>'.
  Type 'C' is not assignable to type 'FunctionComponent<any>'.

121             ? NonNullable<ComponentPropsWithRef<C>["ref"]> extends Ref<
                                                    ~

../../../node_modules/@vue/runtime-core/dist/runtime-core.d.ts:1190:6 - error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.

1190     [BooleanFlags.shouldCast]?: boolean;
          ~~~~~~~~~~~~

../../../node_modules/@vue/runtime-core/dist/runtime-core.d.ts:1191:6 - error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.

1191     [BooleanFlags.shouldCastTrue]?: boolean;
          ~~~~~~~~~~~~

../../../node_modules/@vue/runtime-core/dist/runtime-core.d.ts:1725:10 - error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.

1725 export { TrackOpTypes }
              ~~~~~~~~~~~~

[...]

PR #167, opened 2021-11-13
Fix vue-ui type incompatibilities

lorisleiva:
Hi there 👋 

I was browsing through the TS documentation today and came across the `types` option which basically whitelists which types within `@types` should be included in the compilation.

So I tried to see if it could fix our `vue-ui` compiling issues (#142) and it did!

Before these changes I was getting the same errors are in #142 when running `yarn docs` and after the changes, I was able to run `yarn docs` successfully. 🎉


jordansexton:
Nice find, thanks!

PR #176, opened 2021-11-22
Add Aldrin wallet adapter

rindeveloper:


jordansexton:
Just got back from vacation, thanks for the contribution! I mentioned this here: https://github.com/solana-labs/wallet-adapter/pull/156#issuecomment-966493330

> Please don't include changes to the usage, examples, and starter projects.

You only need to change the core wallets package to add Aldrin, so please revert the other changes.

Is Aldrin's icon an SVG or PNG? The encoded base64 looks quite large for an SVG.

jordansexton:
Closing for now since there's been no response.

PR #180, opened 2021-11-24
Feat: Torus solana wallet integration 

rathishubham7:
Following things were added-
- Torus solana wallet integration (https://docs.tor.us/solana-wallet/api-reference/installation)
- Added `detected` parameters in the wallets
- Updated /packages/ui/react-ui components 
- Updated /pacakages/starter/react-ui-starter 
- Updated wallet flow with easier signup process.


vpontis:
As a user of this package, I would recommend against tying support with Torus until they fix things like [this issue](https://github.com/torusresearch/eccrypto/pull/14) which makes it hard to this package into your app.

rathishubham7:
@vpontis The PR has been merged and released in 1.1.8.

jordansexton:
This PR seems a bit more broad than it needs to be. It contains

- changes to the Torus adapter
- a new `detected` property added to the `Wallet` interface
- changes to UI components

Could we limit the scope of this PR to just the Torus changes? This would help us review and merge quickly.

I can see the value of the `detected` change, and I think we should evaluate it in a separate PR. The adapters have a `ready` property which

1. serves effectively the same purpose
2. is expected to be updated asynchronously and fires events

This is important because wallet extensions/injections aren't always loaded immediately. Relying on this synchronously in the UI isn't very safe, so the UI changes that require the `detected` property should be a followup once we get the design for detecting wallets right. This may mean removing the `ready` API we have now and decoupling the `ready` event from the adapter.

@rathishubham7 what do you think?

rathishubham7:
Yes. The ready property could be used for wallet detection. I'll limit the scope of this PR to torus related changes only. 
Closing this.

PR #181, opened 2021-11-24
Add phantom's signAndSendTransaction method

jordansexton:


jordansexton:
The base `signTransaction` method:
https://github.com/solana-labs/wallet-adapter/blob/75388bd6774e0b5ed68cf0a4737d656adf14110f/packages/core/base/src/signer.ts#L13-L40

This sets the fee payer, recent blockhash, and partially signs if signers are provided.

Does this need to be added to the overridden implementation for Phantom also?

fragosti:
@jordansexton Phantom will already add this stuff internally https://github.com/phantom-labs/sandbox/blob/main/src/App.tsx#L155

fragosti (approved):

jordansexton:
Successfully published:
 - @solana/wallet-adapter-phantom@0.8.1

PR #187, opened 2021-12-02
WalletConnect V2

AlexBHarley:
Trying to get this working with [Steakwallet](https://steakwallet.fi). The repository for Steakwallet is private at the moment, but I can release a version of it with the Solana <> WalletConnect enabled if anyone is interested in trying it. 

Adapter mostly works but would be nice to:

- [ ] checks persisted state on reload
- [ ] handle multiple sessions 
- [ ] confirm with someone that we want our over the wire signature encoding to be base64, probably not


jordansexton:
Wow, thanks for taking this on! I'd love to see how you're integrating this to be able to test it out.

AlexBHarley:
Take a look at https://twitter.com/steakwallet/status/1466803157233803270, it's not live right now on app stores though. 

We already support Cosmos and Ethereum via WalletConnect V2, so Solana wasn't much of a stretch. Happy to just follow whatever you or the WC team recommend for the over the wire format.

packages/starter/react-ui-starter/src/Navigation.tsx
```diff
@@ -1,16 +1,42 @@
...
+                fromPubkey: publicKey,
+                toPubkey: Keypair.generate().publicKey,
+                lamports: 1,
+            })
+        );
+
+        const signature = await sendTransaction(transaction, connection);
+        await connection.confirmTransaction(signature, 'confirmed');
+        alert(`done ${signature}`);
+    }, [publicKey, sendTransaction, connection]);
```
jordansexton, line 26: Just a quick note -- the `example` project (used by the demo) is helpful for testing out wallets like this since it already comes with these things.

jordansexton:
This looks really clean.

I like that the JSON wire format of the transaction should easily allow a wallet to show it. It should probably include any signatures that were added (e.g. with `partialSign`).

It could also [serialize the transaction](https://solana-labs.github.io/solana-web3.js/classes/Transaction.html#serialize) (provided `{ requireAllSignatures: false, verifySignatures: false }` as options) and then [deserialize it](https://solana-labs.github.io/solana-web3.js/classes/Transaction.html#from).

Upside is a lot of bytes saved over the wire (accounts get compressed) and it follows the common format, but downside is the client needs to know how to deserialize it (not really an issue if the client has web3.js, but maybe a problem for a mobile app).

What do you think?

packages/wallets/walletconnect/src/adapter.ts
```diff
@@ -164,19 +183,33 @@ export class WalletConnectWalletAdapter extends BaseSignerWalletAdapter {
...
+                    topic: this._session.topic,
                     request: {
                         method: WalletConnectRPCMethod.SOL_SIGN_TRANSACTION,
-                        params: transaction,
+                        // TODO: If not manually encoded you get weird { _bn: '' } type objects
+                        params: {
+                            feePayer: transaction.feePayer!.toBase58(),
+                            instructions: transaction.instructions.map((i) => ({
+                                programId: i.programId.toBase58(),
+                                data: i.data.toString(),
```
jordansexton, line 198: We might want to explicitly base58 this -- `data` is just a Buffer, right?

jordansexton:
Actually, the more I think about it, the wallet app is going to need to know how to serialize the tx anyway to sign it. So we should probably use the ordinary wire format for it?

packages/wallets/walletconnect/src/adapter.ts
```diff
@@ -164,19 +183,33 @@ export class WalletConnectWalletAdapter extends BaseSignerWalletAdapter {
...
+                                    pubkey: k.pubkey.toBase58(),
+                                })),
+                            })),
+                            recentBlockhash: transaction.recentBlockhash,
+                        },
                     },
                 });
 
-                transaction.addSignature(publicKey, signature);
+                transaction.addSignature(publicKey, Buffer.from(signature, 'base64'));
```
jordansexton, line 210: We probably want to use base58 over the wire instead of base64 for the sig too.

AlexBHarley:
Thanks for the feedback!

So, I'd defer to you for what we're sending over the wire, you know the Solana ecosystem better than myself. This was just a quick proof of concept, but whatever we ship here will likely become part of the WalletConnect [documentation](https://docs.walletconnect.com/json-rpc/cosmos) so I'd hate to rush the format and not have it be a long term standard.

I agree that wallets should always be able to decode a serialized transaction and present some info to users when being signed. There's a few examples so far: 

- [eth_sendTransaction](https://docs.walletconnect.com/json-rpc/ethereum#eth_sendtransaction) is a well formed JSON object with the properties `gas`, `to`, `from`, but obviously `data` can be an encoded contract call
- [cosmos_signDirect](https://docs.walletconnect.com/json-rpc/cosmos#cosmos_signdirect) is a well formed JSON object with `signerAddress`, `chainId` but the actual interesting `bodyBytes` field is encoded
- Stellar [stellar_signAndSubmitXDR ](https://docs.walletconnect.com/json-rpc/stellar#stellar_signandsubmitxdr) is totally opaque with just an `xdr` field.

@pedrouid do you have an opinion on this?

jordansexton:
Linking this important docs PR for reference: https://github.com/WalletConnect/walletconnect-docs/pull/108

Syaba999:
> Linking this important docs PR for reference: [WalletConnect/walletconnect-docs#108](https://github.com/WalletConnect/walletconnect-docs/pull/108)

Hi. PR has been merged

jordansexton:
Closing for now since this isn't implemented in WalletConnect yet. We can reopen when it's ready there!

AlexBHarley:
Closing in favour of https://github.com/solana-labs/wallet-adapter/pull/366

PR #188, opened 2021-12-04
Torus wallet adapter 

rathishubham7:
Scope: 
1. Torus adapter changes 
2. Added a "Get started" button


packages/core/wallets/package.json
```diff
@@ -40,7 +40,7 @@
         "@solana/wallet-adapter-sollet": "^0.8.0",
         "@solana/wallet-adapter-solong": "^0.7.0",
         "@solana/wallet-adapter-tokenpocket": "^0.2.0",
-        "@solana/wallet-adapter-torus": "^0.8.0"
+        "@solana/wallet-adapter-torus": "^0.9.1"
```
jordansexton, line 43: I think this can just be 0.9.0?

packages/starter/react-ui-starter/src/Wallet.tsx
```diff
@@ -24,14 +24,12 @@ export const Wallet: FC = () => {
...
             getPhantomWallet(),
-            getSlopeWallet(),
             getSolflareWallet(),
-            getTorusWallet({
-                options: { clientId: 'Get a client ID @ https://developer.tor.us' },
-            }),
-            getLedgerWallet(),
             getSolletWallet({ network }),
+            getSlopeWallet(),
+            getLedgerWallet(),
```
jordansexton, line 32: Can we keep the order of these the same for now?

packages/ui/react-ui/package.json
```diff
@@ -36,5 +36,6 @@
         "@types/react": "^17.0.24",
         "react": "^17.0.2",
         "react-dom": "^17.0.2"
-    }
+    },
+    "dependencies": {}
```
jordansexton, line 40: Is this empty deps list necessary?

packages/ui/react-ui/src/Button.tsx
```diff
@@ -3,28 +3,26 @@ import React, { CSSProperties, FC, MouseEvent, ReactElement } from 'react';
 export interface ButtonProps {
     className?: string;
     disabled?: boolean;
-    endIcon?: ReactElement;
     onClick?: (e: MouseEvent<HTMLButtonElement>) => void;
     startIcon?: ReactElement;
+    endText?: string;
```
jordansexton, line 8: Hmm this is a breaking API change for this component because it removes the `endIcon` prop. We should really just evaluate the Torus changes on their own.

packages/ui/react-ui/src/WalletModalProvider.tsx
```diff
@@ -6,7 +6,7 @@ export interface WalletModalProviderProps extends WalletModalProps {
     children: ReactNode;
 }
 
-export const WalletModalProvider: FC<WalletModalProviderProps> = ({ children, ...props }) => {
+export const WalletModalProvider: FC<WalletModalProviderProps> = ({children, ...props}) => {
```
jordansexton, line 9: Please run `yarn fmt`, it'll fix the formatting with Prettier.

packages/wallets/torus/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@solana/wallet-adapter-torus",
-    "version": "0.8.1",
+    "version": "0.9.1",
```
jordansexton, line 3: 0.9.0 here also.

packages/wallets/torus/src/adapter.ts
```diff
@@ -1,109 +1,98 @@
...
-import { Keypair, PublicKey, Transaction } from '@solana/web3.js';
-import OpenLogin, { OPENLOGIN_NETWORK, OpenLoginOptions } from '@toruslabs/openlogin';
-import { getED25519Key } from '@toruslabs/openlogin-ed25519';
+import { PublicKey, Transaction, Message, Cluster } from '@solana/web3.js';
+import Torus, { TorusParams } from '@toruslabs/solana-embed';
 
 export interface TorusWalletAdapterConfig {
-    options: Partial<OpenLoginOptions> & Omit<OpenLoginOptions, 'network'>;
+    torusParams?: TorusParams;
+    // network?: WalletAdapterNetwork;
```
jordansexton, line 23: If `network` isn't used, it can be removed.

packages/wallets/torus/src/adapter.ts
```diff
@@ -1,109 +1,98 @@
...
         super();
-        this._options = { uxMode: 'popup', network: OPENLOGIN_NETWORK.MAINNET, ...config.options };
         this._connecting = false;
-        this._openLogin = null;
-        this._keypair = null;
+        this._torus = null;
+        this._publicKey = null;
+        this._config = config.torusParams || {};
+        // if (!this.ready) pollUntilReady(this, config.pollInterval || 1000, config.pollCount || 3);
+        // this.ready = true;
```
jordansexton, line 46: Likewise, if polling isn't needed, it can be removed, along with the config for it.

packages/wallets/torus/src/adapter.ts
```diff
@@ -1,109 +1,98 @@
...
     }
 
     get publicKey(): PublicKey | null {
-        return this._keypair?.publicKey || null;
+        return this._publicKey;
     }
 
     get ready(): boolean {
-        return typeof window !== 'undefined';
+        return true
```
jordansexton, line 54: This change should be reverted -- the code below relies on `window`, and I imagine Torus Embed does too?

packages/wallets/torus/src/adapter.ts
```diff
@@ -1,109 +1,98 @@
...
         try {
             if (this.connected || this.connecting) return;
             this._connecting = true;
 
-            let openLogin: OpenLogin;
-            let privateKey: string;
+            // Check if torus is initialized, torus.init({config})
+            let torus = typeof window !== 'undefined' && window.torus;
+            if (!torus) torus = new Torus();
+            window.torus = torus;
```
jordansexton, line 73: This probably shouldn't happen in practice, but this code will error if called when `window` is undefined because it tries to set `window.torus` in either case. It should throw `WalletNotReadyError` instead of proceeding.

packages/wallets/torus/src/adapter.ts
```diff
@@ -1,109 +1,98 @@
...
 } from '@solana/wallet-adapter-base';
-import { Keypair, PublicKey, Transaction } from '@solana/web3.js';
-import OpenLogin, { OPENLOGIN_NETWORK, OpenLoginOptions } from '@toruslabs/openlogin';
-import { getED25519Key } from '@toruslabs/openlogin-ed25519';
+import { PublicKey, Transaction, Message, Cluster } from '@solana/web3.js';
+import Torus, { TorusParams } from '@toruslabs/solana-embed';
 
 export interface TorusWalletAdapterConfig {
-    options: Partial<OpenLoginOptions> & Omit<OpenLoginOptions, 'network'>;
+    torusParams?: TorusParams;
```
jordansexton, line 22: Suggestion: `params`

packages/wallets/torus/src/adapter.ts
```diff
@@ -1,109 +1,98 @@
...
+declare const window: TorusWindow;
 
 export class TorusWalletAdapter extends BaseSignerWalletAdapter {
-    private _options: OpenLoginOptions;
     private _connecting: boolean;
-    private _openLogin: OpenLogin | null;
-    private _keypair: Keypair | null;
+    private _torus: Torus | null;
+    private _publicKey: PublicKey | null;
+    private _config: TorusParams;
```
jordansexton, line 37: Suggestion: `_params`

packages/wallets/torus/src/adapter.ts
```diff
@@ -1,109 +1,98 @@
...
-                            );
-                        });
-                    } finally {
-                        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-                        window.removeEventListener('unhandledrejection', listener!);
-                    }
-                }
+                loginResult = await torus.login();
             } catch (error: any) {
                 if (error instanceof WalletError) throw error;
```
jordansexton, line 81: This check is no longer necessary -- `torus.login()` shouldn't throw `WalletError`.

packages/wallets/torus/src/adapter.ts
```diff
@@ -1,109 +1,98 @@
...
                 throw new WalletConnectionError(error?.message, error);
             }
+            if (!loginResult) throw new WalletConnectionError();
```
jordansexton, line 84: Can `loginResult` be falsy without already throwing an error?

packages/wallets/torus/src/adapter.ts
```diff
@@ -114,34 +103,31 @@ export class TorusWalletAdapter extends BaseSignerWalletAdapter {
...
     async disconnect(): Promise<void> {
-        const openLogin = this._openLogin;
-        if (openLogin) {
-            this._openLogin = null;
-            this._keypair = null;
-
+        // torus.logout
+        const wallet = this._torus;
+        if (wallet) {
+            window.torus = null
```
jordansexton, line 109: I'm a little bit sketched out by setting and unsetting window globals like this -- is this something that the Torus Embed code expects?

packages/wallets/torus/src/adapter.ts
```diff
@@ -114,34 +103,31 @@ export class TorusWalletAdapter extends BaseSignerWalletAdapter {
     }
 
     async disconnect(): Promise<void> {
-        const openLogin = this._openLogin;
-        if (openLogin) {
-            this._openLogin = null;
-            this._keypair = null;
-
+        // torus.logout
```
jordansexton, line 106: Should this comment be removed?

packages/wallets/torus/src/adapter.ts
```diff
@@ -114,34 +103,31 @@ export class TorusWalletAdapter extends BaseSignerWalletAdapter {
...
+        const wallet = this._torus;
+        if (wallet) {
```
jordansexton, line 108: Suggestion: `torus` since that's used elsewhere.

packages/wallets/torus/src/adapter.ts
```diff
@@ -150,21 +136,38 @@ export class TorusWalletAdapter extends BaseSignerWalletAdapter {
...
 
-            return transactions;
+    async signMessage(message: Uint8Array): Promise<Uint8Array> {
+        try {
+            const wallet = this._torus;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const signature = await wallet.signMessage(message);
+                return Uint8Array.from(signature);
```
jordansexton, line 160: What's the return type of `Torus::signMessage`? Is it necessary to cast with `Uint8Array.from`?

jordansexton (changes requested):
Thanks for this! I have a few questions and small change requests. I would also like to completely isolate the PR to make only changes to the Torus package. While the UX improvements for the components will be great, they aren't specific to Torus and should be in their own PR.

jordansexton:
I can't write to this branch, moving to https://github.com/solana-labs/wallet-adapter/pull/194

PR #191, opened 2021-12-09
OneKey wallet adapter

zuozhuo:


jordansexton:
Sorry, I can't accept this PR. It looks like it copies the entirety of the Phantom adapter, and only wraps the Phantom and Sollet adapters, adding no functionality. It also incorrectly implements the `BaseMessageSignerWalletAdapter` interface by throwing errors from the core methods, or no-oping them.

PR #192, opened 2021-12-10
[Vue] Add support for global stores

lorisleiva:
Hi there 👋 

## Context

A while ago, we decided to use the `inject`/`provide` API from VueJS in the `vue` core package (PR #122). Our main motivations were:
- **It provides a similar API to the other core packages**. The `inject`/`provide` API can be used to create a similar pattern to Contexts in React by offering a `WalletProvider` component that would provide the wallet store to its descendants.
- **It enables us to hook some logic upon unmounting the component**. Because we always inject/provide inside a VueJS component, we can invalidate the adapter when the component is destroyed. However, doing this did not fix another issue we had when users were reloading the page because VueJS does not unmount components on page refresh. We had to explicitly add a `beforeunload` event listener for that purpose.

## Problem

Whilst these are valid motivations, they also have some downsides in the VueJS community. I've had some feedback following a tutorial series I'm writing from VueJS users that couldn't use this pattern properly in their setup. Here are a few downsides that I have experienced and/or heard about.
- **This pattern is confusing for some VueJS users**. In VueJS, the `inject`/`provide` API is mostly tailored for library developers that want to pass data between components that they provide. As such, it is not very popular in the community and some developers are having a hard time getting used to it.
- **This pattern does not integrate well with store management solutions in VueJS**. Whether it's using Vuex, custom stores or something else, VueJS developers tend to use stores a lot to manage their data. As such, they are having to go through extra steps to configure a "workspace" store (containing their Anchor program, etc) when they could just call `initWallet` inside the initialisation method of that store.
- **This pattern cannot provide the wallet store outside of a component**. If we wanted to create an asynchronous function used when someone clicks on a button, we couldn't access the wallet store inside that function. Instead, we'd need to pass it through from the `setup` method making refactoring more difficult.
- **This pattern does not work well with Client-Server frameworks such as NuxtJS** that needs to clearly identify which components should be part of the frontend world and which components could be statically generated.

## Solution

The solution I offer in this PR is to support **both** the `inject`/`provide` pattern and the global store pattern whilst exposing the method that creates a new wallet store so developers can create their own custom solutions and still benefit from the code we offer.

The best way to understand this solution is to have a look at the new `useWallet` implementation.

```ts
const walletStoreKey: InjectionKey<WalletStore> = Symbol();
let globalWalletStore: WalletStore | null = null;

export const useWallet = (): WalletStore => {
    const providedWalletStore = inject(walletStoreKey, undefined);
    if (providedWalletStore) return providedWalletStore;
    if (globalWalletStore) return globalWalletStore;
    throw new Error('Wallet not initialized. Please use the `WalletProvider` component to initialize the wallet.');
};
```

As you can see:
1. We start by checking if there is a wallet store that can be injected. If there is, that means we're inside a `WalletProvider` component and therefore we should use the store it provides.
2. If there's nothing to inject from a `WalletProvider`, then we try to access the global wallet store.
3. If both of the solutions above fail to return a wallet store, then we throw an error because we haven't initialized it. The error still recommends using a `WalletProvider` to stay consistent with the other packages and to ensure this PR has no breaking changes.

## Changes

Here are the changes that were made in this PR.
- As we've seen the `useWallet` method has been modified to support both patterns.
- A new `createWalletStore` method has been provided that contains all of the logic necessary to create a brand new `WalletStore` instance from a given set of `WalletStoreProps`. We're also exposing this method to allow developers to use that code to create more custom solutions.
- A new `provideWallet` method has been added. It delegates immediately to the `createWalletStore` method and uses the `provide` method to provide it to its descendants. This method is now used on the `WalletProvider` component instead of the previous `initWallet` method.
- The `initWallet` method is now used to initialise the `globalWalletStore` object.

## Usage

Here's a quick overview of the different ways we can use the `vue` core package.

### Inject/Provide Pattern

Nothing has changed here.

```vue
<script setup>
import { WalletProvider } from '@solana/wallet-adapter-vue';
</script>

<template>
    <wallet-provider :wallets="wallets" auto-connect>
        <!-- Now, `useWallet` is available for all descendants. -->
    </wallet-provider>
</template>
```

### Global Store Pattern

```vue
<script setup>
import { initWallet } from '@solana/wallet-adapter-vue';
</script>

<template>
    <!-- `useWallet` is available everywhere. Not just in the descendants of this component. -->
</template>
```

### Custom Solution

```js
import { createWalletStore } from '@solana/wallet-adapter-vue';
const walletStore = createWalletStore({ wallets, autoConnect: true });
// Persist that `walletStore` any way you want.
```

## Tests and final notes

There's one last little thing I'd like to mention. One downside of using global stores is that we miss out on this advantage from the inject/provide pattern: "It enables us to hook some logic upon unmounting the component". However, I think that this advantage is overrated because most usages of the `WalletProvider` component will be straight after the `App` root component to ensure the entire application can benefit from the wallet store. That means, essentially that the `WalletProvider` component would never be unmounted. Both patterns can, however, benefit from the `beforeunload` event listener to invalidate the adapter on page refresh.

Regarding tests, I've done some manual testing using a dummy counter application for both patterns. Here are the links to the relevant branches of that repository.
- [Test repository for the inject/provide pattern](https://github.com/lorisleiva/solana-vue3-counter/tree/feature/vue-support-global-store) (ensuring everything still works the same).
- [Test repository for the global store pattern](https://github.com/lorisleiva/solana-vue3-counter/tree/feature/vue-support-global-store-2).


timhanlon:
This is going to make a huge difference to usability. Thanks Loris!

jordansexton:
Thanks @lorisleiva! Will review as soon as I can

jordansexton:
Looks great! Thank you so much @lorisleiva

jordansexton:
Because it offers both APIs this is a nonbreaking change, right? Just checking so I can release it.

lorisleiva:
Thanks! 🍺 

Yes, there should be no breaking changes for anyone using the `WalletProvider` component. 👌

jordansexton:
Published: @solana/wallet-adapter-vue@0.4.1

PR #193, opened 2021-12-11
[fix] Prevent default onError from breaking wallet provider memoization

steveluscher:
# Problem

The `WalletProvider` presently supplies a default error handler as part of the component's method signature.

```
function WalletProvider({
  onError: _onError = (error: WalletError) => console.error(error),
  /* ... */
}) { /* ... */ }
```

The way that default properties work in ES6 is such that every time you call the function, it will construct a brand new default function, distinct from the one constructed in the previous function call ([demo](https://jsbin.com/manasidali/edit?js,console)).

Because we use `_onError` as a memoization parameter to `useMemo`, we're busting wallet memoization _way_ more than is necessary – _if_ the developer didn't supply their own, stable, `onError` function.

I've seen this manifest in dApps as a tendency of the app to fetch and refetch the same API calls over and over. One of the root causes, is that we're passing apps a brand new wallet object through context, which causes their React effects to run again, and some of those effects might think ‘oh hey, the wallet changed; I should refetch some data.’

# This PR

In this PR, we inline the default error handler implementation so that:

1. It doesn't get recreated over and over every time the provider re-renders, and
2. Only a genuine change in a user-supplied `onError` function will bust memoization.


packages/core/react/src/WalletProvider.tsx
```diff
@@ -131,19 +131,23 @@ export const WalletProvider: FC<WalletProviderProps> = ({
...
     // Handle the adapter's error event, and local errors
-    const onError = useCallback(
+    const handleError = useCallback(
         (error: WalletError) => {
             // Call the provided error handler unless the window is unloading
-            if (!isUnloading.current) _onError(error);
+            if (!isUnloading.current) {
+                onError
+                    ? onError(error)
+                    : console.error(error);
```
steveluscher, line 140: This is where the default error handler implementation moved to.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -131,19 +131,23 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     }, [isUnloading, setName]);
 
     // Handle the adapter's error event, and local errors
-    const onError = useCallback(
+    const handleError = useCallback(
```
steveluscher, line 134: Rather than to differentiate the user-supplied handler from our synthesized one by an underscore, I left the user-supplied one as `onError` and renamed our synthesized one to the active-voice `handleError`.

jordansexton: Changed the others for consistency.

jordansexton:
Makes sense, thanks for this!

jordansexton:
Published: @solana/wallet-adapter-react => 0.14.1

PR #196, opened 2021-12-12
New starter wallet design 

metallicalfa:
Following changes were added in this PR:
1. modified components in @solana/wallet-adapter-react-ui
2. modified components in @solana/wallet-adapter-react-ui-starter


packages/ui/react-ui/styles.css
```diff
@@ -1,260 +1,383 @@
...
     list-style: none;
-    background: #2c2d30;
-    border-radius: 10px;
-    box-shadow: 0px 8px 20px rgba(0, 0, 0, 0.6);
-    opacity: 0;
-    visibility: hidden;
-    transition: opacity 200ms ease, transform 200ms ease, visibility 200ms;
+    width: 100%;
+    max-height: 17.5rem;
+    overflow-y: auto;
```
jordansexton, line 130: The sizing of the scrolling list is such that a user may be very likely to miss the options below the fold:

Screen Shot 2021-12-15 at 5 59 59 PM

It's not obvious that you can scroll this list and there are more wallets.

jordansexton (commented):
I like how it looks! I left a few comments where I don't think that we should be quite so opinionated in this package.

jordansexton:
Josh has sent me Figma files, so I can add to this PR soon. I Implemented #204, which provides lazy loading of all the wallet dependencies. This was done in preparation for #201, which will enable us to filter the list of wallets shown to the user intelligently.

packages/ui/react-ui/src/WalletSVG.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <stop offset="0.99" stopColor="#00D18C"/>
+        <stop offset="1" stopColor="#00D18C"/>
+        </linearGradient>
+        <clipPath id="clip0_880_5115">
+        <rect width="48" height="48" fill="white" transform="translate(24.5 24)"/>
+        </clipPath>
+        </defs>
+        </svg>
+    )
+}
```
jordansexton, line 51: I think we can remove this SVG, see https://github.com/solana-labs/wallet-adapter/pull/214#issuecomment-999081376 for the latest designs from Josh's Figma file. If we can avoid using colors and graphics, this will fit in better with more dapp designs and be easier to customize.

metallicalfa:
@jordansexton All the comments have been settled except for one.

packages/ui/react-ui/src/WalletSVG.tsx
metallicalfa: Yes, this SVG is used in the "no wallets detected" screen, which is consistent with the design. We have tried to keep the colors relatively neutral. 

image

jordansexton:
I updated this to match Josh's design and behavior more closely. We can make incremental improvements in a followup PR.

jordansexton:
Deployed @ https://solana-labs.github.io/wallet-adapter/example/ and published:

 - @solana/wallet-adapter-example@0.17.0
 - @solana/wallet-adapter-nextjs-starter@0.7.0
 - @solana/wallet-adapter-react-ui-starter@0.8.0
 - @solana/wallet-adapter-react-ui@0.9.0
 - @solana/wallet-adapter-torus@0.11.1

PR #199, opened 2021-12-14
Protect the React wallet adapter against all manner of localStorage tomfoolery

steveluscher:
# Problem

There's nothing worse than a non-critical `WebStorage` API error taking down your entire site. There are many reasons that `localStorage` can fail:

* `localStorage` is out of space
* `localStorage` doesn't exist
* Your content settings in Chrome prohibit writes
* You have ‘block all cookies’ set in Safari
* You're using a private window in Firefox
* You're using an incognito window in Safari
* You're server-side rendering
* you've loaded a `file://` in IE9
* …and on and on.

This is, for instance, why Mango Markets (https://trade.mango.markets) hangs on startup when you try to run it in a private window in Firefox.

We use local storage in the React wallet provider to store the name of your last-used wallet across sessions (eg. so that you can autoconnect next time you visit). Let's protect against fatals in `localStorage` from taking you out.

# This PR

The first two commits in this PR add Jest unit tests that cover:

* the happy path behaviour of `useLocalStorage`
* what should happen when `localStoage` does not exist
* what should happen when `localStoage` is not readable
* what should happen when `localStoage` is not writable
* what should happen when merely accessing `localStorage` yields a fatal

Then, in the next three commits I fix each bug one by one.


steveluscher:
Test run looks like this:

```
 % yarn test
PASS src/__tests__/useLocalStorage-test.tsx
  useLocalStorage
    getting the persisted value
      when local storage has a value for the storage key
        ✓ returns that value (6 ms)
      when local storage has no value for the storage key
        ✓ returns the default value (1 ms)
      when merely accessing local storage results in a fatal error
        ✓ renders with the default value (29 ms)
      when local storage fatals on read
        ✓ renders with the default value (3 ms)
      when local storage does not exist
        ✓ renders with the default value (1 ms)
      when local storage contains invalid JSON
        ✓ renders with the default value
    setting the persisted value
      when setting to a non-null value
        ✓ sets that value in local storage (3 ms)
        ✓ re-renders the component with the new value (1 ms)
      when setting to `null`
        ✓ removes the key from local storage (1 ms)
        ✓ re-renders the component with `null` (1 ms)
      when merely accessing local storage results in a fatal error
        ✓ re-renders the component with the new value (4 ms)
      when local storage fatals on write
        ✓ re-renders the component with the new value (2 ms)
      when local storage does not exist
        when setting to a non-null value
          ✓ re-renders the component with the new value (1 ms)
        when setting to `null`
          ✓ re-renders the component with `null` (1 ms)

Test Suites: 1 passed, 1 total
Tests:       14 passed, 14 total
Snapshots:   0 total
Time:        2.135 s
Ran all test suites.
```

jordansexton:
Wow, thanks for this. Great work on this PR.

PR #206, opened 2021-12-17
Detect ready wallets

### LINKED ISSUE #201: Detect which wallets are installed / able to connect
https://github.com/solana-labs/wallet-adapter/pull/195 has added persistent adapter objects, which makes it much more practical to detect all the available wallets when the page loads.

The React / Vue / Angular / Svelte packages should provide a simple way to do this so the user isn't shown a bunch of invalid options to select from.

jordansexton:
Partially implements #201 (it detects, but doesn't filter)

Filtering of the wallets shouldn't be done in the React context necessarily -- it may be better to handle this in UI components where it can be done more flexibly rather than just removing them from the list.

### DIFF
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -71,25 +71,16 @@ export const WalletProvider: FC<WalletProviderProps> = ({
             const ready = await Promise.all(wallets.map((wallet) => wallet.adapter.ready()));
             // If the wallets haven't changed while waiting, update the details
             if (wallets === waiting) {
-                setDetails(
-                    wallets.reduce<Record<WalletName, WalletDetails>>((details, wallet, index) => {
-                        details[wallet.name] = { ready: ready[index] };
-                        return details;
-                    }, {})
-                );
+                const details = wallets.reduce<Record<WalletName, WalletDetails>>((details, wallet, index) => {
+                    details[wallet.name] = { ready: ready[index] };
+                    return details;
+                }, {});
+                setDetails(details);
             }
         })();
     }, [wallets]);
 
-    const [name, setName] = useLocalStorage<WalletName | null>(localStorageKey, null);
-    const [{ wallet, adapter, ready, publicKey, connected }, setState] = useState(initialState);
-    const [connecting, setConnecting] = useState(false);
-    const [disconnecting, setDisconnecting] = useState(false);
-    const isConnecting = useRef(false);
-    const isDisconnecting = useRef(false);
-    const isUnloading = useRef(false);
-
-    // Map of wallet names to wallets
+    // When the wallets change, map the wallet names to the wallets
     const walletsByName = useMemo(
         () =>
             wallets.reduce<Record<WalletName, Wallet>>((walletsByName, wallet) => {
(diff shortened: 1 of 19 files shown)


packages/core/react/src/WalletProvider.tsx
```diff
@@ -273,7 +286,7 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     return (
         <WalletContext.Provider
             value={{
-                wallets,
+                wallets: readyWallets,
```
danmt, line 289: I'm wondering if we could extend the `Wallet` interface by adding a `ready` boolean attribute. This way the consumers of the WalletContext will have a choice to whether show only ready wallets or something else. Let's say I have a Dapp that supports just a few wallets and I want to show the install status (installed/not installed), I agree that this wouldn't be the main use case, but I think it's nice to have that level of flexibility.

If there was such an attribute for `Wallet`, we could potentially update the `ready` field for all the wallets in the `useEffect` that's currently filtering, that way consumers will be able to do what they prefer with the wallets but have the `ready ` status of all the wallets available.

jordansexton: Yeah, this is what I'm unhappy with about the design. I don't want to modify the `Wallet` interface or mutate the instances of it, but I think something like you're talking about would be good. Extending the `Wallet` interface in the React context could work.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -42,41 +43,71 @@ export const WalletProvider: FC<WalletProviderProps> = ({
...
+            // If the wallets haven't changed while waiting, update the details
+            if (wallets === waiting) {
+                const details = wallets.reduce<Record<WalletName, WalletDetails>>((details, wallet, index) => {
+                    details[wallet.name] = { ready: ready[index] };
+                    return details;
+                }, {});
+                setDetails(details);
+            }
+        })();
+    }, [wallets]);
```
jordansexton, line 70: @danmt `details` sucks as a name, but the idea here is to provide an extensible lookup table for wallet metadata. This will be nonbreaking, the UI will still get the full list, and then can sort/filter with whatever custom logic applies. Perhaps the adapter could be expanded on to include information based on device type, etc. that would allow the app to prioritize the wallets.

danmt: I don't like `details`  as a name either but it does work for this (maybe we could use `metadata` instead)... I like the idea of a lookup table for wallet metadata because it's a non-breaking change, but I do think that having the ready/isMobile/etc... as part of the wallet when loading them is probably easier to reason about. I'm happy with either approach tho but I'm leaning a bit onto adding this information straight to the wallet and avoiding the `details` object...

If we commit to the `details` object, is there a reason why not use a `Map` instead of a plain object?

jordansexton: > I do think that having the ready/isMobile/etc... as part of the wallet when loading them is probably easier to reason about

I agree. I haven't come up with a clean way to implement this. Any ideas?

> why not use a Map instead of a plain object?

What's the benefit of using a Map?

danmt: > I agree. I haven't come up with a clean way to implement this. Any ideas?

What if... the factory functions we have per wallet (i.e. getPhantomWallet) returned a Promise instead, the responsibility of the Promise is to create a new adapter, get the ready status and return a `Wallet` that includes all the values we already have plus `ready` boolean field. This also opens the door for adding more fields in the future (like isMobile).

The reason why I like this idea is that:

1. We call ready once per wallet when loading them
2. For consumers is not a breaking change
3. There's no need for a lookup or maintaining more info in the state.

BTW - I haven't done full-time research on the pros/cons of this strategy, I'd love to hear your angle on this.

> What's the benefit of using a Map?

There's no major benefit, I like them more than objects because it has built-in methods like `get` and `has`, another cool thing is that's iterable and can be called with a `for..of` loop.

jordansexton:
Merging but won't release yet. @steveluscher feel free to weigh in on the design.

Filtering / UI component revamp left to a followup PR.

PR #207, opened 2021-12-17
fix(svelte core): autoconnect

cogoo:
autoconnect only needs to run if the adapter is ready


cogoo:
@jordansexton changed the placement for autoconnect as it needs to run when the adapter's ready state changes

PR #208, opened 2021-12-18
feat(angular): update dependencies and solve breaking changes

danmt:
This PR includes the list of changes below:

- Update ready status of wallet asynchronously.
- Improve local storage service usage.
- Update angular and every dependency relying on it to 13.
- Load wallets imperatively. (BREAKING CHANGE)
- Handle window unload event.

Here's a [repository showcasing how to use the package](https://github.com/danmt/wallet-adapter-angular-sample)


packages/core/angular/src/wallet/wallet.types.ts
```diff
@@ -1,21 +1,20 @@
...
 import { PublicKey } from '@solana/web3.js';
 
 export interface WalletState {
     wallets: Wallet[];
-    name: WalletName | null;
     wallet: Wallet | null;
     adapter: Adapter | null;
     connecting: boolean;
     disconnecting: boolean;
+    unloading: boolean;
```
jordansexton, line 10: Is it possible for this to be just an internal implementation detail / private field of the store? I don't think we want to expose it on the state.

packages/core/angular/src/wallet/local-storage.ts
```diff
@@ -1,37 +1,32 @@
...
+        const value = this._value.getValue();
+
+        if (newValue === value) return;
+
+        this._value.next(newValue);
+
+        if (newValue === null) {
+            localStorage.removeItem(this._key);
+        } else {
+            localStorage.setItem(this._key, JSON.stringify(newValue));
```
jordansexton, line 29: This try/catch block should be left as is, I think. This was a fix made in https://github.com/solana-labs/wallet-adapter/pull/199. Same situation in the getter above.

packages/core/angular/src/connection/connection.types.ts
```diff
@@ -1,5 +1,8 @@
+import { WalletAdapterNetwork } from '@solana/wallet-adapter-base';
 import { Connection } from '@solana/web3.js';
 
 export interface ConnectionState {
     connection: Connection | null;
+    network: WalletAdapterNetwork;
```
jordansexton, line 6: I don't think the `network` is actually relevant to the Connection object -- the external interface can just be `endpoint`.

packages/core/angular/src/wallet/wallet.store.ts
```diff
@@ -120,23 +125,37 @@ export class WalletStore extends ComponentStore<WalletState> {
...
+
+            concatMap((adapter) =>
+                from(defer(() => adapter.ready())).pipe(tap((ready) => this.patchState({ ready: !!ready })))
+            )
+        )
+    );
+
+    // If the window is closing or reloading, ignore disconnect and error events from the adapter
+    readonly handleUnload = this.effect(() =>
+        fromEvent(window, 'beforeunload').pipe(tap(() => this.patchState({ unloading: true })))
```
jordansexton, line 156: `window` should always be guarded for SSR / just to be safe, it's been the source of several bugs with the library.

packages/core/angular/src/wallet/wallet.store.ts
```diff
@@ -304,7 +326,9 @@ export class WalletStore extends ComponentStore<WalletState> {
                     return throwError(error);
                 }
 
-                return from(defer(() => adapter.sendTransaction(transaction, connection, options)));
+                return from(defer(() => adapter.sendTransaction(transaction, connection, options))).pipe(
+                    map((txId) => txId as string)
```
jordansexton, line 330: nit: let's still use `TransactionSignature` here (even though it's just a string)

jordansexton (commented):
Thanks, have a few notes, looks good.

packages/core/angular/src/wallet/wallet.types.ts
danmt: Agree... It's not really exposed tho, the component store has this behavior where you have to use a selector to access something from the state, given that there's no unloading selector, nothing is really being exposed. Nevertheless, I do think it makes no sense to be part of the state.

packages/core/angular/src/wallet/wallet.store.ts
danmt: you're right!

packages/core/angular/src/connection/connection.types.ts
danmt: Indeed, I just wanted to store the network somewhere in the state. I agree that the ConnectionStore is probably not the best place.

packages/core/angular/src/wallet/local-storage.ts
danmt: Got it

packages/core/angular/package.json
```diff
@@ -8,25 +8,14 @@
...
-    "main": "lib/bundles/solana-wallet-adapter-angular.umd.js",
-    "module": "lib/fesm2015/solana-wallet-adapter-angular.js",
-    "types": "lib/solana-wallet-adapter-angular.d.ts",
-    "es2015": "lib/fesm2015/solana-wallet-adapter-angular.js",
-    "esm2015": "lib/esm2015/solana-wallet-adapter-angular.js",
-    "fesm2015": "lib/fesm2015/solana-wallet-adapter-angular.js",
-    "files": [
-        "lib",
-        "src"
-    ],
```
jordansexton, line 22: Isn't removing all of this a problem for the published package consumers?

packages/core/angular/src/connection/connection.store.ts
```diff
@@ -24,11 +32,24 @@ export class ConnectionStore extends ComponentStore<ConnectionState> {
             ...this._config,
         };
 
-        this.setState({ connection: null });
+        this.setState({
+            connection: null,
+            endpoint: clusterApiUrl(WalletAdapterNetwork.Devnet),
```
jordansexton, line 37: In the React version it's required and there is no default. Why does this need a default value?

packages/core/angular/src/wallet/wallet.store.ts
```diff
@@ -103,14 +127,18 @@ export class WalletStore extends ComponentStore<WalletState> {
 
         this.setState({
             ...initialState,
-            wallets: this._config?.wallets || [],
-            name: this._localStorage.value,
+            wallets: [],
```
jordansexton, line 130: What will happen if the user doesn't call `loadWallets`? Is there any way we can get type safety that compels them to configure it correctly?

packages/core/angular/src/wallet/wallet.store.ts
```diff
@@ -120,23 +148,41 @@ export class WalletStore extends ComponentStore<WalletState> {
...
+                from(defer(() => adapter.ready())).pipe(tap((ready) => this.patchState({ ready: !!ready })))
+            )
+        )
+    );
+
+    // If the window is closing or reloading, ignore disconnect and error events from the adapter
+    readonly handleUnload = this.effect(() => {
+        if (typeof window === 'undefined') {
+            return of(null);
+        }
```
jordansexton, line 181: :+1:

packages/core/angular/src/wallet/wallet.store.ts
```diff
@@ -304,7 +353,9 @@ export class WalletStore extends ComponentStore<WalletState> {
                     return throwError(error);
                 }
 
-                return from(defer(() => adapter.sendTransaction(transaction, connection, options)));
+                return from(defer(() => adapter.sendTransaction(transaction, connection, options))).pipe(
+                    map((txId) => txId as TransactionSignature)
```
jordansexton, line 357: Is this cast still needed? I see the other ones are gone.

jordansexton (commented):
Looks like this is almost done, just have a few questions

packages/core/angular/src/wallet/wallet.store.ts
danmt: Good catch, it's not needed anymore.

packages/core/angular/package.json
jordansexton: Hmm, okay. Could this still present issues for other packages in the repo that use this locally? Thinking especially removing the `type": "module"`, sideEffects/main/module/types keys. I guess I'll just try to build it locally and see.

packages/core/angular/src/wallet/wallet.store.ts
danmt: > What will happen if the user doesn't call loadWallets?

No wallets will be shown in the modal.

> Is there any way we can get type safety that compels them to configure it correctly?

I'll give this a try!

What would be considered "correct" in this case? Calling loadWallets at least once? Have at least one wallet?

packages/core/angular/src/connection/connection.store.ts
danmt: I think in this version it should behave the same as React... The default value is not needed anyway. I think there should be a type-safe way to ensure package consumers set the endpoint. (Just like in the other comment)

packages/core/angular/package.json
danmt: `sideEffects`, `module` and `typings` are automatically added by ng-packagr. I believe angular packages don't use a `main` but rather have all those `es2020`, `esm2020`, etc... I do find it curious that the `type: "module"` is not there, it seems like its not needed... Last time I checked it was building properly and I published it under my npm org to check, the repo in the description is using it

packages/core/angular/src/wallet/wallet.store.ts
jordansexton: Yeah, maybe it's a non issue and we just solve this with documentation. Worst case behavior is that they forget to initialize and have no wallets, which they'll notice in testing. Was mostly wondering if there's a way to do this in the constructor or something so that the API is obvious.

jordansexton: Related: can we call it `setWallets`? `loadWallets` implies the wrong behavior I think (the adapters are already loaded)

danmt: My reasoning behind `load` was that I was "loading the wallets into the store", I initially had `setWallets`, so I'm okay either way.

Regarding the no wallets thing, what if I make the wallets type `Wallet[] | null` and make it null by default? That way the consumers should call `setWallets` at least once, even if it's with an empty list.

jordansexton:
Closing this since #285 has removed Angular from this repo and over to its new home: https://github.com/heavy-duty/platform/tree/master/libs/wallet-adapter

PR #211, opened 2021-12-20
Onekey Wallet Adapter

zuozhuo:


jordansexton:
Closing for the same reason as https://github.com/solana-labs/wallet-adapter/pull/191#issuecomment-995165285

Wallet adapters that duplicate other wallet adapters don't add value and they add maintenance cost.

PR #212, opened 2021-12-21
Reduce disconnection race conditions in WalletProvider#select

steveluscher:
Imagine someone calls `select()` on the `WalletProvider`. Because we `await` the last adapter's disconnection procedure _before_ setting the name of the new wallet, you can get yourself into a little bit of a race condition if they call `select()` again.

In this PR we write tests that cover much of the `WalletProvider's` functionality, including a regression test that triggers this bug, then we set about fixing the bug.

The bug fix is to simply fire off the disconnection instruction without waiting on it. This should be safe because:

1. We don't need confirmation of a disconnection to connect a new wallet.
2. By the time we fire the disconnection instruction, we've already unsubscribed from wallet adapter events.

### DIFF
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -116,10 +116,9 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     const select = useCallback(
         async (walletName: WalletName | null) => {
             if (name === walletName) return;
-            if (adapter) await adapter.disconnect();
             setName(walletName);
         },
-        [name, adapter, setName]
+        [name]
     );
 
     // Handle the adapter's connect event
@@ -269,6 +268,13 @@ export const WalletProvider: FC<WalletProviderProps> = ({
             };
         }
     }, [adapter, handleConnect, handleDisconnect, handleError]);
+
+    // When the adapter changes, disconnect the old one.
+    useEffect(() => {
+        return () => {
+            adapter?.disconnect();
+        };
+    }, [adapter]);
 
     return (
         <WalletContext.Provider
--- a/packages/core/react/src/__tests__/WalletProvider-test.tsx
+++ b/packages/core/react/src/__tests__/WalletProvider-test.tsx
@@ -1,0 +1,464 @@
+/**
+ * @jest-environment jsdom
+ */
+
+'use strict';
+
+import 'jest-localstorage-mock';
+import { render, unmountComponentAtNode } from 'react-dom';
+import React, { createRef, forwardRef, useImperativeHandle } from 'react';
+import { act } from 'react-dom/test-utils';
+import { WalletProvider, WalletProviderProps } from '../WalletProvider';
+import { BaseWalletAdapter, Wallet, WalletError, WalletName, WalletNotReadyError } from '@solana/wallet-adapter-base';
+import { PublicKey } from '@solana/web3.js';
+import { useWallet, WalletContextState } from '../useWallet';
+
+type TestRefType = {
+    getWalletContextState(): WalletContextState;
+};
+
+const TestComponent = forwardRef(function TestComponentImpl(props, ref) {
+    const wallet = useWallet();
+    useImperativeHandle(
+        ref,
+        () => ({
+            getWalletContextState() {
+                return wallet;
+            },
+        }),
+        [wallet]
+    );
+    return null;
+});
+
+describe('WalletProvider', () => {
+    let container: HTMLDivElement | null;
+    let ref: React.RefObject<TestRefType>;
+    let fooWalletAdapter: MockWalletAdapter;
+    let barWalletAdapter: MockWalletAdapter;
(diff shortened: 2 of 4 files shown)


packages/core/react/src/__tests__/WalletProvider-test.tsx
```diff
@@ -0,0 +1,464 @@
...
+                            await Promise.resolve(); // Flush all promises in effects after calling `select()`.
+                        });
+                        act(() => {
+                            commitFooWalletDisconnection();
+                        });
+                    });
+                    it('the wallet you selected last should be set in state', () => {
+                        expect(ref.current?.getWalletContextState().adapter).toBe(bazWalletAdapter);
+                    });
+                });
```
steveluscher, line 460: This is the failing test that this PR fixes.

jordansexton:
Wow, this is great. I will learn a lot from the way you've written tests for this.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -116,10 +116,9 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     const select = useCallback(
         async (walletName: WalletName | null) => {
             if (name === walletName) return;
-            if (adapter) await adapter.disconnect();
             setName(walletName);
         },
-        [name, adapter, setName]
```
steveluscher, line 122: Removed `setName` from the dependencies array. See https://github.com/solana-labs/wallet-adapter/pull/216.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -98,6 +98,13 @@ export const WalletProvider: FC<WalletProviderProps> = ({
         }
     }, [name, enhancedWalletsByName, setState]);
 
+    // When the adapter changes, disconnect the old one.
+    useEffect(() => {
+        return () => {
+            adapter?.disconnect().catch((error) => console.warn(error));
+        };
+    }, [adapter]);
+
```
steveluscher, line 107: If you move this above the hook that cleans up the adapter's event listeners, I _think_ that calling `disconnect()` on line 104 will _fire_ the disconnect event on line 164, which will have some unpredictable consequences.

steveluscher: Best to let the adapter's event handler hook disconnect the listeners, and _then_ call `disconnect` on the adapter!

jordansexton: lol yep I broke this 😅

packages/core/react/src/WalletProvider.tsx
```diff
@@ -132,10 +132,9 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     const select = useCallback(
         async (walletName: WalletName | null) => {
             if (name === walletName) return;
-            if (adapter) await adapter.disconnect();
             setName(walletName);
         },
-        [name, adapter, setName]
+        [name]
```
jordansexton, line 137: Is this function needed anymore since `setName` already bails if newName === oldName?

packages/core/react/src/WalletProvider.tsx
```diff
@@ -282,6 +281,13 @@ export const WalletProvider: FC<WalletProviderProps> = ({
         }
     }, [adapter, handleConnect, handleDisconnect, handleError]);
 
+    // When the adapter changes, disconnect the old one.
+    useEffect(() => {
+        return () => {
+            adapter?.disconnect();
```
jordansexton, line 287: Do we need to catch this if it errors?

packages/core/react/src/WalletProvider.tsx
steveluscher: Nope! The only weird part is that if you hand people the state setter directly, then instead of a method with the signature `(walletName: WalletName) => void` we'd be handing them a method with the signature `(walletName: WalletName) => void | ((prevWalletName: WalletName) => WalletName) => void)` which is maybe not what we want.

jordansexton: Well, in this case I think the state setter is https://github.com/solana-labs/wallet-adapter/blob/422d77d84bdd12dd8a999e8c6c3118128863c2e4/packages/core/react/src/useLocalStorage.ts#L15-L16 so it should be equivalent?

FWIW I tried making this change but got an error from the new tests.

packages/core/react/src/WalletProvider.tsx
steveluscher: I don't _think_ so, because I don't think that fatals in non-awaited promises won't bubble up through React. It will appear in the console, and in dev mode, but not in production.

https://codesandbox.io/s/musing-danny-nrpn8?file=/src/App.js

packages/core/react/src/WalletProvider.tsx
steveluscher: I think that #222 will fix this, but I haven't tested that out yet!

PR #213, opened 2021-12-21
[fix] The state initialization effect will no longer run if the wallet list has changed

steveluscher:
Because we await a series of promises in this effect, an effort was made to _prevent_ the state from being set if the list of wallets happens to change while the promise was in flight.

Because of the way that bindings work in React, though, this didn't accomplish what we hoped. The `wallets === waiting` test would _always_ pass because the `wallets` binding is immutable for the life of that effect.

What we can do instead is to track wether the effect was ever invalidated by hooking into its teardown. If it has, we can skip setting the enhanced wallets because one of two things have happened.

1. We're unmounting.
2. `wallets` changed so we'll be called again.


jordansexton:
Good catch!

PR #214, opened 2021-12-21
Add wallet detection metadata & implementation

steveluscher:
_This PR is a slightly different take on #206. Please, give me your feedback. If you like this direction, I can continue to add this to the other non-React frameworks_.

# Overview

We want to enable developers' apps to surface the most useful wallets first. In particular, developers are likely going to want to show their users' _installed_ wallets first.

In this PR, we introduce two new concepts.

## Concept 1: Detected wallets

We make a `Set` available on the `WallerProviderContextState` that contains the names of every wallet we know to be installed and ready. Developers can make use of this list to prioritize which wallets to show, and in which order.

## Concept 2: Detectability

Some wallets are _not installable_ which makes presence detection sort of irrelevant. For instance, you can install Phantom wallet as an extension, but Torus is an SDK that you can always choose to invoke on-demand. We _always_ consider Torus _ready_ since we can invoke it at any time, but that doesn't imply that it's _installed_. The `WalletDetectability` construct communicates this to developers, so that they can make an informed choice of how and when to display any given wallet.

# Example

In the final commit of this PR we give an example of how developers might use this API with React. Given a browser with Phantom and Solflare installed, this is what the wallet modal looks like.

image


steveluscher:
Ignore the bottom two commits in this PR; they're reviewable in https://github.com/solana-labs/wallet-adapter/pull/212.

packages/wallets/blocto/src/wallet.ts
```diff
@@ -4,6 +4,7 @@ import { BloctoWalletAdapter, BloctoWalletAdapterConfig } from './adapter';
 export const BloctoWalletName = 'Blocto' as WalletName;
 
 export const getBloctoWallet = (config: BloctoWalletAdapterConfig = {}): Wallet => ({
+    detectability: 'runtime-only',
```
steveluscher, line 7: Depending on the runtime (eg. a desktop browser vs. a mobile webview) we can return a _different_ detectability. For instance, Blocto [says](https://docs.blocto.app/blocto-app/web3.js-integration-solana):

> Blocto injects web3 provider to web context as window.solana starting from version 2.9.0 on Android/iOS. To detect Blocto specifically, there is a isBlocto flag.

We could return `'installable'` on iOS and Android in this case.

jordansexton:
> We want to enable developers' apps to surface the most useful wallets first. In particular, developers are likely going to want to show their users' _installed_ wallets first.

Definitely. For context, here's the design that these features are working toward:
Option 3
Option 3(1)

> We make a `Set` available on the `WallerProviderContextState` that contains the names of every wallet we know to be installed and ready. Developers can make use of this list to prioritize which wallets to show, and in which order.

I think it may be good to expose these properties on the `Wallet` or `EnhancedWallet` interface rather than use a set. The set implemented implies that the core React package has an opinion about how these properties are used. I think this can be left up to the UI packages, but let me know if you disagree.

> Some wallets are _not installable_ which makes presence detection sort of irrelevant.

Agreed, I wanted to do this as a followup, with platform detection at least. Some wallets have extensions, mobile apps, work in only some browsers, or work better for different user flows. I think we should try to detect and surface as much of this information as we can in the adapters and core, and then have a default UI that has an opinion on how to make sense of it.

> For instance, you can install Phantom wallet as an extension, but Torus is an SDK that you can always choose to invoke on-demand. We _always_ consider Torus _ready_ since we can invoke it at any time, but that doesn't imply that it's _installed_.

Torus is a bit tricky because it's hard to say what it means to be "installed". I think we have several possibilities --

1. Installed (as an extension on desktop, or as an app on mobile or desktop)
2. Installable (as an extension or app, or as an app on mobile or desktop)
3. Runnable in a browser (e.g. Torus, Blocto, Sollet web popup, WalletConnect; lazy loads an SDK)
4. Hardware (Ledger; lazy loads an SDK; maybe just a special case of 3, but this could change with LedgerLive support)

Can you think of any more? Can any of these be reduced to one of the others?

Based on this, I think we would want to:

1. Display installed wallets (1) first
2. Display runnable wallets (3/4) for the current environment next
3. Display installable wallets (2) for the current platform next
4. Not display wallets that are not installable or runnable for the current platform/env

Because the set of displayed wallets here could still be way too many choices large (on a desktop browser, it's all of the browser-compatible extensions + runnable wallets) we still need to figure out how to filter and prioritize the list, and maybe not display some of them.

packages/wallets/blocto/src/wallet.ts
jordansexton: Returning a platform-specific value is what I was thinking as well. Could implement it as a getter, unless there's a reason to do it async? The `url` field could also be platform-specific, along with the general behavior of `connect`. For example, a dapp running in a mobile browser could launch a link to a mobile wallet instead of using an extension, possibly using the same adapter.

steveluscher: I was thinking even something as simple as…

```
export const getBloctoWallet = (config: BloctoWalletAdapterConfig = {}): Wallet => ({
    detectability: somePlatformCheck() ? 'runtime-only' : 'installable',
    /* ... */
});
```

steveluscher:
> I think it may be good to expose these properties on the Wallet or EnhancedWallet interface rather than use a set. The detectedWallets set implies that the core React package has an opinion about how these properties are used.

That makes a lot of sense. In that case, this property will need to be observable one way or another so that different frameworks have a signal to hook into when it changes. How about we…

* Offer an enum (maybe called `readyState`) on the adapter that can be in one of three states
  * `detected` – we've detected the thing
  * `undetected` – we are either still trying to detect the thing or we failed to detect it
  * `runtime-only` – this isn't the kind of wallet that can be installed/detected
* Add an adapter event – `readyStateChange` – and emit it when we update that value. Frameworks can listen to this event and force an update.

The `detectability` property would go away if we chose this design. Now that `readyState` is an enum, it's expressive enough to communicate ‘this wallet can be installed and furthermore we detected it’ as well as ‘this isn't a wallet that can be installed.’

Shall I take a run at that?

> Torus is a bit tricky because it's hard to say what it means to be "installed". I think we have several possibilities…

Interesting. Do you think it makes sense, with a wallet like Torus to:

* Flip `readyState` to `detected` if we happen to detect the extension
* Otherwise leave it as `runtime-only` since there's always a runtime SDK option available.

Contrast this with Phantom where either you're `detected` or you're `undetected`.

I _think_ having a tri-state enum is expressive enough to answer clients' questions about what wallets to show and in what order.

> Display installed wallets (1) first

```
wallets.filter(w => w.adapter?.readyState === 'detected');
```

> Display runnable wallets (3/4) for the current environment next

```
wallets.filter(w => w.adapter?.readyState === 'runtime-only');
```

> Display installable wallets (2) for the current platform next

```
wallets.filter(w => w.adapter?.readyState === 'undetected');
```

> Not display wallets that are not installable or runnable for the current platform/env

We could achieve this either by setting the `adapter` property to `null` if the platform doesn't support the wallet, or by adding a fourth `readyState` enum value, `unsupported`.

jordansexton:
Yeah, I think some kind of enum is the way to go. I was playing with

```
export enum WalletType {
    Injected,
    Library,
    Hardware,
}
```

`type` plus the `ready` boolean allows downstream consumers to decide how to treat it (or if they even want to support it). If something is ready + injected, they might show "detected". If it's ready + library/hardware, they might show it if nothing else is detected.

I'm not sure if this is too brittle -- could a wallet support more than one type, or could it change asynchronously? I feel like the answer is yes. I'm thinking of Torus (library vs injected_ and Ledger (hardware vs Ledger Live, which... I'm not sure how that works) here. To your suggestion, it might make sense to just refactor this all into a more flexible `readyState`.

I've also been rethinking whether `Wallet` even makes sense as its own type. Should we just roll it all into the `WalletAdapter`?

jordansexton:
> * Offer an enum (maybe called `readyState`) on the adapter that can be in one of three states
>   
>   * `detected` – we've detected the thing
>   * `undetected` – we are either still trying to detect the thing or we failed to detect it
>   * `runtime-only` – this isn't the kind of wallet that can be installed/detected
> 
> * Add an adapter event – `readyStateChange` – and emit it when we update that value. Frameworks can listen to this event and force an update.

Overall, this approach sounds very good, especially if we include `unsupported`. I think `runtime-only` might be confusing as a name though.

jordansexton:
Are you concerned at all about a bunch of separate state updates to the wallets array caused by `readyStateChange` firing on each one?

metallicalfa:
I agree with the use of a enum on the adapter, and allowing its consumers to be opinionated in cases where the state isn't binary (detected, undetected). 

> > * Offer an enum (maybe called `readyState`) on the adapter that can be in one of three states
> >   
> >   * `detected` – we've detected the thing
> >   * `undetected` – we are either still trying to detect the thing or we failed to detect it
> >   * `runtime-only` – this isn't the kind of wallet that can be installed/detected
> > * Add an adapter event – `readyStateChange` – and emit it when we update that value. Frameworks can listen to this event and force an update.
> 
> Overall, this approach sounds very good, especially if we include `unsupported`. I think `runtime-only` might be confusing as a name though.

`unsupported` param will help a lot filtering on browsers like safari on desktop and mobile. Instead of using `runtime-only` maybe we can use `onDemand`. (Chrome extension wallets are injected on page load, web wallets are injected on demand)

We don't necessarily need a new adapter event though. As /core can handle the state changes, and update the enhanced wallets in the same way its doing now. We would still need a `ready` state (same as what we have now)

> I'm not sure if this is too brittle -- could a wallet support more than one type, or could it change asynchronously? I feel like the answer is yes. I'm thinking of Torus (library vs injected) and Ledger (hardware vs Ledger Live, which... I'm not sure how that works) here. To your suggestion, it might make sense to just refactor this all into a more flexible `readyState`.

Yes, this would be the case when more wallets come up with their own native app, which might use walletConnect or some bridge to connect. Currently, ledger live isn't supported for solana transactions (could change in the future). In such cases, we can update the adapter's detect method without changing the enum. There are mobile wallets with dapp browsers (a native webview) to consider as well.

steveluscher:
> Are you concerned at all about a bunch of separate state updates to the wallets array caused by readyStateChange firing on each one?

I don't think so. The latest version of React [batches state updates](https://github.com/reactwg/react-18/discussions/21), even outside of event handlers, so slamming the `setState` API serially soon won't even be distinguishable from batching the updates.

steveluscher:
> We don't necessarily need a new adapter event though. As /core can handle the state changes, and update the enhanced wallets in the same way its doing now.

Oh, updating the property is no problem and doesn't need an event. The problem is that the value in _core_ is not ‘reactive,’ in the sense that updating it won't cause (for instance) a React re-render. We need an event that a framework like React can hook into to know that something in the `Wallet/Adapter` object has changed.

steveluscher:
> I'm not sure if this is too brittle -- could a wallet support more than one type, or could it change asynchronously?

I _think_ that asynchronous changes can be covered off by offering a `readyStateChange` event, and I think that we can probably boil every state down to a single enum value. For instance, even if a wallet came in on-demand and installable flavours, if it's installed and been detected, we might as well just call it `detected`, right?

steveluscher:
> Instead of using runtime-only maybe we can use onDemand.

That's better! On-demand… loadable… let me give this PR an overhaul and we can come back for another round of review.

steveluscher (commented):
I got this to the point where I hate it, so that probably means it's close.

I think @jordansexton's `EnhancedWallet` interface is the way to go; I regret now having rolled that back in the current iteration of this PR. It gives us a nice airgap between the mutable adapters and the immutable, reactive state of React/Vue/etc.

The React starter example works. I haven't tested the Vue/Svelte/Angular ones.

packages/core/base/src/adapter.ts
```diff
@@ -31,6 +32,34 @@ export interface WalletAdapterProps {
...
+    /**
+     * Loadable wallets are always available to you. Since you can load them at
+     * any time, it's meaningless to say that they have been detected.
+     */
+    | 'loadable'
+    /**
+     * If a wallet is not supported on a given platform (eg. server-rendering, or
+     * mobile) then it will stay in the `unsupported` state.
+     */
+    | 'unsupported';
```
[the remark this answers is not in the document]
jordansexton: Looks very good, I like "loadable". Can we use a TS enum with these string values?

packages/core/base/src/adapter.ts
```diff
@@ -54,3 +83,16 @@ export enum WalletAdapterNetwork {
...
+    if (document.readyState === 'complete') {
+        performDetection();
+        return;
+    }
+    function listener() {
+        window.removeEventListener('load', listener);
+        performDetection();
+    }
+    window.addEventListener('load', listener);
+}
```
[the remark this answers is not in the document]
jordansexton: Yeah, I think this is good. Why does `performDetection()` return a value?

packages/core/react/src/WalletProvider.tsx
```diff
@@ -23,15 +23,13 @@ export interface WalletProviderProps {
 }
 
 const initialState: {
-    wallet: EnhancedWallet | null;
+    wallet: Wallet | null;
```
jordansexton, line 26: wdyt, can we roll the base Wallet change into this? I figure if we're going to make that change we might as well do it now.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -44,63 +42,62 @@ export const WalletProvider: FC<WalletProviderProps> = ({
...
     const [disconnecting, setDisconnecting] = useState(false);
     const isConnecting = useRef(false);
     const isDisconnecting = useRef(false);
     const isUnloading = useRef(false);
 
-    // When the wallets change, enhance them with additional properties
-    const [enhancedWallets, setEnhancedWallets] = useInitialState(
-        () => wallets.map((wallet) => ({ ...wallet, ready: false })),
+    // Map of wallet names to wallets
+    const walletsByName = useMemo(
```
jordansexton, line 54: Yeah I think you're right that we don't need this.

packages/core/base/src/adapter.ts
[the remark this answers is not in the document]
steveluscher: Nice, you noticed!

When we introduce more aggressive polling, the return value will let us know when to _stop_ polling, because the wallet was detected.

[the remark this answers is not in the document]
jordansexton: Ah, good call. I forgot `pollUntilReady` used to do just that.

packages/core/base/src/adapter.ts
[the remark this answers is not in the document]
steveluscher: Oh, for sure. This was just me, writing Typescript as though it's Flow. Derp.

packages/core/react/src/WalletProvider.tsx
steveluscher: This PR's getting a bit out of control already. How about we bring back `EnhancedWallet`/`WalletWithReadyState` now, and then dissolve the whole thing back into `WalletAdapter` in a future one?

jordansexton: Works for me!

packages/core/react/src/WalletProvider.tsx
```diff
@@ -44,63 +39,72 @@ export const WalletProvider: FC<WalletProviderProps> = ({
...
-        const adapter = wallet && wallet.adapter;
-        if (adapter) {
-            setState({
-                wallet,
-                adapter,
-                ready: enhancedWalletsByName[wallet.name].ready,
-                connected: adapter.connected,
-                publicKey: adapter.publicKey,
-            });
+        const wallet = walletsWithReadyState.find(({ name: walletName }) => walletName === name);
```
steveluscher, line 87: Replaced `walletsByName` with this.

packages/core/react/src/useWallet.ts
```diff
@@ -5,20 +5,18 @@ import {
...
 
-export interface EnhancedWallet extends Wallet {
-    ready: boolean;
+export interface WalletWithReadyState extends Wallet {
+    readyState: WalletReadyState;
 }
-
-export interface WalletContextState extends Omit<WalletAdapterProps, 'ready'> {
-    wallets: EnhancedWallet[];
-    walletsByName: Record<WalletName, EnhancedWallet>;
```
steveluscher, line 17: Took a chance and nerfed `walletsByName` from the context.

packages/core/vue/src/useWallet.ts
```diff
@@ -88,57 +89,67 @@ export const createWalletStore = ({
...
+        wallets.map((wallet) => ({
+            ...wallet,
+            readyState: wallet.adapter.readyState,
+        }))
+    );
+
+    // When the wallets change, start to listen for changes to their `readyState`.
+    watch(wallets, (_oldWallets, _newWallets, onInvalidate) => {
+        function handleReadyStateChange(this: Wallet, nextReadyState: WalletReadyState) {
+            // FIXME: Did not test this; don't know if this is the Vue-y way to achieve this.
```
steveluscher, line 124: I have no idea if this works or not, or if any of this makes sense in Vue. I'd have to dive deeper.

steveluscher (commented):
Alright. All fixed up.

* Re-introduced wrapper interface (renamed from `EnhancedWallet` to `WalletWithReadyState`)
* Removed index between `WalletName` => `Wallet` from context.
* Used a first-class TS enum for `WalletReadyState`

@jordansexton, feel free to throw away and ground-up the Vue/Svelte/Angular commits. I was mostly coding blind and making stuff up.

jordansexton:
@steveluscher I rolled a bunch of Adapter/Wallet changes into this and merged to keep this from running too long, but I think there are some revisions we may still want to follow up with before it's ready for release.

I'm not 100% happy with the new `Wallet` (formerly `EnhancedWallet` / `WalletWithReadyState`) interface. It seems too thin, and doing `wallet.adapter.name`, `wallet.adapter.icon`, etc. is awkward. Probably we need to create some clearer definition around what it represents and how it makes sense for devs to interact with it.

packages/core/vue/src/useWallet.ts
jordansexton: Because of the structural changes to the base package, I just reverted all the Angular, Svelte, and Vue changes for now and kept them on the current version of the base package. We can update them for consistency when we're confident in the API.

packages/core/react/src/useWallet.ts
jordansexton: Yeah, that's the right call, we were exposing too much.

packages/wallets/sollet/src/adapter.ts
```diff
@@ -45,6 +49,20 @@ export class SolletWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+            } else {
+                scopePollingDetectionStrategy(() => {
+                    if (typeof window.sollet?.postMessage === 'function') {
+                        this._readyState = 'installed';
+                        this.emit('readyStateChange', this._readyState);
+                        return true;
+                    }
+                    return false;
+                });
+            }
```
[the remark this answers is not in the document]
jordansexton: Good call out, I'll test it out now that everything is building.

PR #216, opened 2021-12-21
[tidy] Remove state setters from dependency arrays

steveluscher:
In React, state setters are always guaranteed to be stable. The docs recommend omitting them from hook dependencies.

> React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list.

https://reactjs.org/docs/hooks-reference.html#usestate

Eventually, we should install the `rules-of-hooks` and `exhaustive-deps` linters to catch things like this!

https://reactjs.org/docs/hooks-rules.html

### DIFF
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -96,7 +96,7 @@ export const WalletProvider: FC<WalletProviderProps> = ({
         } else {
             setState(initialState);
         }
-    }, [name, enhancedWalletsByName, setState]);
+    }, [name, enhancedWalletsByName]);
 
     // If autoConnect is enabled, try to connect when the adapter changes and is ready
     useEffect(() => {
@@ -116,7 +116,7 @@ export const WalletProvider: FC<WalletProviderProps> = ({
                 isConnecting.current = false;
             }
         })();
-    }, [isConnecting, connecting, connected, autoConnect, adapter, ready, setConnecting, setName]);
+    }, [isConnecting, connecting, connected, autoConnect, adapter, ready]);
 
     // If the window is closing or reloading, ignore disconnect and error events from the adapter
     useEffect(() => {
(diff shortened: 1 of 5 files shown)


packages/core/react/src/WalletProvider.tsx
```diff
@@ -132,23 +132,22 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     const select = useCallback(
         async (walletName: WalletName | null) => {
             if (name === walletName) return;
-            if (adapter) await adapter.disconnect();
```
jordansexton, line 135: This seems to be missing the adapter disconnect effect from #212

steveluscher: Bah! Unintentional commit.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -135,20 +135,20 @@ export const WalletProvider: FC<WalletProviderProps> = ({
             if (adapter) await adapter.disconnect();
             setName(walletName);
         },
-        [name, adapter, setName]
+        [name]
```
jordansexton, line 138: Missing `adapter` as a dependency now 😅 Otherwise looks good to merge! 

And, if you do want to add a lint rule for this, that sounds good to me.

steveluscher: dammit-damn-it

jordansexton:
Thank you!

steveluscher:
Mmm… @jordansexton, this might need to be backed out. I realized from our conversation in #212 that `setName` is _not_ stable, because of the way that `useLocalStorage` is implemented.

How about backing this out (or replacing all the `setName` deps) and then I'll redo it after I fix `useLocalStorage` to return a lifetime-stable setter.

jordansexton:
Yeah, no worries. Reverted for now.

jordansexton:
Good catch though, I see `setLocalStorage` changes every time `value` changes

steveluscher:
Well, the good news is that it was the tests in #212 that caught this! :)

steveluscher:
The same applies for refs, by the way. Refs are stable for the lifetime of the component so adding a ref to a dependency array won't do anything.

If you want to invalidate a hook based on the _value_ of a ref, you have to do this:

```
useEffect(() => {
  // ...
}, [isConnecting.current]);
```

PR #222, opened 2021-12-22
[enhancement] The setter in `useLocalStorage` is now stable across renders

steveluscher:
Typically, state setters are stable across renders in React. This means, among other things, that you don't need to include them as a hook dependency (eg. to `useEffect` or `useMemo`).

> React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list. – https://reactjs.org/docs/hooks-reference.html#usestate

I got into a little trouble in #216 because I assumed this to be the case for the state setter returned by our custom  `useLocalStorage` hook. This was not the case; it changed every time that the value or key changed.

In this PR, we refactor `useLocalStorage` so that the state setter _is_ stable across renders. This should make things more efficient overall, and prevent folks from making the same mistake I made above.

```
# Tests still pass after the refactor.
(cd packages/core/react; yarn test)
```

### DIFF
--- a/packages/core/react/src/useLocalStorage.ts
+++ b/packages/core/react/src/useLocalStorage.ts
@@ -1,7 +1,7 @@
-import { useCallback, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
-export function useLocalStorage<T>(key: string, defaultState: T): [T, (newValue: T) => void] {
-    const [value, setValue] = useState<T>(() => {
+export function useLocalStorage<T>(key: string, defaultState: T): [T, React.Dispatch<React.SetStateAction<T>>] {
+    const state = useState<T>(() => {
         try {
             const value = localStorage.getItem(key);
             if (value) return JSON.parse(value) as T;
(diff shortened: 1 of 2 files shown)


steveluscher:
cc/ @jordansexton

packages/core/react/src/useLocalStorage.ts
```diff
@@ -12,23 +12,22 @@ export function useLocalStorage<T>(key: string, defaultState: T): [T, (newValue:
...
+            } else {
+                localStorage.setItem(key, JSON.stringify(state[0]));
             }
-        },
-        [value, setValue, key]
-    );
+        } catch (error) {
+            console.error(error);
+        }
+    }, [state[0]]);
```
jordansexton, line 30: This makes a ton of sense. Using `useEffect` to respond to changes in the state is conceptually cleaner than hybridizing the state setter.

Why do you use `state[0]`?

steveluscher: I wanted to pass the (state, stateSetter) tuple _straight_ through, rather than to unwrap it by destructuring only to re-wrap back up the same way but in a new array.

jordansexton: Right, I just didn't know if there was a meaningful perf benefit from it?

steveluscher: Miniscule. Just wanted to avoid creating and GC-ing yet-another-array.

PR #223, opened 2021-12-22
update examples for recent blockhash

CalebEverett:
This just adds a recent block hash to the send transaction and also expands the material ui starter example to include the send transaction button.


packages/starter/example/components/SendTransaction.tsx
```diff
@@ -17,18 +17,18 @@ export const SendTransaction: FC = () => {
 
         let signature: TransactionSignature = '';
         try {
-            const transaction = new Transaction().add(
+            let recentBlockhash = await connection.getRecentBlockhash();
+            const transaction = new Transaction({ recentBlockhash: recentBlockhash.blockhash, feePayer: publicKey }).add(
```
jordansexton, line 21: Hmm I'm not sure why this is necessary. We generally set this value if it's not provided in `sendTransaction`:

https://github.com/solana-labs/wallet-adapter/blob/a10828d74c3fae04e40722a3edc50e596f24b1b5/packages/core/base/src/signer.ts#L22

Can you say more about what's failing?

CalebEverett: I was just connecting a Phantom wallet and hitting the send button and it was failing, asking for the recent blockhash.

jordansexton: Got it, thanks. Sounds like there's an underlying bug in the current implementation, I'll check it out.

jordansexton: This has been fixed: https://github.com/solana-labs/wallet-adapter/blob/d26322bf91216583bcb983dab677f000c2c2d941/packages/wallets/phantom/src/adapter.ts#L190-L193

PR #228, opened 2021-12-27
fix: Adapter Ready State

metallicalfa:
The adapters for blocto, torus and wallet connect are broken on master because the tertiary check for unsupported wallets (WalletReadyState.Unsupported) are implemented in reverse.


jordansexton:
Thanks!

PR #230, opened 2021-12-28
feat: Added Wallet Filtering Based On Wallet Detection, Ready States

metallicalfa:
In this PR,

- Filtering mechanism
- Added No Wallet Detection Flow.
- Added additional mobile device checks (optional).

Filter Order -

[Installable.first, Loadable.first, Installable.rest, Loadable.rest, Unsupported]


jordansexton:
Closing since these changes are already part of #196 and this would cause a conflict if merged.

PR #261, opened 2022-01-14
Fix ERR_UNSUPPORTED_DIR_IMPORT when using material UI

fiddledsticks:
Hi all, if this isn't an issue and it's just a me problem, feel free to close this.

That said I tried to use the Next.js starter except swapping React UI for Material UI but kept running into an ERR_UNSUPPORTED_DIR_IMPORT error which pointed to these lines:

```
import { styled } from '@mui/material/styles';
```

The fix was to import from the base directory. I believe alternatively you can append the file extension.

### DIFF
--- a/packages/ui/material-ui/src/WalletDialog.tsx
+++ b/packages/ui/material-ui/src/WalletDialog.tsx
@@ -11,7 +11,7 @@ import {
     ListItem,
     Theme,
 } from '@mui/material';
-import { styled } from '@mui/material/styles';
+import { styled } from '@mui/material';
 import { WalletName } from '@solana/wallet-adapter-base';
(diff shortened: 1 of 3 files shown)


jordansexton:
Thanks for the fix! Looks good to me. FWIW, the example starter project uses Next with Material, so I'm not sure what the issue is.

jordansexton:
I'll publish a patch release tomorrow!

fiddledsticks:
Thanks @jordansexton, I was worried it was something I was doing, I was thinking SSR related, but nothing was working so ended up patching this. Thanks again!

jordansexton:
Published:
 - @solana/wallet-adapter-material-ui-starter@0.13.0
 - @solana/wallet-adapter-material-ui@0.16.3

PR #264, opened 2022-01-16
Polyfill Node APIs for compatibility with Webpack 5 and Create React App 5

### LINKED ISSUE #241: Create React App is broken by lack of default polyfills in Webpack 5
Edit by @jordansexton:

To anyone looking for a very good workaround right now, don't use Create React App. Just use the [react-ui-starter](https://github.com/solana-labs/wallet-adapter/tree/master/packages/starter/react-ui-starter) or [material-ui-starter](https://github.com/solana-labs/wallet-adapter/tree/master/packages/starter/material-ui-starter) projects with Parcel, or the [nextjs-starter](https://github.com/solana-labs/wallet-adapter/tree/master/packages/starter/nextjs-starter) project with Next.js, save yourself a huge hassle, and get much faster builds with SWC instead of Babel.

---

**Describe the bug**
Broken build with CRA v5, unable to resolve modules.

**To Reproduce**
1. Initialize a new CRA project. `npx create-react-app my-app-test`
2. Install and use any wallet-adapter packages
3. See error

```
BREAKING CHANGE: The request './adapter' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

**Solution**
Relative imports must include a file extension to follow ESM strict mode requirements introduced in Webpack 5. It's valid typescript to include `.js` extensions as part of the import path 🤯 :

example.ts

```typescript
import { Example } from './example.js';
```
compiles to

example.js
```javascript
import { Example } from './example.js';
```

steveluscher:
The new version of Webpack – and by extension the new version of Create React App – no longer includes polyfills for Node APIs by default. This means that if your browser-based app requires modules like `assert` and `stream`, Webpack will fail to find them and fail to build.

In this PR, we alias the modules that the wallet adapter depends upon to their related browser polyfills.

I've also followed up with each repository to let them know that they might like to eliminate or polyfill these modules in their own browser builds.

* https://github.com/ethereumjs/ethereumjs-monorepo/pull/1638/
* https://github.com/crypto-browserify/cipher-base/pull/16

If you're having trouble using the Solana wallet adapter with Create React App 5, can you patch this PR in and report back? I might have missed some polyfills.

Fixes #241.

### DIFF
--- a/packages/wallets/ledger/src/polyfills/Buffer.ts
+++ b/packages/wallets/ledger/src/polyfills/Buffer.ts
@@ -1,0 +1,7 @@
+import { Buffer } from 'buffer';
+
+if (typeof window !== 'undefined' && window.Buffer === undefined) {
+    (window as any).Buffer = Buffer;
+}
+
+export {};
--- a/packages/wallets/ledger/package.json
+++ b/packages/wallets/ledger/package.json
@@ -27,6 +27,7 @@
         "postbuild": "echo '{\"type\":\"commonjs\"}' | npx json > lib/cjs/package.json && echo '{\"type\":\"module\"} ' | npx json > lib/esm/package.json"
     },
     "dependencies": {
+        "buffer": "npm:buffer-browserify@^0.2.5",
         "@ledgerhq/hw-transport": "^6.11.2",
         "@ledgerhq/hw-transport-webhid": "^6.11.2",
         "@solana/wallet-adapter-base": "^0.9.1",
--- a/packages/wallets/ledger/src/adapter.ts
+++ b/packages/wallets/ledger/src/adapter.ts
@@ -1,3 +1,5 @@
+import './polyfills/index';
+
 import type Transport from '@ledgerhq/hw-transport';
 import {
     Adapter,
--- a/packages/wallets/ledger/src/polyfills/index.ts
+++ b/packages/wallets/ledger/src/polyfills/index.ts
@@ -1,0 +1,3 @@
+'use strict';
+
+import './Buffer';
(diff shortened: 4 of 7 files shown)


jordansexton:
I like this. This looks like a pretty clean solution to an annoying problem, and we can gradually phase it out as a nonbreaking change when the dependencies change. I haven't tested this yet, but does it fix our CRA problem?

steveluscher:
I was able to boot the old CRA-based starter with this PR and CRA5. Still would love a second person to verify that though!

packages/wallets/ledger/src/polyfills/Buffer.ts
```diff
@@ -0,0 +1,7 @@
+import { Buffer } from 'buffer';
+
+if (window.Buffer === undefined) {
```
jordansexton, line 3: Hmm do we need to check if `typeof window !== 'undefined'` first?

steveluscher: I could do that! I think I just figured that if someone is using a wallet adapter in a non-browser context, then they're gonna have a bad time anyway, but you're probably right. I could see this breaking in SSR or something.

jordansexton:
Heroic work, thanks for this.

steveluscher:
🤞🏻

jordansexton:
Published:
 - @solana/wallet-adapter-ledger => 0.9.3
 - @solana/wallet-adapter-torus => 0.11.4

Issue #256, opened 2022-01-11
VueJS "Wallet not initialized" error 


NViktors:
**Describe the bug**
Hello,
I'm trying to implement "Connect wallet" functionality inside my Vue 3 project.
All required packages are up to date and have latest versions.

How to reproduce:
1. Create fresh project using Vue CLI (with typescript enabled) - `vue create test-app && cd test-app`
2. Add following dependencies: `yarn add @solana/wallet-adapter-base @solana/wallet-adapter-wallets @solana/wallet-adapter-vue @solana/wallet-adapter-vue-ui`
3. Replace App.vue with following
```
<template>
  <wallet-provider auto-connect :wallets="wallets">
    <wallet-modal-provider>
      <wallet-multi-button></wallet-multi-button>
    </wallet-modal-provider>
  </wallet-provider>
</template>

<script lang="ts">
  import {defineComponent} from 'vue';
  import {WalletProvider} from "@solana/wallet-adapter-vue";
  import {PhantomWalletAdapter} from "@solana/wallet-adapter-wallets";
  import {WalletModalProvider, WalletMultiButton} from "@solana/wallet-adapter-vue-ui";

  export default defineComponent({
    name: 'App',
    components: {
      WalletProvider,
      WalletMultiButton,
      WalletModalProvider,
    },
    setup() {
      // Just one wallet for now 
      const wallets = [
        new PhantomWalletAdapter()
      ];

      return {wallets};
    }
  });
</script>
```
4. Start development mode `yarn serve`

And after opening page in browser, there will be error in console:
![image](https://user-images.githubusercontent.com/26169825/148964166-7a679f6f-13f4-4ee8-9fd2-0418bf9274f1.png)

Any ideas about this?

lorisleiva:
Hmm, that's strange. This code should work. 🤔

Could you try using the global store approach to see if that's working?

```html
<script setup>
import { initWallet } from '@solana/wallet-adapter-vue';
import { getPhantomWallet } from '@solana/wallet-adapter-wallets';
const wallets = [getPhantomWallet()];
initWallet({ wallets, autoConnect: true });
</script>

<template>
    <!-- `useWallet` is available everywhere. Not just in the descendants of this component. -->
</template>
```

florentchauveau:
I do not have the error with the global store approach, but another error when using the wallet:

SFC:

```vue
<script setup>
import { computed } from 'vue'
[...]

PR #266, opened 2022-01-17
Updated wallet-adapter-vue to use the newest versions of its dependencies.

Randy808:
The code was using [@solana/wallet-adapter-wallets@0.13.1 ](https://github.com/solana-labs/wallet-adapter/blob/809ce18a2be4ef96c79bb3331386638efa6f17d9/packages/wallets/wallets/package.json) which seems to have never been published (the code was using references to 'Wallet' [which could be retrieved using getWallets in that version](https://github.com/solana-labs/wallet-adapter/blob/809ce18a2be4ef96c79bb3331386638efa6f17d9/packages/wallets/wallets/src/wallets.ts)), so I changed it to use the newest version of its dependencies. I also changed references to @vue/runtime-core to point to vue.

More context in #256

### DIFF
--- a/packages/core/vue/package.json
+++ b/packages/core/vue/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@solana/wallet-adapter-vue",
-    "version": "0.4.4",
+    "version": "0.5.0",
     "author": "Solana Maintainers <maintainers@solana.foundation>",
     "repository": "https://github.com/solana-labs/wallet-adapter",
     "license": "Apache-2.0",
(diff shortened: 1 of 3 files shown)


Randy808:
I didn't change all the places where the correct adapter needs to be in this PR and the event state change for the wallet is not handled appropriately. All this does is removes the provide warning message and ensures the store is there for the inject call. I have a version locally that somewhat works for my use-case but I'll close this issue since I'm not sure if I'll have time to dedicate to a full PR

Randy808:
Re-opened as PR #267

PR #267, opened 2022-01-18
Updated wallet-adapter-vue to use the newest versions of its dependen…

Randy808:
Re-make of a closed PR:

The code was using @solana/wallet-adapter-wallets@0.13.1 which seems to have never been published (the code was using references to 'Wallet' which could be retrieved using getWallets in that version), so I changed it to use the newest version of its dependencies. I also changed references to @vue/runtime-core to point to vue.

More context in #256 

*Decided to keep the PR up to try and get this officially fixed and not use workarounds to have my project use unofficial packages

### DIFF
--- a/packages/core/vue/src/useWallet.ts
+++ b/packages/core/vue/src/useWallet.ts
@@ -1,29 +1,28 @@
 import {
-    Adapter,
-    MessageSignerWalletAdapter,
     SendTransactionOptions,
-    SignerWalletAdapter,
-    Wallet,
+    Adapter,
     WalletError,
     WalletName,
     WalletNotConnectedError,
     WalletNotReadyError,
+    WalletReadyState,
+    SignerWalletAdapter,
+    MessageSignerWalletAdapter
 } from '@solana/wallet-adapter-base';
 import { Connection, PublicKey, Transaction, TransactionSignature } from '@solana/web3.js';
-import { computed, inject, InjectionKey, provide, Ref, ref, watch, watchEffect } from '@vue/runtime-core';
+import { computed, inject, InjectionKey, provide, Ref, ref, watch, watchEffect } from 'vue';
 import { WalletNotSelectedError } from './errors';
 import { useLocalStorage } from './useLocalStorage';
 
 export interface WalletStore {
     // Props.
-    wallets: Wallet[];
+    wallets: Adapter[];
     autoConnect: boolean;
 
     // Data.
-    wallet: Ref<Wallet | null>;
-    adapter: Ref<Adapter | null>;
+    wallet: Ref<Adapter | null>;
     publicKey: Ref<PublicKey | null>;
-    ready: Ref<boolean>;
+    ready: Ref<WalletReadyState>;
     connected: Ref<boolean>;
     connecting: Ref<boolean>;
     disconnecting: Ref<boolean>;
@@ -44,7 +43,7 @@ export interface WalletStore {
 }
 
 export interface WalletStoreProps {
-    wallets: Wallet[];
+    wallets: Adapter[];
     autoConnect?: boolean;
     onError?: (error: WalletError) => void;
     localStorageKey?: string;
@@ -69,29 +68,26 @@ export const initWallet = (walletStoreProps: WalletStoreProps): void => {
 };
 
 export const createWalletStore = ({
-    wallets,
+    wallets: adapters,
     autoConnect = false,
     onError = (error: WalletError) => console.error(error),
     localStorageKey = 'walletName',
 }: WalletStoreProps): WalletStore => {
     const name: Ref<WalletName | null> = useLocalStorage<WalletName>(localStorageKey);
-    const wallet = ref<Wallet | null>(null);
     const adapter = ref<Adapter | null>(null);
     const publicKey = ref<PublicKey | null>(null);
-    const ready = ref<boolean>(false);
+    const ready = ref<WalletReadyState>(WalletReadyState.NotDetected);
     const connected = ref<boolean>(false);
     const connecting = ref<boolean>(false);
     const disconnecting = ref<boolean>(false);
 
     // Helper methods to set and reset the main state variables.
     const setState = (state: {
-        wallet: Wallet | null;
         adapter: Adapter | null;
         publicKey: PublicKey | null;
-        ready: boolean;
+        ready: WalletReadyState;
         connected: boolean;
     }) => {
-        wallet.value = state.wallet;
         adapter.value = state.adapter;
         ready.value = state.ready;
         publicKey.value = state.publicKey;
@@ -99,46 +95,33 @@ export const createWalletStore = ({
     };
     const resetState = () => {
         setState({
-            wallet: null,
             adapter: null,
-            ready: false,
+            ready: WalletReadyState.NotDetected,
             publicKey: null,
             connected: false,
         });
     };
 
-    // Create a wallet dictionary keyed by their name.
-    const walletsByName = computed(() => {
-        return wallets.reduce<Record<WalletName, Wallet>>((walletsByName, wallet) => {
-            walletsByName[wallet.name] = wallet;
-            return walletsByName;
+    // Create a dictionary of wallet adapters keyed by their name.
+    const adaptersByName = computed(() => {
+        return adapters.reduce<Record<WalletName, Adapter>>((adaptersByName, adapter) => {
+            adaptersByName[adapter.name] = adapter;
+            return adaptersByName;
         }, {});
     });
 
-    // Update the wallet and adapter based on the wallet provider.
+    // Update the wallet adapter based on the wallet provider.
     watch(
         name,
         (): void => {
-            const wallet = walletsByName.value?.[name.value as WalletName] ?? null;
-            const adapter = wallet && wallet.adapter;
+            const adapter = adaptersByName.value?.[name.value as WalletName] ?? null;
             if (adapter) {
                 setState({
-                    wallet,
-                    adapter,
-                    ready: false,
+                    adapter: adapter,
+                    ready: adapter.readyState,
                     publicKey: adapter.publicKey,
-                    connected: adapter.connected,
+                    connected: adapter.connected
                 });
-
-                // Asynchronously update the ready state
-                const waiting = name;
-                (async function () {
-                    const readyValue = await adapter.ready();
-                    // If the selected wallet hasn't changed while waiting, update the ready state
-                    if (name === waiting) {
-                        ready.value = readyValue;
-                    }
-                })();
             } else {
                 resetState();
             }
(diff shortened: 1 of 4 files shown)


lorisleiva:
I'm a bit confused as to why this PR renames all `adapter` variables to `wallet`. and removes the `adapter` variable. This doesn't seem to be on par with what the react package is doing.

I also think that in order for the `vue` library to work (as opposed to using `@vue/reactivity-core`, you need to update the TS configs to avoid type conflicts with react (See the TS configs of `vue-ui`).

Randy808:
They could all be renamed to `adapter` but I wanted to make minimal changes to wallet-adapter-vue-ui since components like the WalletProvider [are using 'wallet' when destructuring the store from 'useWallet'](https://github.com/solana-labs/wallet-adapter/blob/master/packages/ui/vue-ui/src/WalletMultiButton.vue#L19) (and I'd have to change all places where that's being done). I don't have both `wallet` and `adapter` because `wallet` held a type for an abstraction that is seemingly no longer needed ([see the removal of the wallet type in wallet-adapter-base](https://github.com/solana-labs/wallet-adapter/commit/37b566336af43ec059e1218ea2f1f0fe57042e24#diff-f290a8216f7b0fb2c49529c4042e32ba875aae0057f29964678ef3c955a7d273L10)). When it was used it had properties that were moved into the adapter (properties like `icon` that [were in the wallet implementation](https://github.com/solana-labs/wallet-adapter/blob/809ce18a2be4ef96c79bb3331386638efa6f17d9/packages/wallets/phantom/src/wallet.ts#L9) can now [be seen in the adapter](https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/phantom/src/adapter.ts#L56)). The React package is wrapping the adapter in something that adheres to a [react-specific wallet interface](https://github.com/solana-labs/wallet-adapter/blob/master/packages/core/react/src/useWallet.ts#L12) which I imagine may have just been to not have to change much in wallet-adapter-react-ui since the change [was confined to just wallet-adapter-react](https://github.com/solana-labs/wallet-adapter/commit/96cc368c3fe4484048e8389a2abe8fe132d3aae1).

Also by `@vue/reactivity-core`, do you mean `@vue/runtime-core`?  I saw that things [were imported from 'vue' originally](https://github.com/solana-labs/wallet-adapter/blame/f2c90f3985270ba1738b8d950d2ea74777e4acc9/packages/core/vue/src/useWallet.ts#L12) so I'll take a look at the tsconfig to make sure nothing's breaking (it's working for me now but I'll try and confirm). Also I'll see if I can create a quick docker image later (and upload it) to re-create the problem starting from a fresh install (to ensure that the problem is reproducible)

jordansexton:
> The React package is wrapping the adapter in something that adheres to a [react-specific wallet interface](https://github.com/solana-labs/wallet-adapter/blob/master/packages/core/react/src/useWallet.ts#L12) which I imagine may have just been to not have to change much in wallet-adapter-react-ui since the change [was confined to just wallet-adapter-react](https://github.com/solana-labs/wallet-adapter/commit/96cc368c3fe4484048e8389a2abe8fe132d3aae1).

Hey, good observation! There were some significant breaking changes to the base `WalletAdapter` interface, absorbing the `Wallet` interface which disappeared from the base package.

This makes the notion of a "wallet" in the React package a bit confusing, because you declare an array of Adapter objects and pass it to the `wallets` property of the provider. This was done to avoid significant breaking changes for consumers who use `<WalletsProvider wallets={[...]}>`, including the UI packages based on it.

In the React package, we should probably rip off the band-aid and pass in `adapters={[ ... ]}`, and should also make the `Wallet` interface that only has `{ adapter: Adapter, readyState: WalletAdapterReadyState }` less anemic, perhaps incorporating a `name` and `url` property or getter, so you don't need to do `wallet.adapter.name` etc.

I was in favor of just shipping the fixes that got us the wallet detection we wanted, but it could be time to think about how this small bit of technical debt messes with people's ability to understand the library.

cc @steveluscher wdyt?

And @Randy808 -- thanks for this PR! I'll try to give a proper review ASAP.

lorisleiva:
Thanks for the update guys. It makes more sense to me now.

I think it might be an idea to push as much logic and type declarations as we can in the base package because, by browsing through the different core and UI packages, I can see more and more discrepancies between them. I understand breaking changes make that difficult but maybe a new major version (or minor for 0.x packages) could fix that.

packages/core/vue/src/useWallet.ts
```diff
@@ -38,13 +37,13 @@ export interface WalletStore {
         options?: SendTransactionOptions
     ): Promise<TransactionSignature>;
 
-    signTransaction: Ref<SignerWalletAdapter['signTransaction'] | undefined>;
-    signAllTransactions: Ref<SignerWalletAdapter['signAllTransactions'] | undefined>;
-    signMessage: Ref<MessageSignerWalletAdapter['signMessage'] | undefined>;
+    signTransaction: Ref<SignerWalletAdapterProps['signTransaction'] | undefined>;
+    signAllTransactions: Ref<SignerWalletAdapterProps['signAllTransactions'] | undefined>;
+    signMessage: Ref<MessageSignerWalletAdapterProps['signMessage'] | undefined>;
```
jordansexton, line 42: Can we keep these lines unchanged? The `Props` should probably have been an internal API.

packages/core/vue/src/useWallet.ts
```diff
@@ -85,22 +83,19 @@ export const createWalletStore = ({
...
-        adapter.value = state.adapter;
         ready.value = state.ready;
         publicKey.value = state.publicKey;
         connected.value = state.connected;
     };
     const resetState = () => {
         setState({
             wallet: null,
-            adapter: null,
             ready: false,
```
jordansexton, line 99: This will need to change too. The ready state is no longer binary.

packages/core/vue/src/useWallet.ts
```diff
@@ -109,36 +104,29 @@ export const createWalletStore = ({
...
-            const adapter = wallet && wallet.adapter;
-            if (adapter) {
+            if (wallet) {
                 setState({
                     wallet,
-                    adapter,
-                    ready: false,
-                    publicKey: adapter.publicKey,
-                    connected: adapter.connected,
+                    ready: wallet.readyState == WalletReadyState.Installed,
```
jordansexton, line 121: Yeah, this isn't a valid assertion. The ready state of a `Loadable` wallet is still "ready", as in it's ready to be loaded on demand. But in general, as with the React package, we want to surface this through the state because the distinction matters for building more intuitive UIs, e.g. https://github.com/solana-labs/wallet-adapter/blob/16fb5cc395e9974a91297549feaf7694191c7093/packages/ui/react-ui/src/WalletModal.tsx#L23-L48

packages/core/vue/src/useWallet.ts
```diff
@@ -109,36 +104,29 @@ export const createWalletStore = ({
...
-                // Asynchronously update the ready state
-                const waiting = name;
-                (async function () {
-                    const readyValue = await adapter.ready();
-                    // If the selected wallet hasn't changed while waiting, update the ready state
-                    if (name === waiting) {
-                        ready.value = readyValue;
-                    }
-                })();
+                wallet.on("readyStateChange", () => {
```
jordansexton, line 127: Do these need to be unregistered if the adapters change?

packages/core/vue/src/useWallet.ts
```diff
@@ -149,34 +137,33 @@ export const createWalletStore = ({
...
-            _adapter.off('error', onError);
+            _wallet.off('connect', onConnect);
+            _wallet.off('disconnect', onDisconnect);
+            _wallet.off('error', onError);
         });
     });
 
     if (typeof window !== 'undefined') {
-        // Ensure the adapter listeners are invalidated before refreshing the page.
+        // Ensure the wallet listeners are invalidated before refreshing the page.
```
jordansexton, line 166: Hmm this looks like a find/replace was done on adapter/wallet which doesn't apply in all cases.

packages/core/vue/src/useWallet.ts
```diff
@@ -38,13 +37,13 @@ export interface WalletStore {
...
-    signAllTransactions: Ref<SignerWalletAdapter['signAllTransactions'] | undefined>;
-    signMessage: Ref<MessageSignerWalletAdapter['signMessage'] | undefined>;
+    signTransaction: Ref<SignerWalletAdapterProps['signTransaction'] | undefined>;
+    signAllTransactions: Ref<SignerWalletAdapterProps['signAllTransactions'] | undefined>;
+    signMessage: Ref<MessageSignerWalletAdapterProps['signMessage'] | undefined>;
 }
 
 export interface WalletStoreProps {
-    wallets: Wallet[];
+    wallets: Adapter[];
```
jordansexton, line 46: Probably what makes sense to do is expose this, or at least alias it internally, as `adapters` and instances are `adapter`. Since this is a breaking change anyway, let's nail down the distinction.

packages/core/vue/src/useWallet.ts
Randy808: Got it, I'll switch it to a WalletReadyState. One thing I was wondering was whether the default value for that should be WalletReadyState.NotDetected or WalletReadyState.Loadable. I can't really see when one is used over another from looking at the Phantom and WalletConnect implementations. The [Phantom implementation defaults to 'NotDetected'](https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/phantom/src/adapter.ts#L62-L65) when not supported while the [WalletConnect implementation defaults to 'Loadable'](https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/walletconnect/src/adapter.ts#L49-L50).

packages/core/vue/src/useWallet.ts
Randy808: Yeah they should be, I'll make the change

packages/core/vue/src/useWallet.ts
Randy808: True, I'll go back and replace back with adapters where necessary

packages/core/vue/src/useWallet.ts
jordansexton: Ah, the usual default is NotDetected. Loadable is only for a few adapters that load a third party library instead of attaching to a browser extension.

The distinction here is that for loadable wallets there nothing to detect (they can always be loaded).

packages/core/vue/src/useWallet.ts
Randy808: These have been changed back

packages/core/vue/src/useWallet.ts
Randy808: I tried to keep the interfaces the same but made some changes to the internal methods to keep the use of 'adapters' and 'adapter' consistent

packages/core/vue/src/useWallet.ts
Randy808: This has been changed to expose the raw type

packages/core/vue/src/useWallet.ts
Randy808: This has been moved to the 'watchEffect' call

packages/core/vue/src/useWallet.ts
Randy808: I tried to make the comments more sensible. Let me know if there's anything I should change in particular

Randy808:
The commit has been updated with some comment-guided changes. Let me know if there's something I misinterpreted and need to change

karnie6:
any update on this @jordansexton @Randy808 ?  Is there an ETA for this to be merged?

jordansexton:
Hey, I started reviewing this today but couldn't finish. I'll pick it up again soon and plan to merge this week. Sorry it's taking a while, just had my hands full with Solana Pay :)

karnie6:
All good @jordansexton - appreciate the response!

jordansexton:
Hey everyone, I apologize for having become a blocker to the advancement of Vue support in this repo.

Vue support has been moved to https://github.com/lorisleiva/solana-wallets-vue via https://github.com/solana-labs/wallet-adapter/pull/297.

Please reopen this work there, where I believe it can be much more quickly integrated.

lorisleiva:
I took this PR into consideration when I extracted the packages. It really helped, thanks! 🙂

PR #275, opened 2022-01-20
Revert "Polyfill Node APIs for compatibility with Webpack 5 and Create React App 5"

jordansexton:
Reverts solana-labs/wallet-adapter#264 which appears to break the example app.


jordansexton:
Screen Shot 2022-01-19 at 10 54 01 PM

PR #276, opened 2022-01-20
Polyfill Node APIs for compatibility with Webpack 5 and Create React App 5 (second try)

### LINKED ISSUE #241: Create React App is broken by lack of default polyfills in Webpack 5

steveluscher:
The new version of Webpack – and by extension the new version of Create React App – no longer includes polyfills for Node APIs by default. This means that if your browser-based app requires modules like `assert` and `stream`, Webpack will fail to find them and fail to build.

In this PR, we alias the modules that the wallet adapter depends upon to their related browser polyfills.

I've also followed up with each repository to let them know that they might like to eliminate or polyfill these modules in their own browser builds.

* https://github.com/ethereumjs/ethereumjs-monorepo/pull/1638/
* https://github.com/crypto-browserify/cipher-base/pull/16

If you're having trouble using the Solana wallet adapter with Create React App 5, can you patch this PR in and report back? I might have missed some polyfills.

Fixes #241.


steveluscher:
This is a revert and patch of https://github.com/solana-labs/wallet-adapter/pull/275.

packages/wallets/ledger/package.json
```diff
@@ -27,6 +27,7 @@
         "postbuild": "echo '{\"type\":\"commonjs\"}' | npx json > lib/cjs/package.json && echo '{\"type\":\"module\"} ' | npx json > lib/esm/package.json"
     },
     "dependencies": {
+        "buffer": "npm:buffer@^6.0.3",
```
steveluscher, line 30: Used `buffer` instead of the olde `buffer-browserify` which didn't have the full API surface needed by Ledger.

steveluscher:
Added test plan:

```
cd packages/starter/example
yarn dev
open http://localhost:3000
```

Next.js example app loaded this time.

TomLisankie:
Thanks so much for doing this @steveluscher. Is there a timeline for this getting merged in? Would be a great help since the building process errors out when trying to use the version up on NPM.

steveluscher:
No problem, @TomLisankie! I think the core team must have been crunching last weekend, supercharging the validators 🚀

Fingers crossed I hit all the cases with this PR, but I want someone more familiar with all the wallet use cases to give it a good test.

Speaking of testing, I've been thinking about how we could write automated tests for this thing. I'm pretty sure there's a world where we could image a test environment that has every wallet installed, and run Playwright end-to-end tests against that. Would make an awesome project for someone.

steveluscher:
I built the React starter app, then gripped the `dist/` folder for each of these removed Node polyfills: https://github.com/webpack/node-libs-browser

```
# eg. grepped for...
(['"]zlib['"]|new zlib\(|\bzlib\.)
```

The only one that I found was `URL`, which is supported natively by all browsers except IE 11 and earlier.

jordansexton:
The example works! https://solana-labs.github.io/wallet-adapter/example/

Not sure if CRA 5 does, but at least everything is building, and the packages have been published.

gaolingxiao:
Is this working with CRA 5?

jordansexton:
> Is this working with CRA 5? Mine seems not working though.

I think it's possible an issue was reintroduced by
https://github.com/solana-labs/wallet-adapter/pull/296

I'm looking into it. cc @armaniferrante

jordansexton:
@gaolingxiao @armaniferrante

This is fixed via #299 but you'll need to use react-app-rewired with config here to do the polyfills correctly. See the new [create-react-app-starter](https://github.com/solana-labs/wallet-adapter/tree/master/packages/starter/create-react-app-starter) project. The key parts are the [config-overrides.js](https://github.com/solana-labs/wallet-adapter/blob/master/packages/starter/create-react-app-starter/config-overrides.js) file and [`npm:process`](https://github.com/solana-labs/wallet-adapter/blob/f8a2f99a1a77f34cdff49c7ca0effc6fe8477f21/packages/starter/create-react-app-starter/package.json#L23) dependency.

The core issue is that the Torus and Ledger JS libraries don't correctly polyfill and incorrectly use globals, and this isn't handled by Webpack 5 without additional configuration, which you can't add with CRA5 alone. We use react-app-rewired to override, as Craco doesn't support CRA5 yet.

Published:
 - @solana/wallet-adapter-ledger@0.9.7
 - @solana/wallet-adapter-torus@0.11.9
 - @solana/wallet-adapter-wallets@0.15.3

PR #290, opened 2022-01-25
UI Button Text Refactor

derivativedegen:
We frequently alter the UI text of "Connect a different wallet" to be shorter "Change wallet"
This way it is similar to the other two button options in length.
Submitting this as simple a UI improvement.


jordansexton:
Thanks!

PR #300, opened 2022-02-07
added more dynamic attributes for button component

cad324:
Added the `id` and button `type` to the button component props. This allows developers to easily add these properties when using the component. For example, using `type="submit"` when the Button is used inside of a form.


jordansexton:
Hmm it would probably be more future proof to allow any props from the HTML button element type to be provided.

Also, why would you use `type="submit"` for the wallet buttons? These aren't supposed to be part of a form.

cad324:
Okay

PR #304, opened 2022-02-09
Add HuobiWallet adapter

enoughpower:
Add HuobiWallet adapter


jordansexton:
Hey @ryoqun, thanks for the ping. I'll review as soon as I can, just pretty slammed at the moment with payments and moving.

jordansexton:
@enoughpower Thanks for the PR! I've made some changes in the `huobi-wallet` branch here: https://github.com/solana-labs/wallet-adapter/tree/huobi-wallet/packages/wallets/huobi

I couldn't push them to this branch because of permissions set on your fork to disallow commits from maintainers.

Most of the changes are just structural, renaming the directory and package from `huobiwallet` to `huobi` which is the convention used for adapters here (the wallet name is still `HuobiWallet`, which is what dapp users would see).

However, one change is that `window.solana` is not accepted for new wallet adapters, and we are phasing use of it out for existing adapters as well. I've renamed this to `window.huobiWallet`, but it could also be `window.huobi` if you prefer.

Please merge that branch into here, and then this should be good to merge!

enoughpower:
[jordansexton](https://github.com/jordansexton) hello,thanks for your review!i had merge your commit now,please check it.

jordansexton:
Merged, will release Monday. Thanks!

PR #310, opened 2022-02-14
add signMessage method

xiaobaixin:


jordansexton:
For this to work, you'll need to extend like this: https://github.com/solana-labs/wallet-adapter/blob/fd543782d9fed22c564855308216dacfc75c0c32/packages/wallets/phantom/src/adapter.ts#L52

xiaobaixin:
> For this to work, you'll need to extend like this:
> 
> https://github.com/solana-labs/wallet-adapter/blob/fd543782d9fed22c564855308216dacfc75c0c32/packages/wallets/phantom/src/adapter.ts#L52

jordansexton:
It's okay, you don't need to close the PR! I just want you to make this change to it so that the API will work with the other packages.

xiaobaixin:
yeah  i'll create new pr now

PR #321, opened 2022-02-18
Detect if Solflare is installed

vsakos:
Improving detection of Solflare presence through all platforms (extension, mobile, web).

image

### DIFF
--- a/packages/wallets/solflare/src/adapter.ts
+++ b/packages/wallets/solflare/src/adapter.ts
@@ -1,5 +1,6 @@
 import {
     BaseMessageSignerWalletAdapter,
+    scopePollingDetectionStrategy,
     WalletAdapterNetwork,
     WalletConnectionError,
     WalletDisconnectedError,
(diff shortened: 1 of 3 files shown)


packages/wallets/solflare/src/adapter.ts
```diff
@@ -166,6 +170,29 @@ export class SolflareWalletAdapter extends BaseMessageSignerWalletAdapter {
         }
     }
 
+    private _detectWallet = async () => {
+        const Solflare = (await import('@solflare-wallet/sdk')).default;
```
jordansexton, line 174: Hmm, I can't accept this change. All adapters that import third party libraries do so when connecting the wallet, which requires a user action or an app to explicitly call `connect`.

This change is causing the library to be loaded immediately when the adapter's constructor is called. It shoudn't be necessary to load a library to detect the extension, so please refactor.

The code used in the other adapters should work here prettymuch out of the box.

packages/wallets/solflare/src/adapter.ts
```diff
@@ -59,7 +61,9 @@ export class SolflareWalletAdapter extends BaseMessageSignerWalletAdapter {
     async connect(): Promise<void> {
         try {
             if (this.connected || this.connecting) return;
-            if (this._readyState !== WalletReadyState.Loadable) throw new WalletNotReadyError();
+            if (this._readyState !== WalletReadyState.Loadable && this._readyState !== WalletReadyState.Installed) {
+                throw new WalletNotReadyError();
+            }
 
             if (!this._wallet) {
                 const Solflare = (await import('@solflare-wallet/sdk')).default;
```
jordansexton, line 69: This should also be changed -- if you can interact with the browser extension without loading a library, please do.

PR #325, opened 2022-02-22
Add support for Glow wallet

vpontis:
Currently Glow only works as a Safari extension on iOS but we are planning support for all browsers.


jordansexton:
Thank you! For now, please scope the changes for this PR to just adding the adapter. It's easier to review and add to docs / examples as a followup after we release it. Let me know when it's ready for review!

vpontis:
@jordansexton this is ready to review — we're launching the Safari Extension this week. Should I invite you to the TestFlight to check it out? DM me the email you use on your Apple account :). 

Also, how can I officially request your review on GitHub?

packages/wallets/glow/package.json
```diff
@@ -0,0 +1,33 @@
+{
+    "name": "@solana/wallet-adapter-glow",
+    "version": "0.15.4",
```
jordansexton, line 3: This can be `0.1.0` to start.

packages/wallets/glow/src/adapter.ts
```diff
@@ -0,0 +1,270 @@
...
+        transaction: Transaction,
+        options?: SendOptions
+    ): Promise<{ signature: TransactionSignature }>;
+    signMessage(message: Uint8Array): Promise<{ signature: Uint8Array }>;
+    connect(): Promise<void>;
+    disconnect(): Promise<void>;
+}
+
+interface GlowWindow extends Window {
+    solana?: GlowWallet;
```
jordansexton, line 47: We aren't accepting new adapters that use `window.solana`, and we've also requested that wallets that used this previously add a custom name. This prevents multiple installed wallets from fighting over declaring `window.solana`, which ultimately prevents good detection. Would you be willing to change this to `window.glowWallet` or something similar?

packages/wallets/glow/src/adapter.ts
```diff
@@ -0,0 +1,270 @@
...
...EoOLuqTmoXLZSxEyolHx96GAF3d3XLkYOtsv3Fl8RsnfiHhKpxMmFUqfR1nJR0Tx9OmPnaV6mMCLeFBUIArLOE6oWJBL52+c425/rLljf19w6809PRHfDTLxAb9UJH3UU5EJLgYBvbELDJPIjhux4iAjcOv14QR6yYlQGcH7r37pZu9CIwaCPegIhw1y+qmyRxvC7ix6JkdwdC8wHD2mjfIWiTbBndojAgKB1qjMvPb0FJ2RgXnuti0ifXXXLtlnQytbivpzuNk1dMdTIColLRSQjGxjYGSMivR0omDZwnNYDCDy6OIsUehBHQSbbxk14Gn+oymRQM3nyjZLueV3iNgLDlbIrqZCZp0AcEoU4iPWrk6FheUfGWn2x+r5mMwreIxJ1wsH/IDqY8vMzG2rlSiTgxJBVzFlCJkEWaCBEJFwOXgSqIRawWhngNC8B6+AHYGIYEUEt6Fh5C4cfQh2BRj8azFdU/S3Y1MDrAiw4vH84lLxG7ix29FfNWxP5t64/a5lXPyRbnF16KUDsNzC68Ar54wTsgqwdRL2I9h3GLVD3rSdTDgDjjaYzLxBDz3JbsD2psNn1NH7pYM595zntqhmoHSs6l954bxmLxTFlpCaw8du9/vPnu+pbmZnplkRbBP6hIS7Dic19dh//+WIkzMlQ9kwWffSymjocdja2A5+QweEg5OFAtw3GWlACyqgzrkAp/qnLmWRZSMW1GHdmX9qIqBpFENaxmZYInRhR4ueNmAMLDS2s36/iPPf2bnatagBsZL7DMECqJjly19KZ73CBcwxPdQDKJNfAPIZgKz8EXvDCiiikQumQqBDI3F+aIYNoNiVW1Yp2EKihbj8CwD3tyI1Uo5mpAKWX4P4MwgKRCBI0e/hHB9fE9xoEk/v2NnWvYA4kYIqysoeG70uJQMv/yxa814iPIWpyzlvAogw/wMFYcWWGkKgm1F3Pwt0BUKpQOMwARjNYViGmDyYPgs2AUBAGQATR2+2Mda1CleZjiPxSEiDBCx3vL9+N3Pv76e++2gLH3GQAqAJUIVh1KtJkntj+RZsO/Xn7rQhC+zAlClGE1pFFuGBQZvEqGRFMqRjL4fygFYIBQEuaeAAzBRhpnVcpIRb0WRlJ6TJQEbk+ErtMKY30HZvvSY7/csY3PVuAfgZ7APwKxbtN/AyEn1KgMAwQMAAAAAElFTkSuQmCC';
```
jordansexton, line 60: This isn't a requirement, but if you have an SVG instead of a PNG, the encoded size will be much smaller.

packages/wallets/glow/src/adapter.ts
```diff
@@ -0,0 +1,270 @@
...
+                if (typeof event.data === 'object' && event.data.__glow_loaded) {
+                    if (this._readyState !== WalletReadyState.Installed) {
+                        this._readyState = WalletReadyState.Installed;
+                        this.emit('readyStateChange', this._readyState);
+                    }
+                    window.removeEventListener('message', handler);
+                }
+            };
+
+            window.addEventListener('message', handler);
```
jordansexton, line 86: Hmm. I like the approach, but I don't like that this will leave a dangling event listener. The adapters can be reinstantiated (this can easily happen when used with React), and this will lead to a memory leak.

However, I can see how the interval used by `scopePollingDetectionStrategy` could cause similar issues. 

https://github.com/solana-labs/wallet-adapter/issues/334 and https://github.com/solana-labs/wallet-adapter/issues/337 may be of some use.

Perhaps what we need is a teardown function for adapters, but maybe I'm not being imaginative enough. @steveluscher wdyt?

packages/wallets/glow/src/adapter.ts
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    async connect(): Promise<void> {
+        try {
+            if (this.connected || this.connecting) {
+                return;
+            }
+            if (this._readyState !== WalletReadyState.Installed) {
+                throw new WalletNotReadyError();
+            }
```
jordansexton, line 124: style nit: project code style is to not use braces on early return or throw.

I wonder if there's a lint fix for this...

packages/wallets/glow/src/adapter.ts
```diff
@@ -0,0 +1,270 @@
...
+            if (wallet && 'signAndSendTransaction' in wallet && !options?.signers) {
+                // TODO: update glow to fix this
+                // HACK: Glow's `signAndSendTransaction` should always set these, but doesn't yet
+                transaction.feePayer = transaction.feePayer || this.publicKey || undefined;
+                transaction.recentBlockhash =
+                    transaction.recentBlockhash || (await connection.getRecentBlockhash('finalized')).blockhash;
+
+                const { signature } = await wallet.signAndSendTransaction(transaction, options);
+                return signature;
+            }
```
jordansexton, line 200: This is undesired behavior and looks like it was copied from Phantom. It would be ideal to fix rather than duplicate.

packages/wallets/wallets/package.json
```diff
@@ -36,6 +36,7 @@
         "@solana/wallet-adapter-coinhub": "^0.3.2",
         "@solana/wallet-adapter-ledger": "^0.9.8",
         "@solana/wallet-adapter-mathwallet": "^0.9.2",
+        "@solana/wallet-adapter-glow": "^0.15.4",
```
jordansexton, line 39: Version fix here too.

packages/wallets/glow/src/adapter.ts
vpontis: I would recommend just using [Prettier](https://prettier.io/) and then not worrying about any style stuff. That's what we do for the most part and it tends to work.

packages/wallets/glow/src/adapter.ts
vpontis: Yea, I'm a little confused by what's going on here haha. I don't think we need any of this code but I'm a little worried that some client that's set up to use Phantom will have a harder time integrating Glow if we leave this out.

packages/wallets/glow/src/adapter.ts
vpontis: Updated to `glowSolana` 👍

packages/wallets/glow/src/adapter.ts
vpontis: We don't have one now but I'll update it once we do 👍

packages/wallets/glow/src/adapter.ts
vpontis: Since I want to get this in soon and iterate on it, I'm happy to do whatever works for other wallets now. 

Should we just remove the event listener part?

jordansexton:
For sure, I can get this in quickly. Would you grant me the ability to push to this branch? I'll test, make any quick changes if needed, and publish tomorrow.

packages/wallets/glow/src/adapter.ts
```diff
@@ -0,0 +1,266 @@
...
...EoOLuqTmoXLZSxEyolHx96GAF3d3XLkYOtsv3Fl8RsnfiHhKpxMmFUqfR1nJR0Tx9OmPnaV6mMCLeFBUIArLOE6oWJBL52+c425/rLljf19w6809PRHfDTLxAb9UJH3UU5EJLgYBvbELDJPIjhux4iAjcOv14QR6yYlQGcH7r37pZu9CIwaCPegIhw1y+qmyRxvC7ix6JkdwdC8wHD2mjfIWiTbBndojAgKB1qjMvPb0FJ2RgXnuti0ifXXXLtlnQytbivpzuNk1dMdTIColLRSQjGxjYGSMivR0omDZwnNYDCDy6OIsUehBHQSbbxk14Gn+oymRQM3nyjZLueV3iNgLDlbIrqZCZp0AcEoU4iPWrk6FheUfGWn2x+r5mMwreIxJ1wsH/IDqY8vMzG2rlSiTgxJBVzFlCJkEWaCBEJFwOXgSqIRawWhngNC8B6+AHYGIYEUEt6Fh5C4cfQh2BRj8azFdU/S3Y1MDrAiw4vH84lLxG7ix29FfNWxP5t64/a5lXPyRbnF16KUDsNzC68Ar54wTsgqwdRL2I9h3GLVD3rSdTDgDjjaYzLxBDz3JbsD2psNn1NH7pYM595zntqhmoHSs6l954bxmLxTFlpCaw8du9/vPnu+pbmZnplkRbBP6hIS7Dic19dh//+WIkzMlQ9kwWffSymjocdja2A5+QweEg5OFAtw3GWlACyqgzrkAp/qnLmWRZSMW1GHdmX9qIqBpFENaxmZYInRhR4ueNmAMLDS2s36/iPPf2bnatagBsZL7DMECqJjly19KZ73CBcwxPdQDKJNfAPIZgKz8EXvDCiiikQumQqBDI3F+aIYNoNiVW1Yp2EKihbj8CwD3tyI1Uo5mpAKWX4P4MwgKRCBI0e/hHB9fE9xoEk/v2NnWvYA4kYIqysoeG70uJQMv/yxa814iPIWpyzlvAogw/wMFYcWWGkKgm1F3Pwt0BUKpQOMwARjNYViGmDyYPgs2AUBAGQATR2+2Mda1CleZjiPxSEiDBCx3vL9+N3Pv76e++2gLH3GQAqAJUIVh1KtJkntj+RZsO/Xn7rQhC+zAlClGE1pFFuGBQZvEqGRFMqRjL4fygFYIBQEuaeAAzBRhpnVcpIRb0WRlJ6TJQEbk+ErtMKY30HZvvSY7/csY3PVuAfgZ7APwKxbtN/AyEn1KgMAwQMAAAAAElFTkSuQmCC';
```
steveluscher, line 60: Thought: This isn't about you – you're just replicating what came before – but we might consider out-lining these inlined icons at some point.

Pro: inlining them is simple and they load instantly.
Con: every visitor pays the cost of the images, on every visit, whether or not they ever open the wallet selection dialog. This icon here, for instance, is 5.7K gzipped. Multiply that by the number of wallets a given app supports and you have a whole hamburger.

steveluscher: One thing you might consider here – if possible – is to drum up an SVG version of this icon. Since it's so geometric, it might very well be tiny.

packages/wallets/glow/src/adapter.ts
steveluscher: This will definitely leak memory as Jordan said, because the event handler holds a reference to `this`. No matter if the dApp wants to dispose of an old reference to `new GlowWalletAdapter`, it will stick around forever.

Like you said, Jordan, we'd have to enforce through `Base*WalletAdapter` that subclasses implement a destructor. While this would work, it will _definitely_ turn us all into janitors for years to come – we'll be finding places where folks forgot to call `destroy()`, or release a subscription, forever.

We'll probably have to end up doing something like this, though. I'm guessing we could minimize the damage by creating a ‘detection starter’ instance method on the base adapter that returns a dispose handle, then make all of our code call that and the dispose handle it returns when the time comes.

```
type DisposeHandle = () => void;
function startDetection(): DisposeHandle {
  function handler() { /* ... */ }
  window.addEventListener('...', handler);
  const diposePoller = scopePollingDetectionStrategy(/* ... */);
  let disposed = false;
  return () => {
    if (disposed) return;
    disposed = true;
    window.removeEventListener('...', handler);
    disposePoller();
  }
}
```

Anyway, we'll figure that out later @vpontis. I'd say, for now, just make sure to `removeEventListener` the `handler` where you `return true` from the `scopePollingDetectionStrategy`. Then, at least, the handler will die when Glow is detected.

steveluscher (changes requested):

vpontis:
> For sure, I can get this in quickly. Would you grant me the ability to push to this branch? I'll test, make any quick changes if needed, and publish tomorrow.

@jordansexton sent you an invite to our repo, not sure how else to do this. The GitHub docs [show a checkbox](https://docs.github.com/assets/cb-44583/images/help/pull_requests/allow-maintainers-to-make-edits-sidebar-checkbox.png) that I can't find.

packages/wallets/glow/src/adapter.ts
vpontis: Updated, thanks Steve!

packages/wallets/glow/src/adapter.ts
vpontis: Yea let me ping our designers and see if they can get me an svg version.

steveluscher (approved):
LGTM, but I don't have merge rights.

PR #326, opened 2022-02-23
Update slope adapter

SlopeFinance:


jordansexton:
This PR looks like it might be incomplete. `window.slopeApp` is detected but not used in the `connect` function.

SlopeFinance:
> This PR looks like it might be incomplete. `window.slopeApp` is detected but not used in the `connect` function.

`window.slopeApp` is a `JavascriptChannel` defined in the Mobile App. The PR is to allow `readyState` to be successfully marked as `Installed`, and the same way done in the Solflare adapter. Thanks for checking.

jordansexton:
The issue I'm seeing is with this line: https://github.com/solana-labs/wallet-adapter/blob/bb78389dd599c151f2cb959b827f4dad3855e75e/packages/wallets/slope/src/adapter.ts#L107

If `connect` is called, and `window.Slope` is not defined, but `window.slopeApp` is, we will reach this line, and it will error upon trying to instantiate `window.Slope` which we have forced to be non-null.

Can you provide more context on how this is supposed to work?

jordansexton:
Please review: https://github.com/solana-labs/wallet-adapter/pull/336

This PR incorporates this work and also addresses the concern I have.

Issue #215, opened 2021-12-21
[rfc] Work with wallet provider ecosystem to improve reliability and performance of wallet detection


steveluscher:
Sooner or later, the ability to determine who the user is and what they own finds itself on the critical path for any dApp to be able to deliver its core experience. For that reason, the **performance** and **reliability** of wallet detection and autoconnection is _super important_. In this RFC I would like to propose a different wallet readiness API than the one we have now, with the goal of being able to make wallet-bound decisions in dApps with 100% reliability and a minimum of delay.

# Background

Today, most wallets announce themselves by adding variables to the global scope. For instance, when the Phantom extension boots, it adds the object `window.phantom.solana` to the global scope.

# The situation today

The addition of an object to the global scope is not something that you can subscribe to. If dApps want to know whether Phantom is installed, they have to _poll the global object_ looking for it.

Today, we do something like this in the Solana wallet adapter.

```
async function phantomIsInstalled(): Promise<boolean> {
  // Is this even a browser?
  if (typeof window === 'undefined' || typeof document === 'undefined') return false;

  function check() {
    return !!window.phantom?.solana;
  }

  // Is the page already loaded? Check now.
  if (document.readyState === 'complete') return check();

  // Wait until the page is loaded, then check.
  return new Promise(resolve => {
    function listener() {
      window.removeEventListener('load', listener);
      resolve(check());
    }
    window.addEventListener('load', listener);
  });
}
```

# Problem

## Reliability vulnerabilities

* If a wallet doesn't install its hooks by the time the window's `load` event has fired, we will miss it. Despite this being extremely difficult to reproduce, I have observed this happen in the wild, with Phantom.

## Performance liabilities

First, take a look at this diagram by Kudzanayi ([link](https://medium.com/@kudzanayi/js-async-defer-ac596631ec61)):

![](https://miro.medium.com/max/1186/1*7piSgVkkxWx1w0cfVjiung.png)

* The `load` event of a window fires _after_ all image resources have loaded. In practice, this can delay our wallet readiness check by _seconds_, depending on what's on the page and how fast your network is; just imagine that a site has a big 25MB banner image, and you're on a slow connection. Simulate this by [throttling your connection using the Chrome dev tools](https://developer.chrome.com/docs/devtools/evaluate-performance/reference/#network-throttle) then running [this playground](https://jsfiddle.net/tb8gy7fc/1/).

[...]

PR #327, opened 2022-02-23
Repeat wallet detection polling periodically

sjchmiela:
Having to refresh the whole page after installing/enabling an extension on a website is a subpar user experience.

After this pull request is merged, wallets will get detected "live", as they become available for the website.

I believe there are two situations when this can happen:
- extension is installed (onboarding flow, I'd say crucial to the network adoption)
- extension is enabled (Safari iOS extension flow, described eg. at [this article](https://www.imore.com/how-add-and-enable-extensions-safari-iphone-and-ipad#enable)).

`performDetection` callbacks are very very light functions, calling them on a few adapters shouldn't impact overall browser performance.

### DIFF
--- a/packages/core/base/src/adapter.ts
+++ b/packages/core/base/src/adapter.ts
@@ -90,14 +90,31 @@ export abstract class BaseWalletAdapter extends EventEmitter<WalletAdapterEvents
 }
 
 export function scopePollingDetectionStrategy(performDetection: () => boolean): void {
+    // Early return when server-side rendering
     if (typeof window === 'undefined' || typeof document === 'undefined') return;
+
+    const onDocumentLoaded = () => {
+        if (performDetection()) {
+            // Wallet detected, nothing more to do
+            return;
+        }
+        // Wallet not detected yet, try detecting every second
+        let repeatingTimer = setInterval(() => {
+            if (performDetection()) {
+                // Wallet detected, nothing more to do
+                clearInterval(repeatingTimer);
+            }
+        }, 1000);
+    };
+
     if (document.readyState === 'complete') {
-        performDetection();
+        onDocumentLoaded();
         return;
     }
+
     function listener() {
         window.removeEventListener('load', listener);
-        performDetection();
+        onDocumentLoaded();
     }
     window.addEventListener('load', listener);
 }


vpontis (commented):
This is awesome thank you @sjchmiela! 

@jordansexton, Stanislaw is on the Glow team and noticed this while building out our Safari Extension.

jordansexton:
Thanks for this! I think we get most of the benefit of https://github.com/solana-labs/wallet-adapter/issues/215 and performance is probably okay. I wonder if it would be beneficial to use a single interval and then loop through wallets registered for detection...

packages/core/base/src/adapter.ts
```diff
@@ -90,14 +90,31 @@ export abstract class BaseWalletAdapter extends EventEmitter<WalletAdapterEvents
...
+            // Wallet detected, nothing more to do
+            return;
+        }
+        // Wallet not detected yet, try detecting every second
+        let repeatingTimer = setInterval(() => {
+            if (performDetection()) {
+                // Wallet detected, nothing more to do
+                clearInterval(repeatingTimer);
+            }
+        }, 1000);
```
steveluscher, line 107: If the wallet isn't detected, we're in one of two states, right?

1. It hasn't been introduced into the `window` scope yet.
2. It's not installed in the browser, at all.

Unless I'm reading this incorrectly, this will keep polling forever, if we hit case 2, right?

steveluscher: Reading back over the description in your PR, I can see that this was your intent :)

steveluscher: In general, we should try to avoid scheduling work forever, so the browser process can sleep. There are a couple of things we could add to this to make it lighter on the system:

1. Build a task queue like @jordansexton suggested: https://developer.mozilla.org/en-US/docs/Web/API/Background_Tasks_API
2. Use `requestIdleCallback` if available to schedule work _only_ when the browser is idle.

```
// https://jsbin.com/dadehafeqe/1/edit?js,console
function detect() {
  console.log('do detection work');
}
function schedule() {
  setTimeout(() => {
    requestIdleCallback(() => {
      if (detect()) {
        return;
      }
      schedule();
    });
  }, 1000);
}
if (!detect()) {
  schedule();   
}
```

sjchmiela: Yea, I agree this definitely isn't the best way to handle this whole situation, but it is _a_ way! I'd much rather see wallets/extensions handle it like we will in Glow, i.e.:
- post a message to `window` once the extension is ready
- catch it in the adapter [like so](https://github.com/solana-labs/wallet-adapter/pull/325/commits/b89d518f11ae7406999888a0208edb531caef032)

This way there's no need to do anything periodically!

Maybe we should:
- add an argument to `scopePollingDetectionStrategy` which would disable the forever-repeating interval
- encourage wallet adapters to implement event-based live detection and disable the interval?

steveluscher:
> post a message to `window` once the extension is ready

Would that still leave us open to a race condition, if the extension posts the message _before_ the app installs a listener for it? I want to nudge us into a place where race conditions become _impossible_, and wallet detection is bulletproof. I think that the architecture in #215 would achieve that! What do you think?

> add an argument to `scopePollingDetectionStrategy` which would disable the forever-repeating interval

If we make it more cooperative (ie. schedule the work with `requestIdleCallback` we'll probably be OK.

steveluscher:
I'm also thinking through an architecture that might obviate the need to poll, ever. Your use case was ‘I just installed the extension, and I want it to be recognized _mid session_.’ I'm wondering if, when you install the extension, the window experiences a `blur` event, and once it's installed it experiences a `focus` event. Maybe all we need to do is to perform a round of detection every time the window focuses!

steveluscher:
@sjchmiela, can you rock this up:

```
window.addEventListener('blur', () => {
  console.log('blur');
  // Log whether or not your wallet is installed.
});
window.addEventListener('focus', () => {
  console.log('f');
  // Log whether or not your wallet is installed.
});
```

Then go through the process of installing Glow from scratch? If you can detect it on `focus`, then we can get rid of most of this polling stuff.

PR #338, opened 2022-02-27
connecting synced to autoConnect

Legendree:
Initializing `connecting` to `false` at all times introduces a flickering for components if you rely on `wallet.connecting` e.g. trying to figure either show the spinner first or "unauthorized" message. Should be synced with `autoConnect` property.


Legendree:
@jordansexton

PR #289, opened 2022-01-25
Warn on the console when the WalletModalProvider is missing

steveluscher:
@21e8 got into a bit of trouble because their app was missing a `WalletModalProvider`.

In this PR, we:

* Make it so that a missing wallet modal provider isn't a runtime error
* A helpful warning appears on the console asking you to add a `<WalletModalProvider>` as an ancestor.

Fixes #286.

Test plan:

1. Remove the `<WalletModalProvider>` from the Next.js example
2. `yarn dev`
3. Click the button

Observe warning on console.

image

PR #342, opened 2022-03-02
Console error when provider is missing

tyirenkyi:
This PR builds on top of solution suggested by #289.


tyirenkyi:
@steveluscher could you check this out?

packages/ui/react-ui/src/useWalletModal.tsx
```diff
@@ -1,11 +1,14 @@
 import { createContext, useContext } from 'react';
+import { defaultValueForModalProvider } from '@solana/wallet-adapter-react';
```
steveluscher, line 2: This sets up a dependency between the `ant-design` package and the `react` package. This will be surprising to people, bloat their `node_modules` folders, and is sure to result in future pain. Let's not do this. See my thoughts on being DRY below.

steveluscher (changes requested):
Thanks for banging this out so fast! I made some simple comments, but I have a more general one, here:

I know that @jordansexton suggested that we DRY (don't repeat yourself) up the implementation of these default contexts so that we can reuse them. Now that you've shown us what that looks like, I notice a few things:

* We've created a cross-package dependency between `react` and `ant-design` so they can share the helper, which we can't afford to do.
* It's _really_ hard to understand what the helpers do without reading their implementation (ie. a newcomer to the code is unlikely to know what `missingProviderErrorMessage('foo', 'Bar', 'Baz', 'write')` actually does.

My suggestion here is to DRY up the implementation, but only _locally_ within the module. Instead of making an external `helpers.ts` file, just create a module local function in _each_ context, so that you don't have to literally type the full text of the error message over and over, and only if that is likely to save someone time in the future (eg. when they need to update the message).

DRY is cool, but sometimes inlining is the more pragmatic and straightforward thing to do. Also, I wouldn't be afraid of simply copy-pasting the message over and over; gzip will crush those repeated elements in a way that's competitive with the DRY version.

packages/ui/ant-design/src/useWalletModal.ts
```diff
@@ -1,11 +1,14 @@
 import { createContext, useContext } from 'react';
+import { defaultValueForModalProvider } from '@solana/wallet-adapter-react';
```
steveluscher, line 2: This sets up a dependency between the `ant-design` package and the `react` package. This will be surprising to people, bloat their `node_modules` folders, and is sure to result in future pain. Let's not do this. See my thoughts on being DRY below.

packages/core/react/src/useWallet.ts
```diff
@@ -37,7 +38,70 @@ export interface WalletContextState {
...
+    signMessage(_message: Uint8Array) {
+        return Promise.reject(
+            console.error(missingProviderErrorMessage("signMessage", "WalletContext", "WalletProvider", "call"))
+        )
+    }
+};
+Object.defineProperty(DEFAULT_CONTEXT, 'wallets', {
+    get() {
+        console.error(missingProviderErrorMessage("wallets", "WalletContext", "WalletProvider", "read"));
+        return false;
```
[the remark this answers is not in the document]
tyirenkyi: Thank you

tyirenkyi:
> Thanks for banging this out so fast! I made some simple comments, but I have a more general one, here:
> 
> I know that @jordansexton suggested that we DRY (don't repeat yourself) up the implementation of these default contexts so that we can reuse them. Now that you've shown us what that looks like, I notice a few things:
> 
> * We've created a cross-package dependency between `react` and `ant-design` so they can share the helper, which we can't afford to do.
> * It's _really_ hard to understand what the helpers do without reading their implementation (ie. a newcomer to the code is unlikely to know what `missingProviderErrorMessage('foo', 'Bar', 'Baz', 'write')` actually does).
> 
> My suggestion here is to DRY up the implementation, but only _locally_ within the module. Instead of making an external `helpers.ts` file, just create a module local function in _each_ context, so that you don't have to literally type the full text of the error message over and over, and only if that is likely to save someone time in the future (eg. when they need to update the message).
> 
> DRY is cool, but sometimes inlining is the more pragmatic and straightforward thing to do. Also, I wouldn't be afraid of simply copy-pasting the message over and over; gzip will crush those repeated elements in a way that's competitive with the DRY version.

Thanks so much for the quick feedback @steveluscher! This helped me think more about what to consider when designing solutions across packages. I followed your suggestions and made some changes.

packages/core/react/src/useWallet.ts
```diff
@@ -37,7 +37,69 @@ export interface WalletContextState {
...
 }
 
-export const WalletContext = createContext<WalletContextState>({} as WalletContextState);
+const EMPTY_ARRAY: ReadonlyArray<never> = [];
+
+const DEFAULT_CONTEXT = {
+    autoConnect: false,
+    wallets: [],
+    wallet: null,
+    publicKey: null,
```
steveluscher, line 46: You can remove these three, because you have defined getters below, right?

steveluscher (approved):
LGTM! Ship it.

packages/core/react/src/useWallet.ts
tyirenkyi: True, but this insufficient overlap error kept showing up. I just fixed it by type casting `DEFAULT_CONTEXT` to `WalletContextState`.
Screenshot 2022-03-04 at 2 48 43 PM

steveluscher: Yeah, makes sense. This might come back to bite us if we update the type of the context but _forget_ to update the implementation of the default context, but it's probably not the worst thing. If you're falling back to this default context you're already in a bad place.

jordansexton:
Nice work @tyirenkyi! This should really reduce the number of difficulties new devs run into with our React contexts.

And many for the reviews @steveluscher!

UsmanDesigner:
You have tried to  read "visible" on a WalletModalContext without providing one. Make sure to render a WalletModalProvider as an ancestor of the component that uses WalletModalContext
this issue came to me in useWalletModal I want to open a modal but it not opens what I do

PR #344, opened 2022-03-03
Commit update Coin98 Wallet Adapter

khiemsoft:


jordansexton (changes requested):
Sorry, I can't accept this PR.

It's difficult to review because there are changes to a bunch of the files here that don't seem substantive. Can you revert and just change the lines that need it?

Also, it looks like a bunch of important things were removed, including the use of `readyState` for detecting the wallet. I think you may need to rebase your fork from the latest code in master.

Issue #317, opened 2022-02-16
use DOMContentLoaded instead of load for scopePollingDetectionStrategy


ArpitAgrawal321:
Currently the detection mechanism waits for the `load` event to check for the wallets. This can take a lot of time incase a resource such as an image required by the page keeps loading for a while. This prevents the user from getting the correct UI, does not portray correct state of installed/loadable wallets.
We can use `DOMContentLoaded` instead of `load`  to `performDetection` once the DOM structure is loaded.

Current implementation:
```
function scopePollingDetectionStrategy(performDetection: () => boolean): void {
    if (typeof window === 'undefined' || typeof document === 'undefined') return;
    if (document.readyState === 'complete') {
        performDetection();
        return;
    }
    function listener() {
        window.removeEventListener('load', listener);
        performDetection();
    }
    window.addEventListener('load', listener);
}
```

Suggested Change:
```
function scopePollingDetectionStrategy(performDetection: () => boolean): void {
    if (typeof window === 'undefined' || typeof document === 'undefined') return;
    if (document.readyState === 'complete') {
        performDetection();
        return;
    }
    function listener() {
        window.removeEventListener('DOMContentLoaded', listener);
        performDetection();
    }
    window.addEventListener('DOMContentLoaded', listener);
}
```

ArpitAgrawal321:
Found #215 already exists, closing.

jordansexton:
#215 is just a proposal -- most of the work is on wallets to implement. If that's the only reason to close this, I'd suggest reopening and let's see if it makes sense to PR.

jordansexton:
Related: https://github.com/solana-labs/wallet-adapter/pull/327

@sjchmiela + @vpontis wdyt?

steveluscher:
Great suggestion, @ArpitAgrawal321! I'm trying to load this all back into my head to figure out why I didn't do this in the first place.

I _think_ maybe it's because the check for `document.readyState` was equivalent to listening to `DOMContentLoaded` in most cases, because usually the wallet code _by definition_ doesn't execute until after `DOMContentLoaded` has fired. This might not be true for all apps, so this could very well be worth adding.

I dug around and found some of my early exploration into this with Phantom: https://jsfiddle.net/tb8gy7fc/1/

In any case, we should definitely:

1. Do what you've suggested.
[...]

PR #346, opened 2022-03-04
Add listener to DOMContentLoaded when document readyState is loading

safaiyeh:
## Problem
The document load event waits for the browser to load all resources before executing [(reference)](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event). The Magic Eden web app is very resource heavily with all our images, videos, and gifs which caused the wallet connection to be delayed. The quick patch we used was to listen to the [DOMContentLoaded event](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event) so wallets can be connected before all the resources loaded.

## Solution 
@ArpitAgrawal321 suggested in https://github.com/solana-labs/wallet-adapter/issues/317 to fully use DOMContentLoaded for this case. @steveluscher [suggested](https://github.com/solana-labs/wallet-adapter/issues/317#issuecomment-1050538881) to also handle the in-between states of the document loading and finished loading.

This PR adds a `DOMContentLoaded` listener when the document state is loading while still watching the window `load` event in the case that the document is in the `complete` state

### DIFF
--- a/packages/core/base/src/adapter.ts
+++ b/packages/core/base/src/adapter.ts
@@ -106,14 +106,24 @@ export function scopePollingDetectionStrategy(detect: () => boolean): void {
         }, 1000);
     };
 
+    function windowDOMContentLoadedListener() {
+        window.removeEventListener('DOMContentLoaded', windowDOMContentLoadedListener);
+        poll();
+    }
+
+    if (document.readyState === 'loading') {
+        window.addEventListener('DOMContentLoaded', windowDOMContentLoadedListener);
+        return;
+    }
+
     if (document.readyState === 'complete') {
         poll();
         return;
     }
 
-    function listener() {
-        window.removeEventListener('load', listener);
+    function windowLoadListener() {
+        window.removeEventListener('load', windowLoadListener);
         poll();
     }
-    window.addEventListener('load', listener);
+    window.addEventListener('load', windowLoadListener);
 }


safaiyeh:
> > while still watching the window `load` event in the case that the document is in the `complete` state
> 
> Looks good, but I'm wondering about this part. It looks like it early returns if the state is `complete`?

Ah right, misspoke there. If state is complete the load event wouldn’t happen again, so it wouldn’t need the event listener.

jordansexton:
Makes sense, thanks! I'll patch release in the morning.

steveluscher (changes requested):
Thanks for drumming this up! Here's how I understand this implementation:

* IF the document is `ready`, detect.
  * IF detection failed, poll indefinitely until wallet is found
* ELSE IF document is loading, listen for document to become `ready`
  * WHEN `ready`, detect
  * IF detection failed, poll indefinitely until wallet is found
* ELSE listen for `load` event
  * WHEN `load` event is detected, detect
  * IF detection failed, poll indefinitely until wallet is found

The interesting thing to not about this is that the decision of when to start polling depends largely on when this code is run.

```
// DOMContentLoaded has not fired by the time you call this code?

                       polling starts here
                       V
loading --------> interactive --------> complete --------> load fires
            (aka DOMContentLoaded)

// DOMContentLoaded already fired by the time you call this code?

                                                           polling starts here
                                                           V
loading --------> interactive --------> complete --------> load fires
            (aka DOMContentLoaded)
```

I think the fix here is to introduce a check for `readyState === 'interactive'`. The `DOMContentLoaded` event and the `interactive` ready state are synonymous. Here, check it out: https://jsbin.com/yoniharigo/1/edit?html,js,console,output

tl;dr:

* `'loading'` implies that neither `DOMContentLoaded` nor the `load` event have fired.
* `'interactive'` implies that `DOMContentLoaded` has fired but the `load` event has not.
* `'complete'` implies that `DOMContentLoaded` and the `load` event have fired.

Also, if we start polling at `DOMContentLoaded`, this code does _not_ set up the `load` event listener. This means that we have, at most, 1000ms precision on detecting the wallet. If there's a wallet that typically becomes available on `load`, we could be late detecting it by up to 1000ms, which is huge in terms of the startup of a website.

Alright, that was a lot of talk. What's my feedback? Implement the missing functionality:

* [X] If `readyState` is loading, set up the `DOMContentLoaded` listener…
  * [x] …_and_ set up the `load` event listener. 
* [ ] If `readyState` is `interactive`, fire the `poll()` method.
  * [X] If detection failed, set up the interval…
  * [ ] …_and_ set up the `load` event listener.
* [X] If `readyState` is `complete`, fire the `poll()` method.
  * [X] If detection failed, set up the interval.
* [ ] If at any point detection succeeds, remove all of the listeners and cancel all of the intervals.

I'll probably add a bunch of stuff to this later, like:

* Exponential backoff of the poll interval (eg. poll less and less often as we get further from startup).
* The use of `requestIdleCallback` to make sure polling never runs when hi-pri startup work is underway.
* Cooperative multitasking, to make sure that we yield the thread between each wallet detector run.

steveluscher:
Haha. Missed the merge by 4 minutes.

steveluscher:
Ignore everything above, I'll send a follow up.

steveluscher:
Closes #317.

PR #348, opened 2022-03-04
Race all scope polling strategies at once; first one wins

steveluscher:
_This is a follow up to #346_.

---

In this PR we refactor the scope polling strategies into a race between the relevant ones. The first one that wins cancels all others.

1. We always fire the detection synchronously.
2. We always employ the interval polling strategy.
3. IF the document is not yet `interactive` or `complete` we employ the `DOMContentLoaded` strategy.
4. IF the document is not yet `complete` we employ the `window` `load` event strategy.

In any event, the first strategy to result in the detection of the given wallet cancels all of the others.

## Test plan

Loaded the React starter (not a very exhaustive test plan).

https://user-images.githubusercontent.com/13243/156705441-6d819ea7-c020-4051-84af-ef4d3b4e8324.mov

### DIFF
--- a/packages/core/base/src/adapter.ts
+++ b/packages/core/base/src/adapter.ts
@@ -89,41 +89,52 @@ export abstract class BaseWalletAdapter extends EventEmitter<WalletAdapterEvents
     ): Promise<TransactionSignature>;
 }
 
+type DisposeFn = () => void;
+
 export function scopePollingDetectionStrategy(detect: () => boolean): void {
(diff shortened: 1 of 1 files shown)


steveluscher:
cc/ @safaiyeh.

safaiyeh:
@steveluscher appreciate the follow up! Looks good, Strat #2 will fix our use case

safaiyeh (approved):

PR #370, opened 2022-03-26
Add WANA Wallet Adapter

twodayslate:
WANA Wallet is a fork of Sollet with Safari Extension support since Sollet is no longer supported.


jordansexton:
This PR looks like it duplicates all the code in the Sollet adapter, despite importing `@project-serum/sol-wallet-adapter`. It appears to add no new functionality to it. Ideally this would extend the [`BaseSolletWalletAdapter`](https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/sollet/src/base.ts) and add meaningful functionality.

It also includes 2 icon images, one of which is a large PNG. Overall, I'd prefer to see a refactor to avoid so much duplication, and also some explanation for why basing a new wallet on a deprecated browser wallet and library is helpful. Please open a new PR if that makes sense!

twodayslate:
@jordansexton 

As far as I am aware WANA Wallet is the only Solana wallet that supports Safari Extensions for iOS. The brings an excellent and native user experience to Solana users and developers. There is no need to leave Safari to connect to a wallet or sign a transaction - it just works!

Currently WANA Wallet does have the `sollet` window property defined to ensure some backwards compatibility with Sollet - however, this is not future proof. The `wana` window property would be preferred so extending `BaseSolletWalletAdapter ` is not the ideal solution in my opinion. This ensures WANA Wallet can continue to build new features for users and users know that WANA Wallet is not the same thing as Sollet. Please let me know if I'm wrong in this or am missing something obvious.

The PNG is 128x128. Is there a recommended size for the icon? Is SVG preferred? There doesn't seem to be any consistency in the current code base between the different wallets.

Thank you!

PR #375, opened 2022-03-30
Better TypeScript Hints for WalletName

biw:
This PR adds an optional type (no breaking changes) on the `WalletName` to support better typescript hints. 

Previously, in application code, the user would see the following as a hint:
Screen Shot 2022-03-30 at 2 32 57 PM

With this PR, they will see:
Screen Shot 2022-03-30 at 2 34 06 PM


jordansexton:
Beautiful, thanks!

PR #381, opened 2022-04-06
Add default getting started wallet param to modal

safaiyeh:
Add the ability to choose which wallet to default to for the get started button.


jordansexton:
I don't think we will expand this API.

jordansexton:
The getting started button was removed in https://github.com/solana-labs/wallet-adapter/pull/680, and any "default" wallets were removed previously. Now the order is strictly consistent between the UI packages.

PR #382, opened 2022-04-09
Add Exodus wallet adapter

olistic:
## Summary

This PR adds the wallet adapter for the Exodus wallet for all platforms, including the soon-to-be-released browser extension.

@jordansexton I'll let you know when the browser extension launches in beta and this can be tested. Opening as draft in the meantime.


FAQ.md
```diff
@@ -101,6 +101,7 @@ Please see [issue #72](https://github.com/solana-labs/wallet-adapter/issues/72#i
 
 Some wallet adapters provide a `signMessage` method for signing arbitrary bytes:
 
+- [Exodus](https://github.com/solana-labs/wallet-adapter/tree/master/packages/wallets/exodus)
```
olistic, line 104: Respects existing alphabetical order.

olistic, line 104: Respects existing alphabetical order.

README.md
```diff
@@ -80,6 +81,7 @@ export const Wallet: FC = () => {
     const wallets = useMemo(
         () => [
             new PhantomWalletAdapter(),
+            new ExodusWalletAdapter(),
```
olistic, line 84: Order here (and in the other arrays below) can be whatever makes sense.

olistic, line 84: Order here (and in the other arrays below) can be whatever makes sense.

olistic:
@jordansexton What's a good way of sending you an Exodus Browser Extension build?

olistic:
@jordansexton Will rebase and update this PR today.

jordansexton:
No problem! I've been pretty swamped, but let me know when this is ready to review.

olistic:
@jordansexton This is ready for review.

Question: once approved, how can we time the merging+release of `@solana/wallet-adapter-exodus` and `@solana/wallet-adapter-wallets` with the release of our browser extension?

README.md
jordansexton: Please remove this change and the changes to the starter projects. We aren't making changes to those at this time, but we actively encourage devs using the project to choose wallets that make sense for their users.

packages/wallets/exodus/src/adapter.ts
```diff
@@ -0,0 +1,268 @@
...
+        options?: SendOptions
+    ): Promise<{ signature: TransactionSignature }>;
+    signMessage(message: Uint8Array): Promise<{ signature: Uint8Array }>;
+    connect(): Promise<void>;
+    disconnect(): Promise<void>;
+    _handleDisconnect(...args: unknown[]): unknown;
+}
+
+interface ExodusWindow extends Window {
+    solana?: ExodusWallet;
```
jordansexton, line 43: Can you attach the object to `window.exodus`? We aren't accepting new adapters that make use of `window.solana`, which hurts our ability to reliably detect all wallets. Wallets that already use it have been asked to migrate. Medium term, the plan is to push all wallets to implement https://github.com/solana-labs/wallet-adapter/issues/215 and resolve detection issues across the board.

packages/wallets/exodus/src/adapter.ts
```diff
@@ -0,0 +1,268 @@
...
+            if (this._readyState !== WalletReadyState.Installed) throw new WalletNotReadyError();
+
+            this._connecting = true;
+
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const wallet = window!.solana!;
+
+            if (!wallet.isConnected) {
+                // HACK: Exodus doesn't reject or emit an event if the popup is closed
+                const handleDisconnect = wallet._handleDisconnect;
```
jordansexton, line 112: Hmm this code looks like it was copied from the Phantom adapter. The point of this `HACK` comment is to show a hack around a broken behavior in Phantom. You probably don't want to copy it, or the behavior. :) Any code that's in here should pertain to your actual wallet extension/app (which hopefully isn't broken).

packages/wallets/exodus/src/adapter.ts
```diff
@@ -0,0 +1,268 @@
...
+        transaction: Transaction,
+        connection: Connection,
+        options?: SendTransactionOptions
+    ): Promise<TransactionSignature> {
+        try {
+            const wallet = this._wallet;
+            // Exodus doesn't handle partial signers, so if they are provided, don't use `signAndSendTransaction`
+            if (wallet && 'signAndSendTransaction' in wallet && !options?.signers) {
+                // HACK: Exodus's `signAndSendTransaction` should always set these, but doesn't yet
+                transaction.feePayer = transaction.feePayer || this.publicKey || undefined;
```
jordansexton, line 192: Likewise -- this should technically be marked with `HACK` but in any case, don't copy the Phantom adapter.

jordansexton (changes requested):
Thanks for the adapter! Please review the changes to the examples, and especially the hacky stuff you inherited from copying the Phantom adapter.

packages/wallets/exodus/src/adapter.ts
jordansexton: Also, HUGE bonus points if you handle partial signers properly out of the gate (these are used by Solana Pay transaction requests too!)

packages/wallets/exodus/src/adapter.ts
olistic: > Can you attach the object to window.exodus?

Yes, addressed in f3f4b33. We were doing both `window.solana` and `window.exodus.solana`, but will pass this on to the team doing the injection.

packages/wallets/exodus/src/adapter.ts
olistic: That is correct. As we made our Solana Provider OOTB compatible with Phantom's when it comes to signatures and return values, we copied the Phantom adapter verbatim. We've made improvements though, like throwing a `UserRejectedRequestError` with `code=4001` if the popup is closed in this case. Cleaning up.

packages/wallets/exodus/src/adapter.ts
olistic: Removing as we're setting this already internally.

olistic: > Also, HUGE bonus points if you handle partial signers properly out of the gate (these are used by Solana Pay transaction requests too!)

Adding this as a follow up, will address as soon as we have the bandwidth in a future PR.

jordansexton:
Thanks for your hard work on this! ❤️

olistic:
It's been a pleasure!

Issue #340, opened 2022-03-02
Set 'react' dependency as a peerDependency


charly-helio:
**Describe the bug**
If we have in **Project A** installed the **solana-adapter-react** as a regular dependency, we can have problems of React versions colliding when linking a **Project A** to a **Project B**. It should be set as a **peerDependency** in **package.json**. 

Since we are going to use (I think that always) the **solana-adapter-react** in an existent React project, we don't need it as a dependency.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a React project (Project A)  using the **solana-adapter-react**
2. Create a React project (Project B) with **react** as a regular dependency.
3. To do it easier, in **Project A** type `yarn link` and then in **Project B** type `yarn link 'name-of-the-project-B'`
4. Run `yarn start` in **Project B** (and, for instance, using any hook such as `useState`)
5. If you go to your browser and check the console, you will see this message:

_> Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 
> You might have mismatching versions of React and the renderer (such as React DOM)
> You might be breaking the Rules of Hooks
> You might have more than one copy of React in the same app See for tips about how to debug and fix this problem._

**Expected behaviour**
Don't see any error and the app to work properly.

**Desktop (please complete the following information):**
 - OS: OS 12.2.1 (Monterey)
 - Browser: Chrome 
 - Version: 98.0.4758.109 (Official Build) (x86_64) 

**Posible solution**
I fixed it setting **wallet-adapter-react** as a **peerDependency** in the **Project A** (the one using it in the code) and installing it in the **Project B** (the one using **Project A** linked) as a regular dependency. And since the **Project A** has also **react** as a peerDependency, there is no problem. I mean, **Project B** is the one serving to all the linked libraries/dependencies the **react** library to be used.

I have this in the Project A package.json:
```
  "peerDependencies": {
    "@solana/wallet-adapter-react": "^0.15.3",
    "react": "^17.0.2",
    "react-dom": "^17.0.2"
  },
```

And then in Project B package.json this:
```
  "dependencies": {
    "@solana/wallet-adapter-react": "^0.15.3",
    "react": "^17.0.2",
    "react-dom": "^17.0.2"
  },
```

So, If I'm not wrong setting the **react** library as a **peerDependency** can fix this issue (and I guess that others related as well).

Thanks! :)
Charly.

asktree:
[...]

PR #389, opened 2022-04-14
updated all packages to use @solana/web3.js as a peer dependency

biw:
Going off the same issues as outlined in #340, this PR updates all packages to use `@solana/web3.js` as a peer dependency. 

I don't see anything breaking with this change as the readme directs users to install `@solana/web3.js` as a dependency.


jordansexton:
So this looks good, but I think these packages now need web3.js as a dev dependency so it gets installed when building each one from source.

biw:
Good catch, updated! Thanks @jordansexton

alirezaayande (approved):

yarn.lock
```diff
@@ -2300,6 +2300,20 @@
     prop-types "^15.7.2"
     react-is "^17.0.2"
 
+"@mui/base@5.0.0-alpha.77":
```
jordansexton, line 2303: Hmm, can you figure out why this is getting included here? It seems like this file shouldn't need to change at all if we just moved some deps into peer and dev deps, right?

jordansexton:
Looks great, I made the order of peer/deps/devDeps consistent. I'm wondering why the yarn lockfile changes when running `yarn install` in this branch though.

jordansexton:
Nevermind the yarn.lock issue, it looks like it's just a result of some nested dependency update. Thanks for this!

PR #391, opened 2022-04-21
Streambird adapter

quangas:
Streambird is a passwordless authentication platform for Web3. We created this Solana Wallet Adapter to make it easier for Web3 dapps/companies using Streambird to onboard users onto the Solana ecosystem thru a Streambird provisioned Solana wallet.

* We created a new wallet adapter that implements the Solana wallet-adapter.
* It supports all the methods like sign transaction, sign all transactions and sign message.
* Our approach when the wallet is clicked, it will trigger the following user flow

1. it will show a screen to enter in your email
2. asks you to verify your email 
3. and then once verified will provide the public wallet address (PublicKey).


quangas:
@jordansexton Hi Jordan, what else do we need to do in order to get this wallet merged?

jordansexton:

packages/starter/react-ui-starter/src/App.tsx
```diff
@@ -7,6 +7,7 @@ import {
     SlopeWalletAdapter,
     SolflareWalletAdapter,
     TorusWalletAdapter,
+    StreambirdWalletAdapter
```
jordansexton, line 10: Please limit changes to adding the package only. We don't add all new wallets to the example or starter projects.

packages/wallets/streambird/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "scripts": {
+        "clean": "shx rm -rf lib/*",
+        "build": "yarn clean && tsc -p tsconfig.json && tsc-esm -p tsconfig.json && tsc -p tsconfig.cjs.json",
+        "postbuild": "echo '{\"type\":\"commonjs\"}' | npx json > lib/cjs/package.json && echo '{\"type\":\"module\"} ' | npx json > lib/esm/package.json"
+    },
+    "dependencies": {
+        "@solana/wallet-adapter-base": "^0.9.4",
+        "@solana/web3.js": "^1.36.0",
+        "@streambird/solana-embed-sdk": "^1.0.4"
+    }
```
jordansexton, line 33: Can you update this to the peer/dev deps structure seen here? https://github.com/solana-labs/wallet-adapter/blob/e230958570f192786a27c57d5b39037b898e69dd/packages/wallets/solflare/package.json#L29-L38

packages/wallets/streambird/src/adapter.ts
```diff
@@ -0,0 +1,189 @@
...
+        throw new WalletConnectionError(error?.message, error);
+      }
+
+      let existingPublicKey: PublicKey | null = null;
+
+      try {
+        existingPublicKey = new PublicKey(base58PublicKey);
+      } catch (error: any) {
+        throw new WalletPublicKeyError(error?.message, error);
+      }
```
jordansexton, line 109: Hmm, you don't need to assign empty string or `null` here. This chunk of code should probably be almost exactly the same as this one: https://github.com/solana-labs/wallet-adapter/blob/e230958570f192786a27c57d5b39037b898e69dd/packages/wallets/clover/src/adapter.ts#L86-L98

packages/wallets/streambird/src/adapter.ts
```diff
@@ -0,0 +1,189 @@
...
+    const wallet = this._wallet;
+
+    if (wallet) {
+      this._wallet = null;
+      this._publicKey = null;
+
+      try {
+        if(wallet.isLoggedIn) {
+          await wallet.cleanUp();
+        }
```
jordansexton, line 133: If it's not actually required for the adapter to wait until `cleanUp` finishes, it might be good to just ignore the promise. What would actually lead to an error here?

packages/wallets/streambird/src/adapter.ts
```diff
@@ -0,0 +1,189 @@
...
+        ({ default: StreambirdClass } = await import('@streambird/solana-embed-sdk'));
+      } catch (error: any) {
+        throw new WalletLoadError(error?.message, error);
+      }
+
+      let wallet: Streambird;
+
+      try {
+        wallet = window.streambird || new StreambirdClass();
+      } catch (error: any) {
```
jordansexton, line 83: If `window.streambird` is defined, do you still need to load the SDK?

jordansexton (commented):

quangas:
In that case, we'll just wait for the new standard.

jordansexton:
> In that case, we'll just wait for the new standard.

I want to point out that the wallet standard will help in the case that Streambird is a browser extension or mobile wallet, but it won't help you if Streambird is a `Loadable` wallet which requires loading an external library.

PR #396, opened 2022-04-24
Removed "didUpdate" condition

MrAkblt:
Removed "didUpdate" condition for better state and localStorage synchronization.


jordansexton:
I think this code is working as intended. If you think we should remove it, please provide more context.

PR #407, opened 2022-05-05
Pass network to wallet when calling sign tx in Glow

vpontis:
Glow is different than other wallets in that we require the dApp to specify the network rather than falling back to the wallet the user chose in the Extension. 

This is for a few reasons:

- the dApp should know what network the transaction should be on
- we don't allow approving transactions if simulation fails and mis-configured network is a common reason for simulation failure


packages/wallets/glow/src/adapter.ts
```diff
@@ -22,12 +22,14 @@ interface GlowWalletEvents {
     disconnect(...args: unknown[]): unknown;
 }
 
+type Network = 'mainnet' | 'devnet'
```
jordansexton, line 25: Could you use the `WalletAdapterNetwork` type here?

https://github.com/solana-labs/wallet-adapter/blob/a827686dbc109ef43b5da0031811cb5013aaaac8/packages/core/base/src/types.ts#L6-L10

Ideally this type should be compatible with the other adapters.

packages/wallets/glow/src/adapter.ts
```diff
@@ -174,7 +176,7 @@ export class GlowWalletAdapter extends BaseMessageSignerWalletAdapter {
     async sendTransaction(
         transaction: Transaction,
         connection: Connection,
-        options?: SendTransactionOptions
+        options?: SendTransactionOptions & { network: Network }
```
jordansexton, line 179: Extending these options is no good, unfortunately. However, you can add the option for glow to the constructor of the adapter like this: https://github.com/solana-labs/wallet-adapter/blob/a827686dbc109ef43b5da0031811cb5013aaaac8/packages/wallets/solflare/src/adapter.ts#L29-L31

Store the config in the adapter's state, then pass it to your wallet extension's methods where needed.

jordansexton (changes requested):
Thanks! If you could just refactor this a little bit, it'll be good to go.

packages/wallets/glow/src/adapter.ts
vpontis: Hmm, I think that's a bit annoying for developers. They should be able to decide at the time of the sending tx what network it should go to, it doesn't need to be picked when initializing the `wallet`.

packages/wallets/glow/src/adapter.ts
vpontis: Will update 👍

packages/wallets/glow/src/adapter.ts
jordansexton: What's the user story for needing to pick it except when connecting to the wallet? Dapps aren't really built around cluster switching, and if they do, most wallets use this on init.

jordansexton: But anyway, I don't want to change the interface for this, especially because we are looking to standardize the API soon (which will reduce the need to support need adapters at all).

vpontis: > What's the user story for needing to pick it except when connecting to the wallet?

It makes sense to switch between `mainnet` / `devnet` if you want to run something as a test on `devnet` before running it on `mainnet`.

vpontis:
@jordansexton updated w/ your feedback :)

PR #411, opened 2022-05-09
Add Ledger Connect adapter

harrisonf-ledger:
Adds support for the upcoming [Ledger Connect Safari extension](https://get-connect.ledger.com).


jordansexton:

packages/wallets/ledgerconnect/src/adapter.ts
```diff
@@ -0,0 +1,236 @@
...
+                    throw error;
+                }
+                throw new WalletConnectionError(error?.message, error);
+            }
+
+            if (!wallet.publicKey) {
+                throw new WalletAccountError();
+            }
+
+            const publicKey: PublicKey = wallet.publicKey;
```
jordansexton, line 133: Hmm, it could be an issue to use the same referential instance of `publicKey` here. All the other adapters make a copy from a string or bytes.

packages/wallets/ledgerconnect/src/adapter.ts
```diff
@@ -0,0 +1,236 @@
...
+                if (typeof event.data === 'object' && event.data.__ledgerConnect_loaded) {
+                    if (this._readyState !== WalletReadyState.Installed) {
+                        this._readyState = WalletReadyState.Installed;
+                        this.emit('readyStateChange', this._readyState);
+                    }
+                    window.removeEventListener('message', handler);
+                }
+            };
+
+            window.addEventListener('message', handler);
```
jordansexton, line 78: Hmm, adding an event listener here will cause a memory leak whenever the adapter is instantiated, because it's only removed when the event is fired. This can happen very easily in the context of React applications.

packages/wallets/ledgerconnect/src/adapter.ts
```diff
@@ -0,0 +1,236 @@
...
+
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const wallet = window!.ledgerConnectSolana!;
+
+            try {
+                await wallet.connect();
+            } catch (error: any) {
+                if (error instanceof WalletError) {
+                    throw error;
+                }
```
jordansexton, line 125: This kind of check isn't necessary, because `wallet` would never throw `WalletError` (which is thrown by this library). The `instanceof` check would never match.

packages/wallets/ledgerconnect/src/adapter.ts
```diff
@@ -0,0 +1,236 @@
...
+            // Ledger Connect doesn't handle partial signers, so if they are provided, don't use `signAndSendTransaction`
+            if (wallet && 'signAndSendTransaction' in wallet && !options?.signers) {
+                // HACK: Ledger Connect's `signAndSendTransaction` should always set these, but doesn't yet
+                transaction.feePayer = transaction.feePayer || this.publicKey || undefined;
+                transaction.recentBlockhash =
+                    transaction.recentBlockhash || (await connection.getLatestBlockhash('finalized')).blockhash;
+
+                const { signature } = await wallet.signAndSendTransaction(transaction, options);
+                return signature;
+            }
```
jordansexton, line 183: Please don't copy the Phantom adapter's workarounds. Partially signed transactions should also be supported.

jordansexton (commented):

jordansexton:
Closing this PR as stale, but I'm happy to reopen it if you want to continue development.

PR #422, opened 2022-05-19
Retry Glow wallet detection

sjchmiela:
We've noticed an elevated number of errors on some pages which seem to be trying to connect to Glow wallet before it gets to register itself (in Safari extension). While this technically shouldn't happen since Safari extension content scripts are expected to run on `document_start`, [so before any page scripts](https://developer.chrome.com/docs/extensions/mv2/content_scripts/#run_time), some Safari settings (Preload Top Hit) may break this behavior as noted in [this article](https://lapcatsoftware.com/articles/2021-6-11.html).

This pull request aims to mitigate this behavior by waiting for wallet to register itself in `connect()`. If after 3 seconds the wallet still isn't ready we throw `WalletNotReadyError` as we do right now. Similar solution has been implemented in https://github.com/saber-hq/saber-common/pull/500.

### DIFF
--- a/packages/wallets/glow/src/adapter.ts
+++ b/packages/wallets/glow/src/adapter.ts
@@ -109,10 +109,37 @@ export class GlowWalletAdapter extends BaseMessageSignerWalletAdapter {
         return this._readyState;
     }
 
+    private async awaitReadyState() {
+        if (this._readyState === WalletReadyState.Installed) {
+            return;
+        }
+
+        // Some webpages try to connect to wallet before
+        // it is registered. Let's wait 3 seconds
+        // for wallet registration.
+        const MAX_RETRIES = 3;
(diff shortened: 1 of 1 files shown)


sjchmiela:
From my understanding of the issue we saw that prompted this pull request it was dApps (or, at least ome) which were manually calling connect on adapters (maybe one they saw was used last?) before _any_ wallet could have set itself up properly (see Safari bug mentioned in the description). Just making sure it’s clear we don’t mark our wallet ready while it’s not or something.

Fortunately the issue magically went away after a couple of days.

jordansexton:
Huh, that's good. Well, apps shouldn't call connect if the adapter isn't ready, so that's on them! It's possible that it was a bug in the React autoconnect logic that was fixed at some point.

PR #426, opened 2022-05-23
add Nightly wallet adapter

awojciak:
Adapter for Nightly wallet.


NorbertBodziony:
If for review, extension will be needed please ping me. I will provide internal link.

awojciak:
@jordansexton hi, I see that this PR didn't get any feedback so far, is there anything else needed besides link to extension?

packages/starter/example/components/ContextProvider.tsx
```diff
@@ -7,6 +7,7 @@ import { ConnectionProvider, WalletProvider } from '@solana/wallet-adapter-react
 import { WalletModalProvider as ReactUIWalletModalProvider } from '@solana/wallet-adapter-react-ui';
 import {
     GlowWalletAdapter,
+    NightlyWalletAdapter,
```
jordansexton, line 10: Please limit changes to adding the package only. We don't add all new wallets to the example or starter projects.

packages/wallets/nightly/src/adapter.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    get connecting(): boolean {
+        return this._connecting;
+    }
+
+    get readyState(): WalletReadyState {
+        return this._readyState;
+    }
+
+    public async signAllTransactions(transactions: Transaction[]): Promise<Transaction[]> {
```
jordansexton, line 66: nit: could you reorder the methods to match the other adapters? It just makes review and maintenance of all these easier.

packages/wallets/nightly/src/types.ts
```diff
@@ -0,0 +1,25 @@
...
+    solana: SolanaNightly;
+    private readonly _nightlyEventsMap;
+    constructor();
+    invalidate(): void;
+}
+
+export declare class SolanaNightly {
+    publicKey: SolanaPublicKey;
+    _onDisconnect: () => void;
+    private readonly _nightlyEventsMap;
```
jordansexton, line 13: Don't think you need to declare private / `_`-prefixed props and methods in the exported type?

packages/wallets/nightly/src/adapter.ts
```diff
@@ -0,0 +1,150 @@
...
+            } catch (error: any) {
+                throw new WalletSignTransactionError(error?.message, error);
+            }
+        } catch (error: any) {
+            this.emit('error', error);
+            throw error;
+        }
+    }
+
+    async connect(onDisconnect?: () => void) {
```
jordansexton, line 104: Changing the method signature of `connect` isn't allowed. This needs to match https://github.com/solana-labs/wallet-adapter/blob/e230958570f192786a27c57d5b39037b898e69dd/packages/core/base/src/adapter.ts#L31

packages/wallets/nightly/src/adapter.ts
```diff
@@ -0,0 +1,150 @@
...
+        try {
+            if (this.connected || this.connecting) return;
+            if (this._readyState !== WalletReadyState.Installed) throw new WalletNotReadyError();
+
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const wallet = window!.nightly!.solana!;
+
+            try {
+                this._connecting = true;
+                await wallet.connect(onDisconnect);
```
jordansexton, line 114: Similarly, if you need to acquire a reference to `onDisconnect`, this has to be implemented here (by adding a method to handle disconnection, as seen in other adapters).

packages/starter/example/components/ContextProvider.tsx
awojciak: ok, removed

packages/wallets/nightly/src/types.ts
```diff
@@ -0,0 +1,25 @@
...
+    disconnect(): Promise<void>;
+    signTransaction(tx: SolanaTx): Promise<SolanaTx>;
+    signAllTransactions(txs: SolanaTx[]): Promise<SolanaTx[]>;
+}
+
+export interface NightlyWindow extends Window {
+    nightly?: {
+        solana?: SolanaNightly
+    };
+}
```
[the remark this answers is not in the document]
awojciak: moved

packages/wallets/nightly/src/adapter.ts
awojciak: done

packages/wallets/nightly/src/adapter.ts
```diff
@@ -0,0 +1,150 @@
...
+            } catch (error: any) {
+                if (error instanceof WalletError) throw error;
+                throw new WalletConnectionError(error?.message, error);
+            }
+
+            this._connected = true;
+            this._wallet = wallet;
+            this._publicKey = wallet.publicKey;
```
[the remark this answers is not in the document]
awojciak: done

packages/wallets/nightly/src/types.ts
awojciak: my bad, I've overlooked these, removed

packages/wallets/nightly/src/adapter.ts
awojciak: callback removed

packages/wallets/nightly/src/adapter.ts
awojciak: removed

packages/wallets/nightly/src/adapter.ts
```diff
@@ -0,0 +1,174 @@
...
+    WalletNotConnectedError,
+    WalletNotReadyError,
+    WalletPublicKeyError,
+    WalletReadyState,
+    WalletSignTransactionError,
+} from '@solana/wallet-adapter-base';
+import { PublicKey, Transaction } from '@solana/web3.js';
+
+interface SolanaNightly {
+    publicKey: PublicKey;
```
jordansexton, line 18: Should this be `publicKey?: PublicKey;` or something? Wondering what the value is when the wallet isn't connected since it's accessed below

packages/wallets/nightly/src/adapter.ts
```diff
@@ -0,0 +1,174 @@
...
+            } catch (error: any) {
+                if (error instanceof WalletError) throw error;
+                throw new WalletConnectionError(error?.message, error);
+            }
+
+            if (!wallet.publicKey) throw new WalletAccountError();
+
+            let publicKey: PublicKey;
+            try {
+                publicKey = new PublicKey(wallet.publicKey.toString());
```
jordansexton, line 105: `wallet.publicKey.toBytes()` will be faster and is what we use elsewhere, you don't need to serialize to base58 and then deserialize back

packages/wallets/nightly/src/adapter.ts
```diff
@@ -0,0 +1,174 @@
...
+            if (this._readyState !== WalletReadyState.Installed) throw new WalletNotReadyError();
+
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const wallet = window!.nightly!.solana!;
+
+            try {
+                this._connecting = true;
+                await wallet.connect();
+            } catch (error: any) {
+                if (error instanceof WalletError) throw error;
```
jordansexton, line 97: `error instanceof WalletError` will never be true (the error is coming from the wallet), so this check isn't needed.

jordansexton (commented):
Looks good, only a few small notes

packages/wallets/nightly/src/adapter.ts
awojciak: it's set to '11111111111111111111111111111111' when not connected

packages/wallets/nightly/src/adapter.ts
awojciak: removed

packages/wallets/nightly/src/adapter.ts
awojciak: done

packages/wallets/nightly/src/adapter.ts
jordansexton: Isn't it the case then that the check on L99 (`!wallet.publicKey`) will never fail, even if there's no wallet to connect to?

awojciak: I must have overlooked it; now I've changed the condition, I hope everything will be fine now

PR #431, opened 2022-05-25
Add wallet adapter for NuFi

gyorgy-kurucz-vacuum:
This adds support for the upcoming NuFi extension.

(The interface is based on Phantom's, they should be compatible, except for changing the `window.solana` key to `window.nufiSolana` of course.)


jordansexton:

packages/wallets/nufi/src/adapter.ts
```diff
@@ -0,0 +1,232 @@
...
+                        wallet.connect().catch((reason: any) => {
+                            wallet.off('connect', connect);
+                            reject(reason);
+                        });
+                    });
+                } catch (error: any) {
+                    if (error instanceof WalletError) throw error;
+                    throw new WalletConnectionError(error?.message, error);
+                }
+            }
```
jordansexton, line 104: It looks like this was all copied from the Phantom adapter, which has hacks to access events because of bad connection behavior. Please don't copy these things!

packages/wallets/nufi/src/adapter.ts
```diff
@@ -0,0 +1,232 @@
...
+            const wallet = this._wallet;
+            // Partial signers are not handled, so if they are provided, don't use `signAndSendTransaction`
+            if (wallet && !options?.signers) {
+                transaction.feePayer = transaction.feePayer || this.publicKey || undefined;
+                transaction.recentBlockhash =
+                    transaction.recentBlockhash || (await connection.getRecentBlockhash('finalized')).blockhash;
+
+                const { signature } = await wallet.signAndSendTransaction(transaction);
+                return signature;
+            }
```
jordansexton, line 162: Same here, please don't copy the Phantom adapter's workarounds. Partially signed transactions should also be supported.

packages/wallets/nufi/src/publicTypes.ts
```diff
@@ -0,0 +1,80 @@
+/* This file should be kept in sync between the NuFi and the `wallet-adapter`
+ * codebases. */
```
jordansexton, line 2: Please inline this with the adapter file, and only include the types needed by the adapter. Docs aren't needed because it's an internal type.

jordansexton (commented):

packages/wallets/nufi/src/adapter.ts
gyorgy-kurucz-vacuum: Thanks, fixed.

packages/wallets/nufi/src/adapter.ts
gyorgy-kurucz-vacuum: Thanks, fixed.

packages/wallets/nufi/src/publicTypes.ts
gyorgy-kurucz-vacuum: Thanks, I inlined it.

gyorgy-kurucz-vacuum:
Thank you for the review, feel free to point out if there are any other redundant parts left as a result of copying the Phantom code that I didn't catch.

packages/wallets/nufi/src/adapter.ts
```diff
@@ -0,0 +1,289 @@
...
+interface NufiEvents {
+    connect(): void;
+    disconnect(): void;
+}
+
+/** The type of the connector object injected by the NuFi extension.
+ *
+ * Note that the interface uses some objects from `@solana/web3.js`. It is
+ * guaranteed that NuFi uses at least version `1.42.0` of this library.
+ */
```
jordansexton, line 28: In the interest of brevity, these comments can probably all be removed since they document a purely internal interface.

packages/wallets/nufi/src/adapter.ts
```diff
@@ -0,0 +1,289 @@
...
+declare const window: NufiWindow;
+
+export interface NufiWalletAdapterConfig {}
+
+export const NufiWalletName = 'NuFi' as WalletName<'NuFi'>;
+
+export class NufiWalletAdapter extends BaseMessageSignerWalletAdapter {
+    name = NufiWalletName;
+    url = 'https://nu.fi';
+    icon =
```
jordansexton, line 108: nit: line break here, can you run Prettier (`yarn run fmt`) on the project root?

packages/wallets/nufi/src/adapter.ts
```diff
@@ -0,0 +1,289 @@
...
+    async signMessage(message: Uint8Array): Promise<Uint8Array> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const { signature } = await wallet.signMessage(message);
+                return signature;
+            } catch (error: any) {
+                throw new WalletSignTransactionError(error?.message, error);
```
jordansexton, line 269: This should be `WalletSignMessageError` I think. Will also need to import.

jordansexton: FWIW, this issue was found and fixed with other adapters today: https://github.com/solana-labs/wallet-adapter/commit/112c54463a3d069e5b756dacf78e20a0711939ae

packages/wallets/nufi/src/adapter.ts
gyorgy-kurucz-vacuum: It is prettier that introduces this line break. (Running `yarn run fmt` does not change anything.)

packages/wallets/nufi/src/adapter.ts
gyorgy-kurucz-vacuum: I very much disagree. This is the interface between the wallet and the adapter, and it's anything but internal. I feel like a precise definition of this interface is important. And we don't have this interface documentation published anywhere else at the moment.

Nonetheless, I removed them for now.

packages/wallets/nufi/src/adapter.ts
gyorgy-kurucz-vacuum: Thanks, fixed.

jordansexton:
Thanks! This will be published today or tomorrow.

jordansexton:

PR #432, opened 2022-06-01
fix: handle accountChanged event for phantom

ZhengYuTay:
There's a scenario where in phantom, when you change wallet for a trusted domain, it will call `accountChanged` instead of `disconnect` event, this has caused quite a number of transaction failed from our analytic at Jupiter.


jordansexton:
Thank you!

PR #433, opened 2022-06-03
update torus solana embed v0.2.1

ieow:
update torus Solana embed v0.2.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,6 @@
         "typescript-esm": "^2.0.0"
     },
     "resolutions": {
-        "**/@toruslabs/solana-embed/@solana/web3.js": "=1.36.0"
+        "**/@toruslabs/solana-embed/@solana/web3.js": "^1.36.0"
     }
 }
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
         "@typescript-eslint/eslint-plugin": "^5.14.0",
         "@typescript-eslint/parser": "^5.14.0",
         "babel-eslint": "^10.1.0",
+        "crypto-browserify": "^3.12.0",
         "eslint": "^8.11.0",
         "eslint-config-prettier": "^8.5.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -52,5 +52,8 @@
         "typedoc": "^0.22.13",
         "typescript": "^4.6.2",
         "typescript-esm": "^2.0.0"
+    },
+    "resolutions": {
+        "**/@toruslabs/solana-embed/@solana/web3.js": "=1.36.0"
```
jordansexton, line 57: We can't do this. This will cause the end user developer to end up with two versions of web3.js in their code. In addition to bloat, this will cause all kinds of compile-time and runtime issues (such as type mismatches). Torus should probably update their version of web3.js used so that it's appropriately permissive (i.e., use `^`).

ieow: @toruslabs/solana-embed is using web3.js version `^1.43.0`
I am having issue to build the wallet-adapter repo as it complaint about two version of web3.js (1.36.0 in adapter and 1.43.0 in torus)
Should torus downgrade the version back to `^1.36.0` ?

ieow:
@jordansexton , I update the Solana-embed with web3.js `1.36.0`
cc @metallicalfa2

packages/wallets/torus/package.json
```diff
@@ -31,21 +31,24 @@
     },
     "dependencies": {
         "@solana/wallet-adapter-base": "^0.9.4",
-        "@toruslabs/solana-embed": "^0.1.5",
+        "@toruslabs/solana-embed": "^0.2.2",
```
jordansexton, line 34: Does this version update have any breaking API changes we should be concerned with?

ieow: No. There is no breaking API changes.
The minor version update is about the changes were in controllers only and doesn't affect the APIs used wallet adapter

jordansexton:

PR #437, opened 2022-06-09
Add signMessage function to Bitkeep adapter

mauser-pmon:
Add signMessage function to the Bitkeep adapter. 

Code is copied from Phantom adapter, so should be straight forward.

### DIFF
--- a/packages/wallets/bitkeep/src/adapter.ts
+++ b/packages/wallets/bitkeep/src/adapter.ts
@@ -19,6 +19,7 @@ interface BitKeepWallet {
     getAccount(): Promise<string>;
     signTransaction(transaction: Transaction): Promise<Transaction>;
     signAllTransactions(transactions: Transaction[]): Promise<Transaction[]>;
+    signMessage(message: Uint8Array): Promise<{ signature: Uint8Array }>;
 }
 
 interface BitKeepWindow extends Window {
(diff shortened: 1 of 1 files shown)


jordansexton:
Thanks @mauser-pmon!

> WIP: Just found out that the BitKeep Chrome Addon uses strings instead of Uint8Array for input/output of signMessage. Need to refactor more.

Is this still true?

mauser-pmon:
> 

@jordansexton : Yes, just tried with the latest Version of the BitKeep Chrome Wallet. When I pass a Uint8Array to the SignMessage method of window.bitkeep.solana, I see an error in the Wallet 'failed TypeError: Expected String" when trying to sign the message. 
I tried to make a work around by converting it to text before passing it in, but the result was now correct or the same as for the other wallets.

jordansexton:
> When I pass a Uint8Array to the SignMessage method of window.bitkeep.solana, I see an error in the Wallet 'failed TypeError: Expected String" when trying to sign the message.

I'm not sure I understand. It sounds like it may be required to detect this in Chrome somehow? Or get Bitkeep to make their implementation consistent? I don't think we should merge this if it's broken in Chrome, or works differently in different browsers.

mauser-pmon:
> 
> I'm not sure I understand. It sounds like it may be required to detect this in Chrome somehow? Or get Bitkeep to make their implementation consistent? I don't think we should merge this if it's broken in Chrome, or works differently in different browsers.

I am trying to get in touch with the Bitkeep team to add/fix the Solana signMessage implementation for their Wallets (mobile/chrome extension) ... had no luck until now to get an answer.

keepresolve:
Our uint8array type is changed to hex communication

mauser-pmon:
> SignMessage is supported in v1.2.0 . chrome extensionsThe latest version is v1.2.1. you can try it

Yes thanks, it seems to work now in the chrome extension. In the Mobile Bitkeep apps it is still only uses strings unfortunately.

mauser-pmon:
@jordansexton : I have now the confirmation from the BitKeep team that also the Mobile Wallet should be fixed in the next days. 
In my opinion you could merge this PR.

keepresolve:
Please use version 7.1.2 for app mobile.

mauser-pmon:
> Please use version 7.1.2 for app mobile.

Cool thanks, will check it out as soon as its available in my Google Play store. (don't have the option to update from 7.1.1 for now)

mauser-pmon:
@keepresolve : Finally got the update on my Android Phone .. .singMessage works like a charm!

@jordansexton: Would be great if we could merge this PR!

jordansexton:
Published:
 - @solana/wallet-adapter-bitkeep => 0.3.5

PR #438, opened 2022-06-10
 Replace BitKeep Wallet logo

woibldd:
Upload picture again


jordansexton:
Thanks! We'll need to add a patch version bump for the Bitkeep adapter for this to get released

PR #440, opened 2022-06-14
Add spot support

Azkae:
This PR adds support for Spot wallet. (https://www.spot-wallet.com/)

How to test:
- Install chrome extension: https://www.spot-wallet.com/extension
- You will be asked to install Spot on iOS and link it to the extension.
- You can try the example in `packages/starter/example`

Thank you!


jordansexton:

packages/starter/example/components/ContextProvider.tsx
```diff
@@ -12,6 +12,7 @@ import {
     SolflareWalletAdapter,
     TorusWalletAdapter,
     WalletConnectWalletAdapter,
+    SpotWalletAdapter,
```
jordansexton, line 15: Please limit changes to adding the package only. We don't add all new wallets to the example or starter projects.

packages/wallets/spot/package.json
```diff
@@ -0,0 +1,33 @@
...
+    },
+    "scripts": {
+        "clean": "shx rm -rf lib/*",
+        "build": "yarn clean && tsc -p tsconfig.json && tsc-esm -p tsconfig.json && tsc -p tsconfig.cjs.json",
+        "postbuild": "echo '{\"type\":\"commonjs\"}' | npx json > lib/cjs/package.json && echo '{\"type\":\"module\"} ' | npx json > lib/esm/package.json"
+    },
+    "dependencies": {
+        "@solana/wallet-adapter-base": "^0.9.3",
+        "@solana/web3.js": "^1.36.0"
+    }
```
jordansexton, line 32: Can you update this to the peer/dev deps structure seen here? https://github.com/solana-labs/wallet-adapter/blob/e230958570f192786a27c57d5b39037b898e69dd/packages/wallets/solflare/package.json#L29-L38

packages/wallets/spot/src/adapter.ts
```diff
@@ -0,0 +1,246 @@
...
...71rLYTy/EAVU4KYO0IyxQvgLHT0gkVVNZNe6MdbFgex+gqm86zTn/tB9en8kSndHgCLlogFTxjHatRr1xymqElE62GDfSrHgKql6Bz6oNMNDqTgajkVQDZTOjPdBfLjVvbYvNrjgs+lo4dAAj/u3ecuvm5V136u4TZG3zHUJ2ImY+h0V0TQIymVofoPrw4DOK2CPhh6Rf4lcI1Ew9615nERONftFO6zrste2uFta4ymZyBEtalMgEwyU6WJtoBcgZM8iWoaPPbzKpjirLraWMtD0EleF5UdPd2sfGGdtV1v2vxjn9wM6zjpqmnx9VIuT/7px65UX8a+fXqxKsWbfdEBABg6Z2CiHGKT/C8VTD7A8BY+lVXJHmW++Sfy3yf/dThYyMe+KSMvJd4bFqPOnI8wDnpIMuPRNu+l+vkHbIFlP16uNTbkvrj9vJdCrwIwrJN4NVYoMUY66mj8AtBH0jWh1h2FL6Uw8xjia+EE+tCj3Ill4Om4lkd89/RTmy8efGv/v7KHb4SPyofTR8lu5b/Ld/y2NXLM+deqz+q+BqthVcxmAoAL+sDoOa5jsPCWXa6gQqA4lwRwAfY2OMPNha4CQg6CTCgEAB9v3YfJJ/82LupS1cfZ8Zj49gLWuzn2OCKx8ApW38lAKJtjwu+ZA0wY/J4AmjGCR3gzKesC2d3sAfPOyjGAMDT8L0+o6u0It/u13Ppb28nT7518WM39Q92QvDiXgezF6c4l37nO5ebP/euz79isbf79fr1zpfrLKLvG+5t44C4NHMECE6zg+Qwl5MWgVCy5PDjcGVgcwZiyzNSzyEEHz/F7O8860o/AyBsYU8Ay9EHB0C0j13LZrDyUzZ1nyJAj2CxjPpDcjAquOKmVvQfeh3+DDxyENEB8MH7a4MglhNs6Buky49IXo9xLfQkz+95/+Kn4vN8m3sKL0MXnoKVNao3fcvy5INn77pxubt4qb7te72+FHmNAuIajfvZOjVfoV94u0wBcbkcekqrh76fsNyUQ/UXTMgrAGK5DUfH7AwwJxDxkQPBgTTKVAAAfNrLWbo+AEImgmg1AHwTKwOwbNEPQKVvOLECIHANW3auXrDJqkHSuHc1xgsisKzp14eW/DD6o/qZwsckph8KWn5O+d2i363l6nbxPtEu37lVM52fnfqCp/8JEycDApvcoj0AAAAASUVORK5CYII=';
```
jordansexton, line 54: This looks pretty large, any chance you can provide an embedded SVG?

packages/wallets/spot/src/adapter.ts
```diff
@@ -0,0 +1,246 @@
...
+            // Spot doesn't handle partial signers, so if they are provided, don't use `signAndSendTransaction`
+            if (wallet && 'signAndSendTransaction' in wallet && !options?.signers) {
+                // HACK: Spot's `signAndSendTransaction` should always set these, but doesn't yet
+                transaction.feePayer = transaction.feePayer || this.publicKey || undefined;
+                transaction.recentBlockhash =
+                    transaction.recentBlockhash || (await connection.getRecentBlockhash('finalized')).blockhash;
+
+                const { signature } = await wallet.signAndSendTransaction(transaction, options);
+                return signature;
+            }
```
jordansexton, line 176: Please don't copy the Phantom adapter's workarounds. Partially signed transactions should also be supported.

packages/wallets/spot/src/adapter.ts
```diff
@@ -0,0 +1,246 @@
...
+
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const wallet = window!.spotSolWallet!;
+
+            try {
+                await wallet.connect();
+            } catch (error: any) {
+                if (error instanceof WalletError) {
+                    throw error;
+                }
```
jordansexton, line 113: This kind of check isn't necessary, because `wallet` would never throw `WalletError` (which is thrown by this library). The `instanceof` check would never match.

jordansexton (commented):

packages/wallets/spot/src/adapter.ts
Azkae: I can't use a SVG due to the gradient, but I have updated to a smaller png: 20kb -> 10kb

Azkae:
Thank you for your review! I have updated everything regarding your comments.

packages/wallets/spot/src/adapter.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    async sendTransaction(
+        transaction: Transaction,
+        connection: Connection,
+        options?: SendTransactionOptions
+    ): Promise<TransactionSignature> {
+        try {
+            const wallet = this._wallet;
+            if (wallet && 'signAndSendTransaction' in wallet) {
```
jordansexton, line 164: Is there ever a case where `'signAndSendTransaction' in wallet` would not be true? This check was a hack added for Phantom, because they had multiple versions of the wallet in production and the older ones didn't have this method. If you remove it, it should be possible to simplify this method, and you don't need to call the super method below.

packages/wallets/spot/src/adapter.ts
```diff
@@ -0,0 +1,237 @@
...
+    async signMessage(message: Uint8Array): Promise<Uint8Array> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const { signature } = await wallet.signMessage(message);
+                return signature;
+            } catch (error: any) {
+                throw new WalletSignTransactionError(error?.message, error);
```
jordansexton, line 217: I only caught this today in https://github.com/solana-labs/wallet-adapter/commit/112c54463a3d069e5b756dacf78e20a0711939ae but this should be `WalletSignMessageError`

packages/wallets/spot/src/adapter.ts
Azkae: For the sake of simplicity, I have removed it. I will add it back once we support it properly.

packages/wallets/spot/src/adapter.ts
Azkae: Good catch, fixed

jordansexton:
Thanks! This will be published today or tomorrow.

jordansexton:

Azkae:
Awesome, thanks!

PR #442, opened 2022-06-16
add particle auth wallet adapter

chaichuanfa:
website: https://particle.network/


jordansexton:

packages/wallets/particlenetwork/src/adapter.ts
```diff
@@ -0,0 +1,178 @@
...
+    async connect(): Promise<void> {
+        try {
+            if (this.connected || this.connecting) return;
+            if (this._readyState !== WalletReadyState.Loadable) throw new WalletNotReadyError();
+
+            this._connecting = true;
+
+            let wallet: ParticleNetwork;
+            try {
+                wallet = window.particle || new ParticleNetwork(this._config);
```
jordansexton, line 75: Since this requires instantiating `ParticleNetwork` through a loaded SDK, the loading needs to be done async in here. See how the Torus adapter does this: https://github.com/solana-labs/wallet-adapter/blob/e230958570f192786a27c57d5b39037b898e69dd/packages/wallets/torus/src/adapter.ts#L72-L86

packages/wallets/particlenetwork/src/adapter.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    async signTransaction(transaction: Transaction): Promise<Transaction> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const tx = await wallet.getSolanaProvider().signTransaction(bs58.encode(transaction.serialize()));
+                return Transaction.from(Buffer.from(tx, 'base64'));
```
jordansexton, line 133: This looks pretty strange -- you're encoding the serialized tx as base58, but deserializing it from base64. Using base64 everywhere (except for string signature values) will be faster. Also, please don't use `Buffer`, which is a Node.js class. Instead, refactor to use `Uint8Array`.

packages/wallets/particlenetwork/src/adapter.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    async signMessage(message: Uint8Array): Promise<Uint8Array> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const signature = await wallet.getSolanaProvider().signMessage(bs58.encode(message));
+                return Buffer.from(signature, 'base64');
```
jordansexton, line 169: Ditto

chaichuanfa:
@jordansexton  Thanks for reviewed.

 Fixed strange code, and test with starter demo. please check and commit～！Thanks

packages/starter/material-ui-starter/src/App.tsx
```diff
@@ -8,6 +8,7 @@ import {
     SlopeWalletAdapter,
     SolflareWalletAdapter,
     TorusWalletAdapter,
+    ParticleAdapter,
```
jordansexton, line 11: Please don't change any of the starter projects.

packages/wallets/particle/package.json
```diff
@@ -0,0 +1,40 @@
+{
+    "name": "@solana/wallet-adapter-particle",
+    "version": "0.1.1",
```
jordansexton, line 3: Start with `0.1.0` please!

packages/wallets/particle/src/adapter.ts
```diff
@@ -0,0 +1,179 @@
...
+    WalletName,
+    WalletNotConnectedError,
+    WalletNotReadyError,
+    WalletPublicKeyError,
+    WalletReadyState,
+    WalletSignMessageError,
+    WalletSignTransactionError,
+} from '@solana/wallet-adapter-base';
+import { PublicKey, Transaction } from '@solana/web3.js';
+import { SolanaWallet } from '@particle-network/solana-wallet';
```
jordansexton, line 17: You can't import this here except as types. All third party libraries must be loaded async.

packages/wallets/particle/src/adapter.ts
```diff
@@ -0,0 +1,179 @@
...
+            } catch (error: any) {
+                throw new WalletConnectionError(error?.message, error);
+            }
+
+            if (!wallet.publicKey()) throw new WalletAccountError();
+
+            let publicKey: PublicKey;
+            try {
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                publicKey = wallet.publicKey()!;
```
jordansexton, line 99: Why would this throw an error, or be undefined? Also, we should copy the public key bytes and instantiate a `PublicKey` here.

packages/wallets/wallets/package.json
```diff
@@ -55,6 +55,7 @@
         "@solana/wallet-adapter-solong": "^0.9.4",
         "@solana/wallet-adapter-spot": "^0.1.1",
         "@solana/wallet-adapter-tokenpocket": "^0.4.5",
-        "@solana/wallet-adapter-torus": "^0.11.12"
+        "@solana/wallet-adapter-torus": "^0.11.12",
+        "@solana/wallet-adapter-particle": "^0.1.1"
```
jordansexton, line 59: These should remain alphabetical (and start with `0.1.0`)

packages/wallets/wallets/src/index.ts
```diff
@@ -26,3 +26,4 @@ export * from '@solana/wallet-adapter-solong';
 export * from '@solana/wallet-adapter-spot';
 export * from '@solana/wallet-adapter-tokenpocket';
 export * from '@solana/wallet-adapter-torus';
+export * from '@solana/wallet-adapter-particle';
```
jordansexton, line 29: Same here, alphabetical please.

jordansexton (commented):
Looks good, biggest change needed is that any libs need to be loaded async. You can `Promise.all` them, or even better, consolidate them into one library.

chaichuanfa:
@jordansexton Thanks for reviewed.

Fixed the problem in the comments, please check and commit～！Thanks

packages/wallets/particle/src/adapter.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    private _connecting: boolean;
+    private _wallet: SolanaWallet | null;
+    private _publicKey: PublicKey | null;
+    private _config: ParticleAdapterConfig;
+    private _readyState: WalletReadyState =
+        typeof window === 'undefined' ? WalletReadyState.Unsupported : WalletReadyState.Loadable;
+
+    constructor(config: ParticleAdapterConfig) {
```
jordansexton, line 43: This can be optional, e.g. https://github.com/solana-labs/wallet-adapter/blob/4d95a46f7cabdf423123d029db55667923df972e/packages/wallets/phantom/src/adapter.ts#L68

packages/wallets/particle/src/adapter.ts
```diff
@@ -0,0 +1,185 @@
...
+        super();
+        this._connecting = false;
+        this._publicKey = null;
+        this._wallet = null;
+        this._config = config;
```
jordansexton, line 48: No need  to assign this, or have the private property, since you're not using the config object at this time. The existence of a config param is to allow for future compatibility in case you add options later.

packages/wallets/particle/src/adapter.ts
```diff
@@ -0,0 +1,185 @@
...
+
+            let ParticleClass: typeof ParticleNetwork;
+            let WalletClass: typeof SolanaWallet;
+            try {
+                const classes = await Promise.all([
+                    import('@particle-network/auth'),
+                    import('@particle-network/solana-wallet'),
+                ]);
+                ParticleClass = classes[0].ParticleNetwork;
+                WalletClass = classes[1].SolanaWallet;
```
jordansexton, line 82:
```suggestion
                const [{ ParticleClass: ParticleNetwork }, { WalletClass: SolanaWallet }] = await Promise.all([
                    import('@particle-network/auth'),
                    import('@particle-network/solana-wallet'),
                ]);
```

packages/wallets/particle/src/adapter.ts
```diff
@@ -0,0 +1,185 @@
...
+            } catch (error: any) {
+                throw new WalletLoadError(error?.message, error);
+            }
+
+            let wallet: SolanaWallet;
+            try {
+                wallet = window.particle || new WalletClass(new ParticleClass(this._config).auth);
```
jordansexton, line 89: Looking at how this is used, why is it necessary to load two separate SDKs? This will lead to two network calls for files that have to load, and they could be combined because you're just passing one into the other.

Also, what would cause `window.particle` to be set, do you also have a browser extension? And if so, when it's available, can you use it directly without loading an SDK?

jordansexton (commented):
Thanks! Mostly looks good, but I have some questions about the SDKs being loaded and configured, this seems like it can be more efficient.

chaichuanfa:
@jordansexton Thanks for reviewed.

Fixed the problem you raised , please check～！Thanks

jordansexton:
Thanks! This should be published tomorrow or Friday.

jordansexton:
Hmm, there's an issue when building the repo now --

```
../../../node_modules/@particle-network/solana-wallet/node_modules/@solana/web3.js/lib/index.d.ts(2,1): error TS6200: Definitions of the following identifiers conflict with those in another file:
```

`@particle-network/solana-wallet` is including some bad deps:
```json
  "dependencies": {
    "bs58": "^5.0.0",
    "@particle-network/auth": "^0.4.16",
    "@solana/web3.js": "^1.43.3"
  },
  "devDependencies": {
    "@types/bs58": "^4.0.1"
  },
```

`@solana/web3.js` is resolving to its own version of web3.js. This should probably be declared as a peerDependency and devDependency rather than a dependency so we don't end up with two versions of it, which causes build errors.

Additonally, `bs58` is set to `^5.0.0`, while the types are `^4.0.1`, and web3.js uses `"bs58": "^4.0.1"`.

PR #445, opened 2022-06-21
Add Brave Wallet

darkdh:
Resolve https://github.com/brave/brave-browser/issues/19468

Brave has both `solana.isPhantom` and `solana.isBraveWallet` set to true for maximum compatibility
Screen Shot 2022-06-21 at 14 27 46

Brave Wallet for Solana dApp support is only available in [Nightly](https://brave.com/download-nightly/)


packages/wallets/bravewallet/package.json
```diff
@@ -0,0 +1,38 @@
+{
+    "name": "@solana/wallet-adapter-bravewallet",
+    "version": "0.9.3",
```
jordansexton, line 3: Start with 0.1.0 please!

packages/wallets/bravewallet/package.json
```diff
@@ -0,0 +1,38 @@
+{
+    "name": "@solana/wallet-adapter-bravewallet",
```
jordansexton, line 2: Can this just be `@solana/wallet-adapter-brave`?

packages/wallets/bravewallet/src/adapter.ts
```diff
@@ -0,0 +1,268 @@
...
+        options?: SendOptions
+    ): Promise<{ signature: TransactionSignature }>;
+    signMessage(message: Uint8Array): Promise<{ signature: Uint8Array }>;
+    connect(): Promise<void>;
+    disconnect(): Promise<void>;
+    _handleDisconnect(...args: unknown[]): unknown;
+}
+
+interface BraveWindow extends Window {
+    solana?: BraveWallet;
```
jordansexton, line 43: Using `window.solana` isn't permitted for new adapters, and we are working with the few wallets that do use it to migrate all adapters away from it because conflicts over the variable make detection of available wallets impossible. Please consider changing your extension to `window.brave` or similar so it doesn't conflict with others!

packages/wallets/bravewallet/src/adapter.ts
```diff
@@ -0,0 +1,268 @@
...
+                            wallet.off('connect', connect);
+                            reject(reason);
+                        });
+                    });
+                } catch (error: any) {
+                    if (error instanceof WalletError) throw error;
+                    throw new WalletConnectionError(error?.message, error);
+                } finally {
+                    wallet._handleDisconnect = handleDisconnect;
+                }
```
jordansexton, line 138: It looks like this was all copied from the Phantom adapter, which has hacks to access private APIs because of bad connection behavior. Please don't copy these things!

packages/wallets/bravewallet/src/adapter.ts
```diff
@@ -0,0 +1,268 @@
...
+        options?: SendTransactionOptions
+    ): Promise<TransactionSignature> {
+        try {
+            const wallet = this._wallet;
+            // Brave doesn't handle partial signers, so if they are provided, don't use `signAndSendTransaction`
+            if (wallet && 'signAndSendTransaction' in wallet && !options?.signers) {
+                // HACK: Brave's `signAndSendTransaction` should always set these, but doesn't yet
+                transaction.feePayer = transaction.feePayer || this.publicKey || undefined;
+                transaction.recentBlockhash =
+                    transaction.recentBlockhash || (await connection.getRecentBlockhash('finalized')).blockhash;
```
jordansexton, line 194: Same here, please don't copy the Phantom adapter's workarounds. Partially signed transactions should also be supported.

packages/wallets/bravewallet/src/adapter.ts
```diff
@@ -0,0 +1,268 @@
...
+} from '@solana/wallet-adapter-base';
+import { Connection, PublicKey, SendOptions, Transaction, TransactionSignature } from '@solana/web3.js';
+
+interface BraveWalletEvents {
+    connect(...args: unknown[]): unknown;
+    disconnect(...args: unknown[]): unknown;
+}
+
+interface BraveWallet extends EventEmitter<BraveWalletEvents> {
+    isBraveWallet?: boolean;
```
jordansexton, line 27: Probably don't need this check either when not using `window.solana`.

jordansexton (commented):
Thanks for this! Please don't copy the Phantom adapter's behavior though.

I also want to mention that we're working on an actual wallet standard, which will be introduced soon. More context on that [here](https://twitter.com/jordaaash/status/1534987011781804039) and [here](https://github.com/solana-labs/wallet-adapter/pull/384#issuecomment-1159115764).

When that's ready for review, I'd love to get your feedback and work on a migration plan.

packages/wallets/bravewallet/package.json
darkdh: fixed and squashed.

packages/wallets/bravewallet/src/adapter.ts
darkdh: fixed and squashed.

packages/wallets/brave/src/adapter.ts
```diff
@@ -0,0 +1,260 @@
...
+        this.emit('disconnect');
+    }
+
+    async sendTransaction(
+        transaction: Transaction,
+        connection: Connection,
+        options?: SendTransactionOptions
+    ): Promise<TransactionSignature> {
+        try {
+            const wallet = this._wallet;
```
jordansexton, line 188: Make sure to check for `if (!wallet) throw new WalletNotConnectedError();` as seen below, this will throw on the next line otherwise.

darkdh: fixed and also added throwing `WalletSendTransactionError`

packages/wallets/bravewallet/src/adapter.ts
darkdh: removed that hack since Brave event emitting doesn't get affected by popup

packages/wallets/bravewallet/src/adapter.ts
yrliou: @jordansexton Is there a documentation we could reference for how partially signed transactions should be supported? What's the expect behavior for the wallet, like what should we do with options.signers? As I understand for cases like signTransaction and signAllTransactions, some signatures can be provided via transaction object's signatures property, I assume it would be the same when using signAndSendTransaction, and I'm not sure what should we do with options.signers.

packages/wallets/bravewallet/src/adapter.ts
bbondy: @jordansexton we're purposely exposing window.solana so that there is Phantom wallet compatibility for sites that don't use this adapter or don't update it, but we'll recommend that people use this adapter that we're trying to land here.

Note that we have a setting for controlling if we expose our provider or not. And by default if the user has Phantom it will use Phantom.

jordansexton: I understand, and I can't tell you to not write to `window.solana`, but we're not accepting new adapters that use `window.solana`, and we're actively migrating away from this.

We've told this to all other wallets for almost a year, and there are a few good reasons for this restriction, so we must remain consistent about this.

Once the wallet standard provides a location for wallets to attach to, any supporting app will detect Brave Wallet without needing to make a change to support it specifically.

packages/wallets/bravewallet/src/adapter.ts
jordansexton: Hmm, specific documentation might be hard to come by, but here's how we implement it in the base class:

https://github.com/solana-labs/wallet-adapter/blob/244af02cb3dcf9334abf61cb9a11e497bde86d57/packages/core/base/src/signer.ts#L25-L33

The reason we have this is to support program / multisig wallets that need to accept meta-transactions, create a new transaction, then sign it with the partial signers provided.

I think in Brave's case this doesn't matter, so you can just accept the array of optional signers, and if provided, sign with them along with the user.

packages/wallets/brave/README.md
```diff
@@ -0,0 +1,5 @@
+# `@solana/wallet-adapter-bravewallet`
```
jordansexton, line 1:
```suggestion
# `@solana/wallet-adapter-brave`
```

darkdh: fixed in https://github.com/solana-labs/wallet-adapter/pull/445/commits/db4c9ed630324c2db52183bebd76b75692bbd016

packages/wallets/bravewallet/src/adapter.ts
darkdh: fixed in https://github.com/solana-labs/wallet-adapter/pull/445/commits/db4c9ed630324c2db52183bebd76b75692bbd016

packages/wallets/brave/src/adapter.ts
```diff
@@ -0,0 +1,250 @@
...
+    async signMessage(message: Uint8Array): Promise<Uint8Array> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const { signature } = await wallet.signMessage(message);
+                return signature;
+            } catch (error: any) {
+                throw new WalletSignTransactionError(error?.message, error);
```
jordansexton, line 230: I just caught this in several other adapters today, but this should be `WalletSignMessageError`

https://github.com/solana-labs/wallet-adapter/commit/112c54463a3d069e5b756dacf78e20a0711939ae fixes this elsewhere

jordansexton (commented):
LGTM, last issue is something I just discovered -- `signMessage` throws the wrong error type.

packages/wallets/brave/src/adapter.ts
darkdh: good call, fixed in https://github.com/solana-labs/wallet-adapter/pull/445/commits/b2b4e713bbe75bc8008a39829def9ce751745fc3

jordansexton:

packages/wallets/brave/src/adapter.ts
```diff
@@ -0,0 +1,251 @@
...
+        return await super.sendTransaction(transaction, connection, options);
+    }
+
+    async signTransaction(transaction: Transaction): Promise<Transaction> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                return (await wallet.signTransaction(transaction)) || transaction;
```
steveluscher, line 196: When would you want to return the original transaction? Is this in cases where it's _already_ signed?

steveluscher: Cool! If Brave's `wallet.signTransaction` can never return `null`, you should uninherit this part :)

darkdh: addressed [here](https://github.com/solana-labs/wallet-adapter/pull/480)

PR #471, opened 2022-07-18
fix: `children` overwriting account address

### LINKED ISSUE #470: Prevent `children` from overwriting account address
## Description
There's several issues with passing `children` as an argument for all of these components :)

It's most apparent in this case below, where `children` could simply be text

https://github.com/solana-labs/wallet-adapter/blob/ee1816196bb49f761f43a81740ac2fb6f4a2d674/packages/ui/material-ui/src/WalletMultiButton.tsx#L56-L60

If `children` passed is the text that goes inside of the button, for example:

Screen Shot 2022-07-17 at 7 34 20 PM

then, this will overwrite the account address that is displayed. Another example to illustrate this is below:
Screen Shot 2022-07-17 at 7 34 47 PM

## Desired solution
We should be able to change both the button text displayed when a wallet is NOT connected _and_ when the wallet IS connected

This can easily be solved by avoiding the use of `children` as an argument and instead being VERY specific with the potential arguments that can be passed

Alternatively, you could also define the type for `children` in `Button.tsx` so that devs know what `children` must NOT be:
https://github.com/solana-labs/wallet-adapter/blob/f6787bbb8b86b401be41dfdca1386e4b24c66340/packages/ui/react-ui/src/Button.tsx#L3-L27

Notice that `children` is not defined in the `ButtonProps` type in the code snippet above

platocrat:
Fixes #470 

## Update
I offer two potential solutions in this PR:
1. Commit https://github.com/solana-labs/wallet-adapter/pull/471/commits/ecad40d5e55e44f0fdea6cba0420a8a2bac706aa: simple change that diplays `address` if detected, and if not, displays `children` (defined as `content` instead
2. Commits https://github.com/solana-labs/wallet-adapter/pull/471/commits/13c541748d44a102a2ee690440f04405e7a7a4ee through https://github.com/solana-labs/wallet-adapter/pull/471/commits/36ba33d9c4f74e8022b4a90f5ff7c0beea51e1e3: defines two 2 new _optional_ properties on `Button.tsx`: `buttonText` and `account`, which are both used in `WalletMultiButton.tsx`

## Note
Idk how to build this project locally to run a test that this fix is bug free

### DIFF
--- a/packages/ui/react-ui/src/Button.tsx
+++ b/packages/ui/react-ui/src/Button.tsx
@@ -1,7 +1,6 @@
 import React, { ReactNode, CSSProperties, FC, MouseEvent, PropsWithChildren, ReactElement } from 'react';
 
 export type ButtonProps = PropsWithChildren<{
-    children?: ReactNode;
     buttonText?: string;
     address?: string;
     className?: string;
--- a/packages/ui/react-ui/src/Button.tsx
+++ b/packages/ui/react-ui/src/Button.tsx
@@ -1,6 +1,9 @@
-import React, { CSSProperties, FC, MouseEvent, PropsWithChildren, ReactElement } from 'react';
+import React, { ReactNode, CSSProperties, FC, MouseEvent, PropsWithChildren, ReactElement } from 'react';
 
 export type ButtonProps = PropsWithChildren<{
+    children?: ReactNode;
+    buttonText?: string;
+    address?: string;
     className?: string;
     disabled?: boolean;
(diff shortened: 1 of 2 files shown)


platocrat:
after asking a question abt this on Solana's Stack Exchange, ig the intent of using `children` is so to keep `react-ui` minimal while letting the user be able to add their own customizations within the component

so, i can configure the button text and account address myself to be displayed when i want without overcomplicating the `react-ui` package

PR #472, opened 2022-07-18
Phantom adapter uses window.phantom.solana 

Tataklim:
Hi! 
We have noticed that there are some of wallets that try to overwrite window.solana.isPhantom. That’s why sometimes we can meet quite unexpectable behavior, when Phantom couldn't be loaded with another wallets' extensions. 
Nowadays Phantom migrated to window.phantom.solana, so it's possible to [avoid](https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/phantom/src/adapter.ts#L48) these errors for everyone who uses wallet-adapter

### DIFF
--- a/packages/wallets/phantom/src/adapter.ts
+++ b/packages/wallets/phantom/src/adapter.ts
@@ -46,7 +46,9 @@ interface PhantomWindow extends Window {
     // Multiple wallet adapters cannot be detected properly if they all try to write to the same window global.
     // All wallets that currently do this have committed to migrating away from using `window.solana`.
     // This must be changed to `window.yourWalletName` in your adapter, and must not use `window.solana`.
-    solana?: PhantomWallet;
+    phantom?:{
+        solana?: PhantomWallet
+    };
 }
 
 declare const window: PhantomWindow;
@@ -77,7 +79,7 @@ export class PhantomWalletAdapter extends BaseMessageSignerWalletAdapter {
 
         if (this._readyState !== WalletReadyState.Unsupported) {
             scopePollingDetectionStrategy(() => {
-                if (window.solana?.isPhantom) {
+                if (window.phantom?.solana?.isPhantom) {
                     this._readyState = WalletReadyState.Installed;
                     this.emit('readyStateChange', this._readyState);
                     return true;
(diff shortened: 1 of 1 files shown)


jordansexton:
Thanks for this! I merged it, but it's worth noting that older versions of Phantom that users may have installed and haven't updated could be using `window.solana` still.

It might be good to add a fallback to that in the adapter for a while so we don't break dapps that upgrade the adapter package for these users.

Tataklim:
@jordansexton yeah, that's a good point. But in this case any other wallet could set window.solana.isPhantom true and mimic that the phantom code is loaded. And when you click on the connection of the phantom wallet, another wallet will open

bfriel:
Thanks for the PR @Tataklim ! 

@jordansexton , we have been injecting `window.phantom` alongside `window.solana` for over 10 months now. After looking at our analytics I believe this is safe to merge.

Tataklim:
@jordansexton Hi! When is the new release planned?

jordansexton:
@Tataklim it was planned for yesterday, but should happen today along with a few other package releases. Thanks for following up!

biw:
Hey @Tataklim & @bfriel, not sure if this is the right place, so please let me know if there's a better venue to bring this up!

It looks like `window.phantom.solana` and `window.solana` reference the same object.

Screen Shot 2022-07-26 at 4 39 18 PM

Because of that, I believe there would still be the same issues with other wallets (who haven't migrated to their own namespace yet) sharing the object. (For example, the [safepal extension](https://chrome.google.com/webstore/detail/safepal-extension-wallet/lgmpcpglpngdoalbgeoldeajfclnhafa) tries to delete the `window.solana` object)


Is the plan on the Phantom side to have these be different objects once these changes are released?

jordansexton:
> Is the plan on the Phantom side to have these be different objects once these changes are released?

Not likely. They'd need to be the same object (or at least, have the same API) for compatibility with dapps (and versions of the Phantom wallet adapter) that have historically used `window.solana`.

Extensions can't stop each other from racing to write `window.solana` or any other window property, so this issue will persist until an [alternative means of wallet discovery](https://github.com/solana-labs/wallet-standard/tree/master/packages/core/standard#present-a-predictable-interface-for-wallets-to-attach-to-the-window) is available.

Tataklim:
Hi @jordansexton ! New version of `wallet-adapter-wallets` doesn't include bumped version of `wallet-adapter-phantom`. Could you please up it?)

jordansexton:
@Tataklim it does, we just don't push new versions of `wallets` with every adapter release. All dependency versions are set with `^`. Fresh installs will get them automatically, others just need to `yarn upgrade` / `npm ugprade`.

jordansexton:
The `wallets` package has been updated since there were several adapter releases. Anyone who already uses it will still have to `yarn upgrade` to see the changes though.

Tataklim:
@jordansexton Hi! Could you help me please - why was it necessary to [add](https://github.com/solana-labs/wallet-adapter/commit/5a274e0a32c55d4376d63a802f0d512947b087af) window.solana fallback?

PR #480, opened 2022-07-21
Brave wallet update

darkdh:
1. Address [comment](https://github.com/solana-labs/wallet-adapter/pull/445#discussion_r926251941)
2. Update example to include BraveWalletAdapter


jordansexton:
Wallet Adapter was intended to be neutral and not provide recommendations of any wallet, or even implicit guarantees that they work properly. It hasn't always done that.

https://github.com/solana-labs/wallet-adapter/pull/498 has removed all adapters from the examples and starter projects, and we will solve this problem with better docs and support rather than appearing to promote or prioritize any wallet.

PR #481, opened 2022-07-23
init backpack 

armaniferrante:


armaniferrante:
Need to rebase master.

packages/ui/react-ui/src/WalletModal.tsx
```diff
@@ -43,6 +43,7 @@ export const WalletModal: FC<WalletModalProps> = ({ className = '', container =
             ? installedWallets[0]
             : wallets.find((wallet: { adapter: { name: WalletName } }) => wallet.adapter.name === 'Torus') ||
                   wallets.find((wallet: { adapter: { name: WalletName } }) => wallet.adapter.name === 'Phantom') ||
+                  wallets.find((wallet: { adapter: { name: WalletName } }) => wallet.adapter.name === 'Backpack') ||
```
armaniferrante, line 46: Is this needed?

armaniferrante:
This is ready for review.

packages/wallets/backpack/package.json
```diff
@@ -0,0 +1,33 @@
...
+    },
+    "scripts": {
+        "clean": "shx rm -rf lib/*",
+        "build": "yarn clean && tsc -p tsconfig.json && tsc-esm -p tsconfig.json && tsc -p tsconfig.cjs.json",
+        "postbuild": "echo '{\"type\":\"commonjs\"}' | npx json > lib/cjs/package.json && echo '{\"type\":\"module\"} ' | npx json > lib/esm/package.json"
+    },
+    "dependencies": {
+        "@solana/wallet-adapter-base": "^0.9.4",
+        "@solana/web3.js": "^1.36.0"
+    }
```
jordansexton, line 32:
```suggestion
    "peerDependencies": {
        "@solana/web3.js": "^1.44.3"
    },
    "dependencies": {
        "@solana/wallet-adapter-base": "^0.9.9"
    },
    "devDependencies": {
        "@solana/web3.js": "^1.44.3"
    }
```

For consistency with the other adapters.

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,259 @@
...
+    isConnected: boolean;
+    signTransaction(transaction: Transaction): Promise<Transaction>;
+    signAllTransactions(transactions: Transaction[]): Promise<Transaction[]>;
+    send(
+        transaction: Transaction,
+        signers?: Signer[],
+        options?: SendOptions,
+        connection?: Connection
+    ): Promise<TransactionSignature>;
+    signMessage(message: Uint8Array): Promise<Uint8Array | null>;
```
jordansexton, line 39: What does it mean for this to return `null`?

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,259 @@
...
+                        wallet.connect().catch(() => {
+                            wallet.off('connect', handleConnect);
+                            reject(new WalletWindowClosedError());
+                        });
+                    });
+                } catch (error: any) {
+                    if (error instanceof WalletError) throw error;
+                    throw new WalletConnectionError(error?.message, error);
+                }
+            }
```
jordansexton, line 131: This code looks like it comes from Phantom's adapter, which now has a warning: https://github.com/solana-labs/wallet-adapter/blob/44a7d3b4910e2ae175ddfef6799a168999fd6ec3/packages/wallets/phantom/src/adapter.ts#L119-L123

If Backpack correctly fulfills the Promise, it can probably be substantially simplified, e.g. https://github.com/solana-labs/wallet-adapter/blob/44a7d3b4910e2ae175ddfef6799a168999fd6ec3/packages/wallets/glow/src/adapter.ts#L130-L134

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,259 @@
...
+            const resp = await wallet.send(transaction, signers, sendOptions, connection);
+            if (!resp) {
+                throw new WalletWindowClosedError();
+            }
+            return resp;
+        } catch (error: any) {
+            this.emit('error', error);
+            throw error;
+        }
+    }
```
jordansexton, line 193: The general structure of this method has been updated elsewhere, e.g. https://github.com/solana-labs/wallet-adapter/blob/44a7d3b4910e2ae175ddfef6799a168999fd6ec3/packages/wallets/phantom/src/adapter.ts#L194-L219

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,259 @@
...
+    async signMessage(message: Uint8Array): Promise<Uint8Array> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                const signature = await wallet.signMessage(message);
+                if (!signature) {
+                    throw new Error('User denied signature request');
+                }
```
jordansexton, line 236: Ah, I see why this could be `null`. But it's probably generally better if the wallet throws an error than just returning nothing, right?

jordansexton (commented):
Thanks for this, I'm very excited to see more support for Backpack!

packages/ui/react-ui/src/WalletModal.tsx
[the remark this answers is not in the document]
armaniferrante: Sounds good!

armaniferrante, line 46: ```suggestion```

armaniferrante, line 46:
```suggestion
```

packages/wallets/backpack/src/adapter.ts
armaniferrante: This promise is correctly filled. @tomlinton to confirm.

packages/wallets/backpack/src/adapter.ts
armaniferrante: Is the main thing here that we need to call `prepareTransaction`?

armaniferrante: It appears that the purpose of `prepareTransaction` https://github.com/solana-labs/wallet-adapter/blob/master/packages/core/base/src/adapter.ts#L91-L99 is to add a fee payer and the recent blockhash. Backpack's `send` implementation does this,  so that we can generate anchor clients without this library.

packages/wallets/backpack/src/adapter.ts
armaniferrante: It returns null when the user denies the request.  My thinking was to not throw an error for a desired return value.

packages/wallets/backpack/src/adapter.ts
armaniferrante: Moving this comment here https://github.com/solana-labs/wallet-adapter/pull/481#discussion_r928173263

packages/wallets/backpack/src/adapter.ts
armaniferrante: I'm not sure what the idiomatic thing is in javascript, but in rust, this would be an enum return value not an error.

armaniferrante: Note that we do this for send, and signTx as well.

jordansexton: Hmm, it's kind of a judgment call, and ultimately we can adapt to whatever Backpack's behavior is. But I'd say when you have something that returns a Promise, you usually expect it to resolve with the thing you want, or fail. The user declining is a likely expected failure mode, so every downstream caller just has to check for null anyway, and also check for errors.

packages/wallets/backpack/src/adapter.ts
jordansexton: Hmm, the most critical thing is that the method signature matches (connection is required, default param isn't but then you don't need the `options || {}`). The method implementation is discretionary though.

And how do you want to handle partial signers? In general I think it makes sense to sign before passing to the wallet, rather than pass signers to the wallet (since the wallet should be able to handle a transaction arriving with partial signatures anyway). And if the partial signatures should be done in the app, the tx will have to have a feepayer/blockhash.

packages/wallets/backpack/src/adapter.ts
jordansexton: Cool. Yeah, as long as it throws when there's an issue (e.g. if a user declines the connection) then the events shouldn't be needed (here, anyway).

packages/wallets/backpack/src/adapter.ts
jordansexton: FWIW, I view the `connect` behavior as the same. If the user declines the connection, returning `null` or any value doesn't seem more logical than throwing.

packages/wallets/backpack/src/adapter.ts
armaniferrante: If I understand correctly, I think both use cases are currently covered. The signers here is a convenience to eliminate the `partialSign` boilerplate (and it conforms to the Anchor interface), allowing the user to additionally (and optionally) pass in those signer keys. If you want to partial sign inside the app as you suggest, you can still do so by not providing empty signers.

packages/wallets/backpack/src/adapter.ts
tomlinton: It's fulfilled correct in failure cases, but we'll clean up the success case and simplify to the example you linked

jordansexton: Ah, okay. Yeah, I guess in the failure case (like if you return instead of throwing if the user declines) we could just check here for the pubkey and throw after. Your call -- I don't want to dictate how Backpack works, just describing how most wallets handle this.

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,259 @@
...
+    publicKey?: { toBytes(): Uint8Array };
+    isConnected: boolean;
+    signTransaction(transaction: Transaction): Promise<Transaction>;
+    signAllTransactions(transactions: Transaction[]): Promise<Transaction[]>;
+    send(
+        transaction: Transaction,
+        signers?: Signer[],
+        options?: SendOptions,
+        connection?: Connection
+    ): Promise<TransactionSignature>;
```
jordansexton, line 38: FWIW this type doesn't declare that it can be null, same with sign and signAll above.

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+
+    async signTransaction(transaction: Transaction): Promise<Transaction> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                return (await wallet.signTransaction(transaction)) || transaction;
```
jordansexton, line 201: Do you return `null` if a user declines to sign? Because if so, the caller will end up with their original, unsigned tx object and no error.

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+
+    async signAllTransactions(transactions: Transaction[]): Promise<Transaction[]> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            try {
+                return (await wallet.signAllTransactions(transactions)) || transactions;
```
jordansexton, line 217: Same here, do you return null if declined?

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,259 @@
...
+        connection?: Connection,
+        options?: SendTransactionOptions
+    ): Promise<TransactionSignature> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            const { signers, ...sendOptions } = options ? options : { signers: undefined };
+
+            const resp = await wallet.send(transaction, signers, sendOptions, connection);
```
jordansexton, line 184: Just to clarify, is `resp` a signature or an object containing a signature field like many of the wallets do?

armaniferrante: It's a signature. I'd prefer to leave it that way due to the Anchor interface https://github.com/coral-xyz/anchor/blob/master/ts/src/provider.ts#L24-L28. 

And now that I see that, I realize we should change the `| null` return values in the connect, send, sign, etc apis to throw to match anchor.

packages/wallets/backpack/src/adapter.ts
armaniferrante, line 39:
```suggestion
    signMessage(message: Uint8Array): Promise<Uint8Array>;
```

packages/wallets/backpack/src/adapter.ts
armaniferrante: Changing to throw. See comments below.

packages/wallets/backpack/src/adapter.ts
armaniferrante: This will throw.

packages/wallets/backpack/src/adapter.ts
armaniferrante: Throwing.

packages/wallets/backpack/src/adapter.ts
armaniferrante: Changing to throw per comment here https://github.com/solana-labs/wallet-adapter/pull/481#discussion_r928177693

packages/wallets/backpack/src/adapter.ts
jordansexton: Ah, all good! I didn't mean to suggest changing it, was only concerned that maybe it was a response object rather than a sig from the varname.

packages/wallets/backpack/src/adapter.ts
armaniferrante: Nulls removed!

armaniferrante:
Ready for another round!

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,242 @@
...
+    async sendTransaction(
+        transaction: Transaction,
+        connection?: Connection,
+        options?: SendTransactionOptions
+    ): Promise<TransactionSignature> {
+        try {
+            const wallet = this._wallet;
+            if (!wallet) throw new WalletNotConnectedError();
+
+            const { signers, ...sendOptions } = options ? options : { signers: undefined };
```
jordansexton, line 168:
```suggestion
    async sendTransaction(
        transaction: Transaction,
        connection: Connection,
        options: SendTransactionOptions = {}
    ): Promise<TransactionSignature> {
        try {
            const wallet = this._wallet;
            if (!wallet) throw new WalletNotConnectedError();

            const { signers, ...sendOptions } = options;
```

`connection` is a required arg, adding default param consistent with other adapters

packages/wallets/backpack/src/adapter.ts
```diff
@@ -0,0 +1,242 @@
...
+
+            try {
+                return await wallet.send(transaction, signers, sendOptions, connection);
+            } catch (error: any) {
+                if (error instanceof WalletError) throw error;
```
jordansexton, line 173:
```suggestion
```

This check isn't necessary because error will never be a `WalletError`

jordansexton:
Thank you! Will ship tomorrow with some other packages pending release

PR #485, opened 2022-07-25
Added `@typescript-eslint/consistent-type-imports` rule to codebase

biw:
Looking at the core packages, the javascript code from `@solana/web3.js` isn't used except for in `ConnectionProvider`, only the typescript types. In order to make it a little easier to figure out dependency trees, I made a small edit to only import the types. The benefits of doing this are [outlined here](https://stackoverflow.com/questions/61412000/do-i-need-to-use-the-import-type-feature-of-typescript-3-8-if-all-of-my-import)

I also updated the Solong URL to be the current site listed on their twitter since https://solongwallet.com/ does not resolve


jordansexton:
Is there an eslint plugin/config we could use to detect imports used only as types, and ideally fix them automatically? My concern with this change is that it increases maintenance burden -- you have to know what imports are later used as types and what are not, and you have to do this consistently for it to be useful. Since IDEs by default tend to import things not as types, we're fighting a losing battle unless we automate it.

wentokay:
there's a [consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md) rule which will do this automatically with `eslint --fix`

this is what's happening with `export`s in this PR https://github.com/coral-xyz/anchor/pull/2086

however, there's a [better option that looks like it might get merged soon](https://github.com/typescript-eslint/typescript-eslint/pull/5050) in the typescript-eslint repo, which will allow inline type imports, so instead of having multiple imports from the same source

```typescript
import type { Foo } from "./baz";
import { bar } from "./baz";
```

it'll generate something like

```typescript
import { type Foo, bar } from "./baz";
```

biw:
Great find @wentokay! If it sounds good to @jordansexton, I can update this PR to include the [consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md?rgh-link-date=2022-07-26T16%3A26%3A33Z) rule and then once the better option update is merged in, I will make sure it lands correctly

jordansexton:
Sounds very good, thank you!

biw:
Just added `@typescript-eslint/consistent-type-imports`! It also is fix with `eslint --fix` so should be very minimal impact for maintainance moving forward 😄

jordansexton:
Thank you so much!

PR #491, opened 2022-07-29
chore: adding the ‘Question’ tag now autocloses issues and directs the poster to Stack Exchange

steveluscher:
#### Problem

We want to make it as easy as possible for core team members to nudge people over to Stack Exchange when questions that are not issues with Solana itself get posted to GitHub.

#### Summary of Changes

* Added a GitHub workflow to comment on and close issues when the `question` tag is added.

### DIFF
--- a/.github/label-actions.yml
+++ b/.github/label-actions.yml
@@ -1,0 +1,27 @@
+question:
+  issues:
+    # Post a comment, `{issue-author}` is an optional placeholder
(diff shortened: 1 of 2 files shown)


steveluscher:
cc/ @jordansexton; one-click Stack Exchange.

jordansexton:
omg based

PR #498, opened 2022-08-02
Encourage developers to hand-pick the selection of wallets most appropriate for their audience

steveluscher:
# Problem

As third-party wallet offerings evolve, our starter apps continue to put forth a starter list of wallets that does not. Besides going stale over time, the decision of which wallets a dApp should support is not a responsibility of `wallet-adapter` itself; ultimately, the decision of which wallets to support rests with the dApp publisher themselves.

# Summary of changes

* Create a `FakeWalletAdapter` that conforms to the wallet adapter interface, but otherwise offers no useful signing/sending services.
* Include `FakeWalletAdapter` in all of the starter packs, along with instructions on where to find real wallet adapters.

# Test plan

1. `yarn && yarn build`
2. `cd packages/starter/example && yarn dev`
3. Open https://localhost:3000

https://user-images.githubusercontent.com/13243/182485005-84a20666-7fe3-471a-afa7-d8dedfc5d44a.mov

1. `yarn && yarn build`
2. `cd packages/starter/react-ui-starter && yarn start`
3. Open https://localhost:1234

https://user-images.githubusercontent.com/13243/182486068-c0b2c255-5810-40e4-b001-495cd90e014f.mov

Observe warning on the developer console that links to [this document](https://github.com/solana-labs/wallet-adapter#usage).

image

PR #507, opened 2022-08-08
Add support for WalletConnect v2

jnwng:
follow-up to #451 to add back `wallet-adapter-walletconnect`. the WalletConnect code hit `2.0.0-rc1` recently, which i hope will lead to some stability here!

here's a demo video: 

https://user-images.githubusercontent.com/1118534/183523644-1d86c631-23a9-4db4-a9a7-2f16a10096a9.mov

if you're building against this in a dApp, here's a working [WCv2-enabled Solana wallet](https://web-examples-git-jw-demo-solana-wallet-jnwng.vercel.app/walletconnect) UI (devnet), and a demo dApp with [wallet-adapter-walletconnect](http://poised-lumber.surge.sh/)

this is a draft because i could use some feedback on:

the WalletConnect client has some functionality to remember previous "pairings"; this means folks don't need to scan their QRCode again until the session is manually disconnected from the dApp. surfacing this functionality in the wallet-adapter requires introduce (yet more) custom UI code that doesn't quite make sense to surface here. i know we already use `@walletconnect/qrcode-modal` but at least that's already packaged. i'm happy to ship this without the pairing UI and have follow-up code that handles the pairing part. high-level thoughts? should i yoink this into a separate repository altogether?

i already plan to fast-follow to make this a [wallet-standard](https://github.com/wallet-standard/wallet-standard)-compatible wallet, but want this in this repository first to make it easier for folks to begin integration work with WCv2

**before **landing...****
- [x] Remove changes to `components/ContextProvider`
- [x] Verify the `yarn.lock` changes


packages/starter/example/components/ContextProvider.tsx
```diff
@@ -59,15 +59,20 @@ const WalletContextProvider: FC<{ children: ReactNode }> = ({ children }) => {
...
-            /**
-             * Select the wallets you wish to support, by instantiating wallet adapters here.
-             *
-             * Common adapters can be found in the npm package `@solana/wallet-adapter-wallets`.
-             * That package supports tree shaking and lazy loading -- only the wallets you import
-             * will be compiled into your application, and only the dependencies of wallets that
-             * your users connect to will be loaded.
-             */
-            new FakeWalletAdapter(),
+            new WalletConnectWalletAdapter({
```
jnwng, line 62: i'll remove this and put it into a README.md

jordansexton: Cool! Yeah, I'd say just move the example stuff to the `packages/wallets/walletconnect` README probably. https://github.com/solana-labs/wallet-adapter/pull/498 removed all preference for wallets, but it would be fine (good, even!) for individual adapter readmes to show how to use them, especially since this one requires config.

packages/wallets/walletconnect/src/adapter.ts
```diff
@@ -33,9 +33,30 @@ export enum WalletConnectRPCMethod {
...
+    pairingTopic,
+});
+
+const getChainId = (network: WalletAdapterNetwork): WalletConnectChainID => {
+    switch (network) {
+        case WalletAdapterNetwork.Mainnet:
+            return WalletConnectChainID.Mainnet;
+        case WalletAdapterNetwork.Devnet:
+        default:
+            return WalletConnectChainID.Devnet;
```
jordansexton, line 53: In general the adapters default to mainnet. But I'm wondering if we want to silently just accept testnet or something if it will actually not work. Seems better to fail fast?

packages/wallets/walletconnect/src/adapter.ts
```diff
@@ -48,25 +69,20 @@ export class WalletConnectWalletAdapter extends BaseSignerWalletAdapter {
...
     private _readyState: WalletReadyState =
         typeof window === 'undefined' ? WalletReadyState.Unsupported : WalletReadyState.Loadable;
+    private _network: WalletAdapterNetwork;
 
     constructor(config: WalletConnectWalletAdapterConfig) {
         super();
 
         this._publicKey = null;
         this._connecting = false;
+        this._network = config.network || WalletAdapterNetwork.Devnet;
```
jordansexton, line 84: `config` is required and `config.network` also required so probably not needed to default?

packages/wallets/walletconnect/src/adapter.ts
```diff
@@ -16,12 +17,11 @@ import {
 } from '@solana/wallet-adapter-base';
 import type { Transaction } from '@solana/web3.js';
 import { PublicKey } from '@solana/web3.js';
-import WalletConnectClient, { CLIENT_EVENTS } from '@walletconnect/client';
+import WalletConnectClient from '@walletconnect/sign-client';
```
jordansexton, line 20: This adapter should be refactored to be a `Loadable` adapter -- the [Strike multisig](https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/strike/src/adapter.ts#L14-L38) or [Torus adapters](https://github.com/solana-labs/wallet-adapter/blob/7612489924877c1cf5ff034c8eadb5d1311573ca/packages/wallets/torus/src/adapter.ts#L78-L90) are examples of this. Imports of types only, dynamic imports of actual SDKs. Ideally we can find a way to consume a single library that wraps any `@walletconnect/*` stuff needed so only one file load is required.

packages/wallets/walletconnect/src/adapter.ts
```diff
@@ -89,74 +105,47 @@ export class WalletConnectWalletAdapter extends BaseSignerWalletAdapter {
...
+                    getWalletConnectParams(getChainId(this._network), lastActivePairing)
+                );
+
+                if (uri) {
+                    QRCodeModal.open(uri, () => {
+                        throw new WalletWindowClosedError();
+                    });
+                }
+
+                session = await approval();
```
jordansexton, line 132: This is already much improved! Wondering if we can move more of this adapter's logic into the async loaded library. Check out some other Loadable adapters (e.g. Keystone, Ledger). We moved all complexity out of the adapter itself to keep the footprint and maintenance light.

jordansexton (commented):
This is great, thanks for carrying this forward. The only really substantive change here will be to reduce the footprint so that WalletConnect SDKs will only be loaded async if a user selects it to connect.

packages/wallets/walletconnect/src/adapter.ts
jnwng: done! check out https://github.com/jnwng/walletconnect-solana/pull/1 (using your fancy-shmancy typescript-monorepo) for all the new logic)

jnwng:
@jordansexton ready for another look... but don't look at this until next week :) enjoy your weekend!

PR #514, opened 2022-08-15
should not return after `wallet.connect()`

chochihim:
The `adapter.connect()` should just keep running and emit the `connect` event after calling `wallet.connect()`


jordansexton:
Ah, good catch, thanks. I'll release this in the morning.

PR #515, opened 2022-08-15
Feat/torus wallet sign and send transaction

ieow:
add signAndSendTransaction API


jordansexton:
Hmm, there's a problem with this. `@toruslabs/solana-embed` hasn't actually been updated (last published v0.2.2 ~2 months ago). This means that the TypeScript type, and the actual functionality, for the `signAndSendTransaction method, isn't there.

I also should have reviewed this PR more closely -- several of the TypeScript types are missing, so there's no way this code would have compiled. I'll be reverting this change on master. Please publish an updated version of the library, and make sure to test your implementation in a future PR.

ieow:
Sorry, my fault.
The open PR suppose to merge to https://github.com/torusresearch/wallet-adapter

PR #516, opened 2022-08-16
packages/wallets/backpack: add publicKey override

armaniferrante:
Backpack automatically updates the injected provider when the user changes wallets inside the extension. This is undesirable for the wallet adapter, which does not handle such updates. These changes use an override so that the correct wallet that the adapter is expecting is always used.

### DIFF
--- a/packages/wallets/backpack/src/adapter.ts
+++ b/packages/wallets/backpack/src/adapter.ts
@@ -26,15 +26,16 @@ interface BackpackWallet extends EventEmitter<BackpackWalletEvents> {
     isBackpack?: boolean;
     publicKey?: { toBytes(): Uint8Array };
     isConnected: boolean;
-    signTransaction(transaction: Transaction): Promise<Transaction>;
-    signAllTransactions(transactions: Transaction[]): Promise<Transaction[]>;
+    signTransaction(transaction: Transaction, publicKey?: PublicKey | null): Promise<Transaction>;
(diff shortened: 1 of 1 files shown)


jordansexton:
If Backpack fires events when the provider is updated, can the adapter update in response?

armaniferrante:
> If Backpack fires events when the provider is updated, can the adapter update in response?

It should probably do what the rest of the adapters do to avoid confusion. My understanding is that others don't respond. The other option is to simply disconnect, but that doesn't seem right either.

jordansexton:
Some wallets don't fire any event when this happens. Some fire `disconnect` events and `connect` events. The UX right now is pretty bad, so if your wallet actually handles this well and has a distinct event, we can switch accounts seamlessly and fire a `connect` event from the adapter. This is probably also what we ought to do for the Phantom adapter which has an `accountChanged` event now.

However, if Backpack actually supports JIT account switching back to the account the dapp previously connected to, then from my perspective that's actually ideal, and we can merge this as is. Let me know what you want to do!

armaniferrante:
> However, if Backpack actually supports JIT account switching back to the account the dapp previously connected to, then from my perspective that's actually ideal, and we can merge this as is. Let me know what you want to do! 

It does indeed support this!

jordansexton:
Published! `@solana/wallet-adapter-backpack@0.1.3`

PR #522, opened 2022-08-18
Adds confirmationStrategy to confirmTransaction

robbestad:
Update SendOneLamportToRandomAddress to use the new confirmationStrategy in confirmTransaction. This replaces the current deprecated function call.


jordansexton:
Thanks!

PR #528, opened 2022-08-19
React: add Wallet Standard support

jordansexton:
This PR adds support for Standard Wallets registered on `window.navigator.wallets` by wrapping them with adapters.

This allows existing apps to update their `@solana/wallet-adapter-react` with a patch version change and get support for standard wallets with no code changes.

It's marked as a draft until the Wallet Standard (and published `@wallet-standard/solana-wallet-adapter-react` package) are out of alpha.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,7 @@ jobs:
         run: pnpm install
 
       - name: Build packages
-        run: pnpm build
+        run: pnpm release
 
       - name: Run tests
         run: pnpm test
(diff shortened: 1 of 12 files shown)


steveluscher:
#yolo changed your mergebase to the `mwa-and-standard-rc` branch.

jordansexton:
My thinking on the order of operations here, but I didn't think too hard:

1. Start with any adapters from props.
2. Initialize and prepend any standard wallet adapters. The hook already does this, preferring wallets we know are there to any others, and unregistering adapters.
3. Add MWA. This should fail fast on `getIsMobile`'s "are any adapters installed check" if step 2 found any standard wallets.

steveluscher (approved):
Looks right to me! Ship it and I'll test it tomorrow to make sure it does what it says on the box.

steveluscher:
That build error is probably actually a thing though. :/

PR #530, opened 2022-08-20
Nightly -  add signMessage functionality

awojciak:
Changes:
- NightlyWalletAdapter updated to be BaseMessageSignerWalletAdapter


awojciak:
in case of error thrown on example - that's how you should verify if message was signed properly (Nightly uses js-sha256 to create hashes internally):
image

jordansexton:
Thanks! This is published

PR #537, opened 2022-08-23
Update Glow readme to give more information about setting network

vpontis:


jordansexton:
It's important for dapps that change the network used at runtime to update their version of `@solana/wallet-adapter-react` -- https://github.com/solana-labs/wallet-adapter/pull/521 fixed a problem where adapters that were reinstantiated with new config weren't being used by the React context.

PR #539, opened 2022-08-24
Bump for new version of `walletconnect-solana`

jnwng:
in attempting to release a new version of [walletconnect-solana](github.com/jnwng/walletconnect-solana), @amper-fb flagged that, contrary to my intent, that releasing a new `0.0.3` version of `walletconnect-solana` was not getting picked up in fresh installs of `@solana/wallet-adapter-walletconnect`. i was able to repro, and after conferring with the high council i learned that `^0.0.2` will not match `^0.0.3` because the caret means "first version that does not increment the first non-zero digit", which means (accurately) that a patch version was not going to get captured. learned something today!

anyway, bumped to `0.1.0` so we don't have to keep sending these updates all the way upstream (here) for small fixes.

### DIFF
--- a/packages/wallets/walletconnect/package.json
+++ b/packages/wallets/walletconnect/package.json
@@ -32,7 +32,7 @@
         "@solana/web3.js": "^1.50.1"
     },
     "dependencies": {
-        "@jnwng/walletconnect-solana": "^0.0.2",
+        "@jnwng/walletconnect-solana": "^0.1.0",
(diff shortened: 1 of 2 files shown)


steveluscher (approved):

jordansexton:
Published `@solana/wallet-adapter-walletconnect@0.1.3`, lmk if it's good!

jnwng:
looks great, thank you for the quick turnaround!

PR #542, opened 2022-08-25
Fix CJS imports

jordansexton:
https://github.com/solana-labs/solana-program-library/pull/3529 for Wallet Adapter


jordansexton:
Published.

PR #544, opened 2022-08-26
Add missing WalletConnect adapter method signAllTransactions

amper-fb:


jordansexton:
This method isn't missing, it's inherited from here:
https://github.com/solana-labs/wallet-adapter/blob/094b80c21f50060373f4c9655b2e68e5098df700/packages/core/base/src/signer.ts#L51

amper-fb:
my bad, didn't notice the repo I was working on forced resolution to an older wallet-adapter-base where this was an abstract

Issue #546, opened 2022-08-28
Need eslint config to enforce .js file extensions


jordansexton:
**Is your feature request related to a problem? Please describe.**
Explicit `.js` file extensions are required on relative imports in TypeScript files to be correctly handled by bundlers like Next.js.

Unfortunately, builds for publishing will succeed without this, leading to downstream failures for projects that depend on Wallet Adapter.

There doesn't appear to be a good way to enforce this convention. See https://github.com/solana-labs/wallet-adapter/issues/532#issuecomment-1221687443. There's an [eslint-imports](https://github.com/solana-labs/wallet-adapter/tree/eslint-imports) branch where I tried some things, but it's not working yet.

**Describe the solution you'd like**
Configuration to enforce that `.js` file extensions are always used for relative imports in the `core`, `ui`, and `wallets` package groups. Ideally, something that works with `eslint --fix` to automate this.

**Describe alternatives you've considered**
A sufficiently advanced shell script using regular expression could work in a pinch.

**Additional context**
An example of the kind of thing this needs to enforce: https://github.com/solana-labs/wallet-adapter/commit/258a609c6720434a14e293e4f927cef1cfa91541

See #545, #542, #541, #534, #533, #532, #531.

jordansexton:
This issue also affects

- [Wallet Standard](https://github.com/wallet-standard/wallet-standard)
- [SPL Token JS](https://github.com/solana-labs/solana-program-library/tree/master/token/js)
- [SPL Memo JS](https://github.com/solana-labs/solana-program-library/tree/master/memo/js)

and other projects derived from the [TypeScript Monorepo](https://github.com/jordansexton/typescript-monorepo) design that ship unbundled ESM and CJS builds.

jordansexton:
An example of the kind of thing this needs to enforce: https://github.com/solana-labs/wallet-adapter/commit/258a609c6720434a14e293e4f927cef1cfa91541

PR #547, opened 2022-08-28
use eslint to enforce .js extensions with relative import paths

johnrees:
An eslint plugin which fixes https://github.com/solana-labs/wallet-adapter/issues/546

It should autofix code with `eslint --fix` or on save with eslint plugins installed etc

https://user-images.githubusercontent.com/601961/187097384-08ef458c-234f-464b-bb5d-58139b39aac1.mov

You might want to give it a better name and publish it as an npm package if it's going to be shared amongst multiple repos, I put it in here as a starting point for convenience

I think it should cover all types of import and exports. I just pasted some code in https://astexplorer.net then copied the node type names


jordansexton:
Holy shit, this is incredible

jordansexton:
Externalized to https://github.com/solana-labs/eslint-plugin-require-extensions. Thank you so much!

PR #558, opened 2022-09-12
Add versioned transaction methods to adapter interface

jstarry:
#### Changes:
- Bump `@solana/web3.js` to v1.59.1 to bring in versioned transaction types
- Add `supportedTransactionVersions` property to the adapter interface so that wallets can declare whether they support versioned transactions and list the versions they support.
- Add a `signVersionedTransaction` method to the adapter which throws an error by default and each wallet should override
- Add a `sendVersionedTransaction` method to the adapter which checks `supportedTransactionVersions` before calling `signVersionedTransaction`

### DIFF
--- a/packages/core/base/src/adapter.ts
+++ b/packages/core/base/src/adapter.ts
@@ -1,4 +1,13 @@
-import type { Connection, PublicKey, SendOptions, Signer, Transaction, TransactionSignature } from '@solana/web3.js';
+import type {
+    Connection,
+    PublicKey,
+    SendOptions,
+    Signer,
+    Transaction,
+    TransactionSignature,
+    TransactionVersion,
+    VersionedTransaction,
+} from '@solana/web3.js';
 import EventEmitter from 'eventemitter3';
 import type { WalletError } from './errors.js';
 import { WalletNotConnectedError } from './errors.js';
@@ -28,6 +37,7 @@ export interface WalletAdapterProps<Name extends string = string> {
     publicKey: PublicKey | null;
     connecting: boolean;
     connected: boolean;
+    supportedTransactionVersions: Set<TransactionVersion> | null;
 
     connect(): Promise<void>;
     disconnect(): Promise<void>;
@@ -35,6 +45,11 @@ export interface WalletAdapterProps<Name extends string = string> {
         transaction: Transaction,
         connection: Connection,
         options?: SendTransactionOptions
+    ): Promise<TransactionSignature>;
+    sendVersionedTransaction(
+        transaction: VersionedTransaction,
+        connection: Connection,
+        options?: SendOptions
     ): Promise<TransactionSignature>;
 }
 
@@ -77,16 +92,27 @@ export abstract class BaseWalletAdapter extends EventEmitter<WalletAdapterEvents
     abstract publicKey: PublicKey | null;
     abstract connecting: boolean;
 
+    get supportedTransactionVersions(): Set<TransactionVersion> | null {
+        return null;
+    }
+
     get connected() {
         return !!this.publicKey;
     }
 
     abstract connect(): Promise<void>;
     abstract disconnect(): Promise<void>;
+
     abstract sendTransaction(
         transaction: Transaction,
         connection: Connection,
         options?: SendTransactionOptions
+    ): Promise<TransactionSignature>;
+
+    abstract sendVersionedTransaction(
+        transaction: VersionedTransaction,
+        connection: Connection,
+        options?: SendOptions
     ): Promise<TransactionSignature>;
 
     protected async prepareTransaction(
--- a/packages/core/base/src/signer.ts
+++ b/packages/core/base/src/signer.ts
@@ -1,16 +1,56 @@
-import type { Connection, Transaction, TransactionSignature } from '@solana/web3.js';
+import { Connection, SendOptions, Transaction, TransactionSignature, VersionedTransaction } from '@solana/web3.js';
 import type { SendTransactionOptions, WalletAdapter } from './adapter.js';
 import { BaseWalletAdapter } from './adapter.js';
 import { WalletSendTransactionError, WalletSignTransactionError } from './errors.js';
 
 export interface SignerWalletAdapterProps {
     signTransaction(transaction: Transaction): Promise<Transaction>;
+    signVersionedTransaction(transaction: VersionedTransaction): Promise<VersionedTransaction>;
     signAllTransactions(transaction: Transaction[]): Promise<Transaction[]>;
+    signAllVersionedTransactions(transactions: VersionedTransaction[]): Promise<VersionedTransaction[]>;
 }
 
 export type SignerWalletAdapter = WalletAdapter & SignerWalletAdapterProps;
 
 export abstract class BaseSignerWalletAdapter extends BaseWalletAdapter implements SignerWalletAdapter {
+    async sendVersionedTransaction(
+        transaction: VersionedTransaction,
+        connection: Connection,
+        options: SendOptions = {}
+    ): Promise<TransactionSignature> {
+        let emit = true;
+        try {
+            const version = transaction.message.version;
+            if (this.supportedTransactionVersions === null) {
+                throw new WalletSendTransactionError(`Sending versioned transactions isn't supported by this wallet`);
+            } else if (!this.supportedTransactionVersions.has(version)) {
+                throw new WalletSendTransactionError(
+                    `Sending transaction version ${version} isn't supported by this wallet`
+                );
+            }
+
+            try {
+                transaction = await this.signVersionedTransaction(transaction);
+
+                const rawTransaction = transaction.serialize();
+
+                return await connection.sendRawTransaction(rawTransaction, options);
+            } catch (error: any) {
+                // If the error was thrown by `signVersionedTransaction`, rethrow it and don't emit a duplicate event
+                if (error instanceof WalletSignTransactionError) {
+                    emit = false;
+                    throw error;
+                }
+                throw new WalletSendTransactionError(error?.message, error);
+            }
+        } catch (error: any) {
+            if (emit) {
+                this.emit('error', error);
+            }
+            throw error;
+        }
+    }
+
     async sendTransaction(
         transaction: Transaction,
         connection: Connection,
@@ -48,10 +88,24 @@ export abstract class BaseSignerWalletAdapter extends BaseWalletAdapter implemen
 
     abstract signTransaction(transaction: Transaction): Promise<Transaction>;
 
+    signVersionedTransaction(transaction: VersionedTransaction): Promise<VersionedTransaction> {
+        const error = new WalletSignTransactionError("Signing versioned transactions isn't supported by this wallet");
+        this.emit('error', error);
+        throw error;
+    }
+
     async signAllTransactions(transactions: Transaction[]): Promise<Transaction[]> {
         const signedTransactions: Transaction[] = [];
         for (const transaction of transactions) {
             signedTransactions.push(await this.signTransaction(transaction));
+        }
+        return signedTransactions;
+    }
+
+    async signAllVersionedTransactions(transactions: VersionedTransaction[]): Promise<VersionedTransaction[]> {
+        const signedTransactions: VersionedTransaction[] = [];
+        for (const transaction of transactions) {
+            signedTransactions.push(await this.signVersionedTransaction(transaction));
         }
         return signedTransactions;
     }
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -1,6 +1,6 @@
 import type { Adapter, SendTransactionOptions, WalletError, WalletName } from '@solana/wallet-adapter-base';
 import { WalletNotConnectedError, WalletNotReadyError, WalletReadyState } from '@solana/wallet-adapter-base';
-import type { Connection, PublicKey, Transaction } from '@solana/web3.js';
+import type { Connection, PublicKey, SendOptions, Transaction, VersionedTransaction } from '@solana/web3.js';
 import type { FC, ReactNode } from 'react';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { WalletNotSelectedError } from './errors.js';
@@ -239,6 +239,16 @@ export const WalletProvider: FC<WalletProviderProps> = ({
         [adapter, handleError, connected]
     );
 
+    // Send a versioned transaction using the provided connection
+    const sendVersionedTransaction = useCallback(
+        async (transaction: VersionedTransaction, connection: Connection, options?: SendOptions) => {
+            if (!adapter) throw handleError(new WalletNotSelectedError());
+            if (!connected) throw handleError(new WalletNotConnectedError());
+            return await adapter.sendVersionedTransaction(transaction, connection, options);
+        },
+        [adapter, handleError, connected]
+    );
+
     // Sign a transaction if the wallet supports it
     const signTransaction = useMemo(
         () =>
@@ -251,6 +261,18 @@ export const WalletProvider: FC<WalletProviderProps> = ({
         [adapter, handleError, connected]
     );
 
+    // Sign a versioned transaction if the wallet supports it
+    const signVersionedTransaction = useMemo(
+        () =>
+            adapter && 'signVersionedTransaction' in adapter
+                ? async (transaction: VersionedTransaction): Promise<VersionedTransaction> => {
+                      if (!connected) throw handleError(new WalletNotConnectedError());
+                      return await adapter.signVersionedTransaction(transaction);
+                  }
+                : undefined,
+        [adapter, handleError, connected]
+    );
+
     // Sign multiple transactions if the wallet supports it
     const signAllTransactions = useMemo(
         () =>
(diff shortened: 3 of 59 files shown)


packages/core/base/src/signer.ts
```diff
@@ -1,16 +1,56 @@
...
+        let emit = true;
+        try {
+            const version = transaction.message.version;
+            if (this.supportedTransactionVersions === null) {
+                throw new WalletSendTransactionError(`Sending versioned transactions isn't supported by this wallet`);
+            } else if (!this.supportedTransactionVersions.has(version)) {
+                throw new WalletSendTransactionError(
+                    `Sending transaction version ${version} isn't supported by this wallet`
+                );
+            }
```
jordansexton, line 30: I'm wary of increasing the public API of Wallet Adapter so much, especially with the Wallet Standard work being the primary focus now. Based on how this is implemented (supported versions exposed and runtime checks for support rather than complete type safety), I think we could just change `sendTransaction`, `signTransaction`, and `signAllTransactions` to accept `Transaction | VersionedTransaction` instead of having new public methods for all of them. Maybe we can keep these methods protected, and have the public methods call them if they are called with a `VersionedTransaction`.

packages/core/base/src/signer.ts
```diff
@@ -48,13 +88,27 @@ export abstract class BaseSignerWalletAdapter extends BaseWalletAdapter implemen
 
     abstract signTransaction(transaction: Transaction): Promise<Transaction>;
 
+    signVersionedTransaction(transaction: VersionedTransaction): Promise<VersionedTransaction> {
```
jordansexton, line 91: nit: make this `async` even though it doesn't `await` for consistency and to hint to the overrider how it should be implemented.

packages/core/base/src/adapter.ts
```diff
@@ -28,6 +37,7 @@ export interface WalletAdapterProps<Name extends string = string> {
     publicKey: PublicKey | null;
     connecting: boolean;
     connected: boolean;
+    supportedTransactionVersions: Set<TransactionVersion> | null;
```
jordansexton, line 40: We could type this such that if `supportedTransactionVersions` is null, `sendTransaction`, `signTransaction`, and `signAllTransactions` would only accept the legacy `Transaction` type. This enforces that the caller needs to check the version before they call it, which is similar to what they have to do now (in checking if the methods for `signTransaction` and `signAllTransactions` are defined).

packages/core/react/src/WalletProvider.tsx
```diff
@@ -239,6 +239,16 @@ export const WalletProvider: FC<WalletProviderProps> = ({
         [adapter, handleError, connected]
     );
 
+    // Send a versioned transaction using the provided connection
+    const sendVersionedTransaction = useCallback(
```
jordansexton, line 243: I think we can just roll all these into the existing API and keep most changes at the adapter level. If we conditionally type the methods' `transaction` inputs, in the context's `sendTransaction`, `signTransaction`, and `signAllTransactions` methods, we'll have to perform the same conditional typing against `adapter.supportedTransactionVersions`.

packages/wallets/blocto/src/adapter.ts
```diff
@@ -157,4 +157,14 @@ export class BloctoWalletAdapter extends BaseWalletAdapter {
             throw error;
         }
     }
+
+    async sendVersionedTransaction(
```
jordansexton, line 161: Shouldn't be needed if we protect this method. And it's possible if we fix Blocto's `supportedTransactionVersions` to null, the type of `sendTransaction` shouldn't need to change either.

packages/core/base/src/adapter.ts
```diff
@@ -77,18 +92,29 @@ export abstract class BaseWalletAdapter extends EventEmitter<WalletAdapterEvents
     abstract publicKey: PublicKey | null;
     abstract connecting: boolean;
 
+    get supportedTransactionVersions(): Set<TransactionVersion> | null {
+        return null;
+    }
```
jordansexton, line 97: Making this a getter may make it harder for subclasses to override with a fixed value. I think maybe we can just define it as abstract here, then define the value in `BaseSignerWalletAdapter` and Blocto's adapter.

packages/wallets/blocto/src/adapter.ts
jstarry: > And it's possible if we fix Blocto's `supportedTransactionVersions` to null, the type of `sendTransaction` shouldn't need to change either.

Can you elaborate more here on why the type of `sendTransaction` shouldn't need to change?

packages/core/base/src/signer.ts
jstarry: I've consolidated `VersionedTransaction` support into `sendTransaction` as you suggested but `signTransaction` has trickier typing because it returns the transaction

packages/core/base/src/adapter.ts
jstarry: Cool, that was my intent. Just to be clear, you weren't suggesting any changes here right?

jordansexton: Ah, I'm proposing that this would change to something like

```ts
interface WalletAdapterProps<Name extends string = string, SupportedTransactionVersions extends Set<TransactionVersion> | null = null> {
    supportedTransactionVersions: SupportedTransactionVersions;
    sendTransaction: SupportedTransactionVersions extends null
        ? (transaction: Transaction, ...) => Promise<TransactionSignature>
        : (transaction: Transaction | VersionedTransaction, ...) => Promise<TransactionSignature>
}
```

Now the caller has type safety and can't call it with `VersionedTransaction` until they check for `supportedTransactionVersions`.

packages/wallets/blocto/src/adapter.ts
jordansexton: This is a guess, but if https://github.com/solana-labs/wallet-adapter/pull/558#discussion_r969010739 is implemented, `sendTransaction` should be statically known to only accept a `Transaction`.

packages/core/base/src/signer.ts
jordansexton: Ah, yeah. I think we can do this with

```ts
<T extends Transaction | VersionedTransaction>signTransaction(transaction: T, ...): Promise<T>;
<T extends Transaction | VersionedTransaction>signAllTransactions(transactions: T[], ...): Promise<T[]>;
```

though this would also be conditional if https://github.com/solana-labs/wallet-adapter/pull/558#discussion_r969010739 is implemented.

packages/core/base/src/adapter.ts
jstarry: Oh, I misunderstood your comment, thanks for the clarification!

jordansexton: I wonder if it's possible to do this more simply... I don't know if this has the same effect?
```ts
interface WalletAdapterProps<Name extends string = string> {
    supportedTransactionVersions: Set<number> | null;
    sendTransaction: this['supportedTransactionVersions'] extends null
        ? (
              transaction: Transaction,
              connection: Connection,
              options?: SendTransactionOptions
          ) => Promise<TransactionSignature>
        : (
              transaction: Transaction | number,
              connection: Connection,
              options?: SendTransactionOptions
          ) => Promise<TransactionSignature>;
}
```

jordansexton: I wonder if it's possible to do this more simply... I don't know if this has the same effect?
```ts
interface WalletAdapterProps<Name extends string = string> {
    supportedTransactionVersions: Set<TransactionVersion> | null;
    sendTransaction: this['supportedTransactionVersions'] extends null
        ? (
              transaction: Transaction,
              connection: Connection,
              options?: SendTransactionOptions
          ) => Promise<TransactionSignature>
        : (
              transaction: Transaction | VersionedTransaction,
              connection: Connection,
              options?: SendTransactionOptions
          ) => Promise<TransactionSignature>;
}
```

jordansexton: If this can work (with generics or not), this feels pretty clean:

```ts
export type TransactionOrVersionedTransaction<SupportedTransactionVersions extends Set<TransactionVersion> | null> =
    SupportedTransactionVersions extends null ? Transaction : (Transaction | VersionedTransaction);

export interface WalletAdapterProps<Name extends string = string> {
    name: WalletName<Name>;
    // ...
    supportedTransactionVersions: Set<TransactionVersion> | null;
    sendTransaction(
        transaction: TransactionOrVersionedTransaction<this['supportedTransactionVersions']>,
        connection: Connection,
        options?: SendTransactionOptions
    ): Promise<TransactionSignature>;
}

export interface SignerWalletAdapterProps<Name extends string = string> extends WalletAdapterProps<Name> {
    signTransaction<T extends TransactionOrVersionedTransaction<this['supportedTransactionVersions']>>(
        transaction: T
    ): Promise<T>;
    signAllTransactions<T extends TransactionOrVersionedTransaction<this['supportedTransactionVersions']>[]>(
        transaction: T
    ): Promise<T>;
}
```

Speaking of which, it's probably time for the generic `Name` to go! It doesn't actually do anything useful (because you don't have type safety for which wallets are available inside of your React context, which is how this is mostly used).

packages/wallets/alpha/src/adapter.ts
```diff
@@ -201,13 +209,20 @@ export class AlphaWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+            for (const transaction of transactions) {
+                if (isVersionedTransaction(transaction))
+                    throw new WalletSendTransactionError(
+                        `Signing versioned transactions isn't supported by this wallet`
+                    );
+            }
+
             try {
-                return (await wallet.signAllTransactions(transactions)) || transactions;
+                return ((await wallet.signAllTransactions(transactions as Transaction[])) as T[]) || transactions;
```
jordansexton, line 225: All the adapters are broken now except this one, which I refactored to check out practicability of the design. This function (and to a lesser degree, `signTransaction`) isn't great. This can be better enforced with types rather than runtime checks, and there's too much in the way of type guards and casting required currently.

packages/core/base/src/signer.ts
```diff
@@ -1,42 +1,87 @@
...
+export interface SignerWalletAdapterProps<
+    Name extends string = string,
+    SupportedTransactionVersions extends Set<TransactionVersion> | null = Set<TransactionVersion> | null
+> extends WalletAdapterProps<Name, SupportedTransactionVersions> {
+    signTransaction(
+        transaction: TransactionOrVersionedTransaction<SupportedTransactionVersions>
+    ): Promise<TransactionOrVersionedTransaction<SupportedTransactionVersions>>;
+    signAllTransactions(
+        transactions: TransactionOrVersionedTransaction<SupportedTransactionVersions>[]
+    ): Promise<TransactionOrVersionedTransaction<SupportedTransactionVersions>[]>;
```
jordansexton, line 16: These types are not perfectly ideal. Let's say you have
```ts
const tx = new Transaction();
const signedTx = await adapter.signTransaction(tx);
```

In this case, if `adapter.supportedTransactionVersions` is `null`, you know `signedTx` _must be_ a `Transaction`, but its actual type is either `Transaction` or `Transaction | VersionedTransaction`.

If we declared it this way, it solves this problem:
```ts
<T extends TransactionOrVersionedTransaction<SupportedTransactionVersions>>signTransaction(transaction: T): Promise<T>;
```
But doing so makes the implementation of the classes that use this interface require a bunch of type wrangling.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -230,8 +238,8 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     }, [isDisconnecting, adapter, setName]);
 
     // Send a transaction using the provided connection
-    const sendTransaction = useCallback(
-        async (transaction: Transaction, connection: Connection, options?: SendTransactionOptions) => {
+    const sendTransaction: WalletAdapterProps['sendTransaction'] = useCallback(
+        async (transaction, connection, options) => {
```
jordansexton, line 242: The types of the methods in this React context are also not ideal. We don't know if the adapter has `supportedTransactionVersions` or not, so the methods all accept `Transaction | VersionedTransaction` even if this won't work.

packages/core/react/src/useAnchorWallet.ts
```diff
@@ -13,7 +13,13 @@ export function useAnchorWallet(): AnchorWallet | undefined {
     return useMemo(
         () =>
             publicKey && signTransaction && signAllTransactions
-                ? { publicKey, signTransaction, signAllTransactions }
+                ? {
+                      publicKey,
+                      signTransaction: signTransaction as (transaction: Transaction) => Promise<Transaction>,
+                      signAllTransactions: signAllTransactions as (
+                          transactions: Transaction[]
+                      ) => Promise<Transaction[]>,
```
jordansexton, line 21: This is a bit of a hack but should be fine -- we know Anchor's interface doesn't handle `VersionedTransaction` yet, so even if the adapter does, we are fine only allowing `Transaction`.

cc @armaniferrante

packages/wallets/alpha/src/adapter.ts
```diff
@@ -53,6 +53,7 @@ export class AlphaWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+    readonly supportedTransactionVersions = null;
```
jordansexton, line 56: Now the only change required for all of the adapters is to statically provide the value of the `SupportedTransactionVersions` generic type to implement the abstract class. Because we set it to null and readonly, we know that our existing `sendTransaction`, `signTransaction`, and `signAllTransaction` methods only need to accept the `Transaction` type they already have.

jordansexton: Implementing adapters should remove the `readonly` modifier and change this value according to what the wallet supports. While this could be done statically, they will probably want to expose and read this from the wallet itself somehow. We should work with the major wallets to define how this is done in a standard way.

packages/core/react/src/WalletProvider.tsx
jstarry: I think that's ok because the app will be doing runtime checks rather than compile time checks for versioned transaction support for a particular wallet.

jstarry: Oh, I see that this applies to `signTransaction` as well. That's less ideal

packages/wallets/alpha/src/adapter.ts
jstarry: Yeah that would be nice. I guess there are cases when browser extension wallets don't auto-update?

packages/core/react/src/WalletProvider.tsx
jordansexton: Yeah. Worse, I think as currently written, it causes `signTransaction` to return `Transaction | VersionedTransaction` which will break the types of all callers.

packages/core/base/src/signer.ts
```diff
@@ -1,42 +1,78 @@
...
-export interface SignerWalletAdapterProps {
-    signTransaction(transaction: Transaction): Promise<Transaction>;
-    signAllTransactions(transaction: Transaction[]): Promise<Transaction[]>;
+export interface SignerWalletAdapterProps<Name extends string = string> extends WalletAdapterProps<Name> {
+    signTransaction<T extends TransactionOrVersionedTransaction<this['supportedTransactionVersions']>>(
+        transaction: T
+    ): Promise<T>;
+    signAllTransactions<T extends TransactionOrVersionedTransaction<this['supportedTransactionVersions']>>(
+        transactions: T[]
+    ): Promise<T[]>;
```
jordansexton, line 13: Okay, I think these types are very reasonable now. By inferring `SupportedTransactionVersions` from the actual implementer, and making the methods return the type they were called with, existing code that expects `signTransaction(transaction: Transaction): Promise<Transaction>` will just work.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -240,10 +247,10 @@ export const WalletProvider: FC<WalletProviderProps> = ({
     );
 
     // Sign a transaction if the wallet supports it
-    const signTransaction = useMemo(
+    const signTransaction: SignerWalletAdapterProps['signTransaction'] | undefined = useMemo(
```
jordansexton, line 250: This type is still _perfect_ in the sense that a dev will see `(transaction: Transaction | VersionedTransaction)` as the parameter type, so they have to know to check the adapter for `supportedTransactionVersions`. But it is functional and nonbreaking, because it will always be expected to return whichever type it was called with.

packages/wallets/alpha/src/adapter.ts
```diff
@@ -185,13 +186,13 @@ export class AlphaWalletAdapter extends BaseMessageSignerWalletAdapter {
         }
     }
 
-    async signTransaction(transaction: Transaction): Promise<Transaction> {
+    async signTransaction<T extends Transaction>(transaction: T): Promise<T> {
```
jordansexton, line 189: We know that transaction will only ever be `Transaction` now because `supportedTransactionVersions` is null, but we still have to implement it this way because the interface declares it like this. All what this means is that if a subtype of `Transaction` was passed to `signTransaction`, the wallet _should_ return the same subtype. In practice, this doesn't matter because there are no subclasses of `Transaction` in web3.js to worry about.

packages/wallets/alpha/src/adapter.ts
```diff
@@ -185,13 +186,13 @@ export class AlphaWalletAdapter extends BaseMessageSignerWalletAdapter {
...
         try {
             const wallet = this._wallet;
             if (!wallet) throw new WalletNotConnectedError();
 
             try {
-                return (await wallet.signTransaction(transaction)) || transaction;
+                return ((await wallet.signTransaction(transaction)) as T) || transaction;
```
jordansexton, line 195: Here we have to perform the one ugly bit, where we cast `Transaction` as `T` even though we know it will be valid.

packages/core/base/src/signer.ts
jordansexton: This is fixed now.

packages/core/react/src/WalletProvider.tsx
jordansexton: This is fixed now.

packages/core/react/src/WalletProvider.tsx
jstarry: That seems totally fine to me

jstarry:
The last round of changes look solid to me. I just built the example app with some test code that called `signTransaction` from `useWallet` and API type resolution works great. Your call but I say ship it now and we can iterate quickly on if there are any minor issues that pop up

jordansexton:
Sounds good, I'll test a bit and merge tomorrow.

PR #564, opened 2022-09-15
Fixing inline style for the react-ui WalletMultiButton

viniciuskloppel:
I recently realized that the style props of the WalletMultiButton wasn't working properly. It was not possible to override the styles.css by passing a backgroundColor, for example. That was happening because the style was not getting at the final component: Button.tsx

This is my first PR as a dev by the way 🥳

### DIFF
--- a/packages/ui/react-ui/src/Button.tsx
+++ b/packages/ui/react-ui/src/Button.tsx
@@ -16,7 +16,7 @@ export const Button: FC<ButtonProps> = (props) => {
         <button
             className={`wallet-adapter-button ${props.className || ''}`}
             disabled={props.disabled}
-            style={{...props.style}}
+            style={props.style}
             onClick={props.onClick}
             tabIndex={props.tabIndex || 0}
             type="button"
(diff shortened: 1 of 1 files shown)


packages/ui/react-ui/src/Button.tsx
```diff
@@ -16,6 +16,7 @@ export const Button: FC<ButtonProps> = (props) => {
         <button
             className={`wallet-adapter-button ${props.className || ''}`}
             disabled={props.disabled}
+            style={...props.style}
```
jordansexton, line 19: Hmm, is this right? Wouldn't it be either `style={{...props.style}}` or just `style={props.style}`?

viniciuskloppel: Definitely. The correct is style={{...props.style}}. I saw that after submitting the pull request, but didn't know if I should send another commit right next.

What is the best practice in that case? Should I have sent another commit fixing my error or should I have waited you review the change like I did?

jordansexton: Always better to fix it before review if you find it first!

jordansexton: I'm not sure why we can't just pass `style={props.style}` though. No reason to spread (copy) the object.

viniciuskloppel: Sure, let's use `style={props.style}` then.

jordansexton:
Thanks so much!

PR #567, opened 2022-09-15
Add versioned tx sending components to example app

jstarry:
Add two components the example app for sending legacy and v0 transactions via the `VersionedTransaction` API. 

Screen Shot 2022-09-15 at 3 59 20 PM

### DIFF
--- a/packages/starter/example/src/components/SendV0Transaction.tsx
+++ b/packages/starter/example/src/components/SendV0Transaction.tsx
@@ -28,6 +28,11 @@ export const SendV0Transaction: FC = () => {
 
         let signature: TransactionSignature = '';
         try {
+            /**
+             * This lookup table only exists on devnet and can be replaced as
+             * needed.  To create and manage a lookup table, use the `solana
+             * address-lookup-table` commands.
+             */
             const lookupTable = (await connection.getAddressLookupTable(new PublicKey("F3MfgEJe1TApJiA14nN2m4uAH4EBVrqdBnHeGeSXvQ7B"))).value;
             if (lookupTable === null) {
                 notify('error', 'Address lookup table wasn\'t found!');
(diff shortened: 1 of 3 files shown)


packages/starter/example/src/components/SendV0Transaction.tsx
```diff
@@ -0,0 +1,66 @@
...
+            return;
+        }
+
+        let signature: TransactionSignature = '';
+        try {
+            const lookupTable = (await connection.getAddressLookupTable(new PublicKey("F3MfgEJe1TApJiA14nN2m4uAH4EBVrqdBnHeGeSXvQ7B"))).value;
+            if (lookupTable === null) {
+                notify('error', 'Address lookup table wasn\'t found!');
+                return;
+            }
```
jordansexton, line 26: What about a button that actually creates an LUT at a random address (once they get airdropped, can this be done using a legacy tx?) -- this would demonstrate how to use those instructions and make the example self-contained

packages/starter/example/src/components/SendV0Transaction.tsx
```diff
@@ -0,0 +1,66 @@
...
+            await connection.confirmTransaction({ blockhash, lastValidBlockHeight, signature });
+            notify('success', 'Transaction successful!', signature);
+        } catch (error: any) {
+            notify('error', `Transaction failed! ${error?.message}`, signature);
+            return;
+        }
+    }, [publicKey, notify, connection, sendTransaction]);
+
+    return (
+        <Button variant="contained" color="secondary" onClick={onClick} disabled={!publicKey}>
```
jordansexton, line 62: These buttons would probably be disabled / hidden if the adapter doesn't support versioned transactions, right?

jstarry: Good point, will fix

packages/starter/example/src/components/SendV0Transaction.tsx
jstarry: Yeah fair, I was initially considering that but cut scope. How would you expect LUT creation to be connected to sending a v0 transaction? Maybe local storage?

jstarry: I'd like to skip this in the interest of time. Address lookup tables can be created using the solana cli using `solana address-lookup-table` commands. I added a comment here explaining that.

jordansexton: Sounds good!

PR #568, opened 2022-09-19
feat: Add OneKey Wallet Adapter

sidmorizon:


jordansexton:
Thanks for your contribution. We are no longer accepting PRs of adapters for this type of wallet. You can [read about this policy here](https://github.com/solana-labs/wallet-adapter/blob/master/WALLET.md), but more importantly, please check out [this guide for wallets](https://github.com/solana-labs/wallet-standard/blob/master/WALLET.md) to implement the Wallet Standard.

However, most of what you've written for this adapter is portable to the Wallet Standard interface, and implementing this interface will grant your wallet distribution on dapps across Solana.

After reading the guide and looking at the reference implementations, if you have questions on how to do this, please open an issue on the https://github.com/solana-labs/wallet-standard repo.

PR #571, opened 2022-09-20
Repair tests and run in CI

steveluscher:
These tests were busted and furthermore not running on PRs. Now they will run on PRs and pushes!


steveluscher:
Wow. That build step is _slow_ in GitHub's CI.

jordansexton:
Thanks!

> That build step is slow in GitHub's CI.

Is there a way we can cache build artifacts and use `pnpm build:ts` to take advantage of incremental builds?

steveluscher:
Surely. Something something `actions/cache@v3`, something something `hash(pnpm-lock.json, **/src)`, something something.

I quickly Googled then gave up for the time being.

PR #572, opened 2022-09-21
feat: sign message encoding param

vbstreetz:
ability to provide encoding to use to sign message

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -124,7 +124,7 @@ export const SignMessageButton: FC = () => {
             // Encode anything as bytes
             const message = new TextEncoder().encode('Hello, world!');
             // Sign the bytes using the wallet
-            const signature = await signMessage(message);
+            const signature = await signMessage(message, 'utf8');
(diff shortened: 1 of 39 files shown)


jordansexton:
The plan is that will be handled as part of https://github.com/solana-labs/solana/pull/26915 (see the "Message Formats" section) and encoded in the bytes rather than an extraneous parameter.

The above linked spec allows a restricted ASCII charset for display on Ledger. In practice, I've also never seen a use case for interpreting the bytes as anything other than UTF-8.

jordansexton:
I searched the existing wallet TS APIs `/signMessage\(.*?\):/` and didn't find any results indicating that wallets support this extra argument. If they do, it's not documented here.

I'm closing this for now, but we could see if this makes sense as an optional parameter to the [Wallet Standard `signMessage` feature](https://github.com/wallet-standard/wallet-standard/blob/master/packages/core/features/src/signMessage.ts).

PR #574, opened 2022-09-22
Enable important React Hooks linters

steveluscher:
The exhaustive-deps linter is so so important. Enabling it now before I go and make a bunch of changes to the React code.


jordansexton:
Thanks!

PR #387, opened 2022-04-12
WIP Fractal Wallet adapter

kcking:
- [x] update logo
- [x] update to production URL

PR #577, opened 2022-09-30
Adds Fractal Wallet Adapter Implementation

obber:
Hey folks at Solana labs. This is an adapter for Fractal wallets and extends on the original PR from @kcking: #387.

Some context:

Fractal wallet is non-custodial, so the private keys for signing transactions are only available when an authenticated user visits `fractal.is`. We don't store the user's keypair in any capacity on our end, so signing can only be done while a user visits fractal.is

For this reason, the `connect` function essentially opens a popup and asks for user approval to share their public key with the requesting domain (whatever domain the dapp is hosted on that's running fractal wallet adapter.)

Same thing goes for the `signTransaction` and `signAllTransactions` methods -- these are re-implemented using a custom sign-transactions page on fractal.is that's opened in a popup and the user can approve those transactions.

Please let me know if this makes sense.

Also, could you let me know how I can publish a private version of this repo to test this end-to-end? We can publish this private version in our npm namespace: `@fractalwagmi/some-package-name`.

### DIFF
--- a/packages/wallets/fractal/src/adapter.ts
+++ b/packages/wallets/fractal/src/adapter.ts
@@ -1,0 +1,60 @@
+import type { SendTransactionOptions, WalletName } from '@solana/wallet-adapter-base';
+import { BaseWalletAdapter, WalletReadyState } from '@solana/wallet-adapter-base';
+import type { Connection, Transaction, TransactionSignature } from '@solana/web3.js';
+import { PublicKey } from '@solana/web3.js';
+
+export const FakeWalletName = 'Fake Wallet' as WalletName<'Fake Wallet'>;
+
+export class FakeWalletAdapter extends BaseWalletAdapter {
+    name = FakeWalletName;
+    url = 'https://github.com/solana-labs/wallet-adapter#usage';
+    icon =
+        'data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMzQiIGhlaWdodD0iMzAiIGZpbGw9Im5vbmUiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PHBhdGggZmlsbC1ydWxlPSJldmVub2RkIiBjbGlwLXJ1bGU9ImV2ZW5vZGQiIGQ9Ik0zNCAxMC42djIuN2wtOS41IDE2LjVoLTQuNmw2LTEwLjVhMi4xIDIuMSAwIDEgMCAyLTMuNGw0LjgtOC4zYTQgNCAwIDAgMSAxLjMgM1ptLTQuMyAxOS4xaC0uNmw0LjktOC40djQuMmMwIDIuMy0yIDQuMy00LjMgNC4zWm0yLTI4LjRjLS4zLS44LTEtMS4zLTItMS4zaC0xLjlsLTIuNCA0LjNIMzBsMS43LTNabS0zIDVoLTQuNkwxMC42IDI5LjhoNC43TDI4LjggNi40Wk0xOC43IDBoNC42bC0yLjUgNC4zaC00LjZMMTguNiAwWk0xNSA2LjRoNC42TDYgMjkuOEg0LjJjLS44IDAtMS43LS4zLTIuNC0uOEwxNSA2LjRaTTE0IDBIOS40TDcgNC4zaDQuNkwxNCAwWm0tMy42IDYuNEg1LjdMMCAxNi4ydjhMMTAuMyA2LjRaTTQuMyAwaC40TDAgOC4ydi00QzAgMiAxLjkgMCA0LjMgMFoiIGZpbGw9IiM5OTQ1RkYiLz48L3N2Zz4=';
+    readonly supportedTransactionVersions = null;
+    private _publicKey: PublicKey | null = null;
+
+    constructor() {
+        super();
+        console.warn(
+            'Your application is presently configured to use the `FakeWalletAdapter`. ' +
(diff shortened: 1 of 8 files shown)


jordansexton:
Awesome, I was wondering when we were going to see an updated PR for this. I'll review it soon, thanks!

Can I also draw your attention to https://github.com/wallet-standard/wallet-standard and have your team review it?

We're still accepting adapters here, but quickly heading in a direction where they won't be required for browser wallets.

Even if Fractal isn't an injected wallet like that, I think it's worth looking at the API. We have a Slack Connect channel for the work being done here if you're interested in joining.

packages/wallets/fractal/src/adapter.ts
```diff
@@ -0,0 +1,207 @@
...
+                const message = Message.from(base58.decode(signedB58Transaction));
+                return Transaction.populate(message);
+            }) as T[];
+
+            resolve(signedTransactions);
+        };
+
+        const nonce = uuid.v4();
+        this.popupManager.open({
+            url: `${FRACTAL_DOMAIN_HTTPS}/wallet-adapter/sign/${nonce}`,
```
dborstelmann, line 172: should we make the whole URL minus the nonce a const?

obber: Done.

packages/wallets/fractal/src/adapter.ts
```diff
@@ -0,0 +1,209 @@
...
+    assertPayloadIsTransactionSignatureNeededResponsePayload,
+} from '@fractalwagmi/popup-connection';
+import type { TransactionSignatureNeededPayload } from '@fractalwagmi/popup-connection';
+import uuid from 'uuid';
+
+const FRACTAL_DOMAIN_HTTPS = 'https://fractal.is';
+const APPROVE_PAGE_URL = `${FRACTAL_DOMAIN_HTTPS}/wallet-adapter/approve`;
+const SIGN_PAGE_URL = `${FRACTAL_DOMAIN_HTTPS}/wallet-adapter/sign`;
+const MIN_POPUP_HEIGHT_PX = DEFAULT_POPUP_HEIGHT_PX;
+const MAX_POPUP_WIDTH_PX = 850;
```
jordansexton, line 34: The `@solana/wallet-adapter-fractal` package should be as dumb and thin as possible, and have no external deps. `@fractalwagmi/popup-connection` and `uuid` must be imported as types only here (`uuid` may not be needed at all). All loadable adapters must load any external SDK asynchronously in the `connect` function. Here's how the adapters for [Torus](https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/torus/src/adapter.ts#L76-L130) and [WalletConnect](https://github.com/solana-labs/wallet-adapter/blob/e0f88df48c26bfd69b9ae6f34b5c1439f926bdfd/packages/wallets/walletconnect/src/adapter.ts#L64-L108) do it. The `uuid` dependency and the constants initialized below should all be moved to a package you maintain, e.g. `@fractalwagmi/solana-wallet-adapter` and imported async as a single dependency in this package with minimal setup in the `connect` function.

packages/wallets/fractal/src/adapter.ts
```diff
@@ -0,0 +1,209 @@
...
+
+export const FractalWalletName = 'Fractal Wallet' as WalletName<'Fractal Wallet'>;
+
+export class FractalWalletAdapter extends BaseSignerWalletAdapter {
+    name = FractalWalletName;
+    url = 'https://github.com/solana-labs/wallet-adapter#usage';
```
jordansexton, line 40: Probably want this to be `https://fractal.is` or even better, a page specific to the wallet that tells users the benefits of it.

jordansexton (changes requested):
I partially reviewed this. Structurally, it should look a lot like the Torus or WalletConnect adapters. It probably makes sense to copy one of them and then just change what's needed for loading your own singular library dependency.

obber:
> Even if Fractal isn't an injected wallet like that, I think it's worth looking at the API. We have a Slack Connect channel for the work being done here if you're interested in joining. 

Yes, I'd like to join this channel. How do I join?

packages/wallets/fractal/src/adapter.ts
obber: Got it, thanks! I'll make those changes and follow up.

I noticed you didn't mention the `bs58` module but I am loading it as an external dep. Should I also load that one asynchronously?

obber: Nevermind on my last comment, when I move everything into the async-loaded lib, bs58 will go along with it so this package should no longer depend on it.

jordansexton: All good! Yeah, good intuition. Also, bs58 is a special case. It gets a free pass because it's a hard dependency of web3.js Pubkey (and other stuff).

jordansexton:
> Yes, I'd like to join this channel. How do I join?

@obber [shoot me a DM](twitter.com/jordaaash) with your email for Slack! I'll add ya

packages/wallets/fractal/src/adapter.ts
obber: Gotcha. ok, I've moved over basically all of the implementation into `@fractalwagmi/solana-wallet-adapter` and cleaned up this `adapter.ts` accordingly by forking over the Torus wallet. PTAL. Thanks!

packages/wallets/fractal/src/adapter.ts
obber: Done. (The page is still being worked on on our end, but I've updated the URL preemptively.)

PR #580, opened 2022-10-04
Fix supportedTransactionVersions type

vsakos:
This pull request fixes the following Typescript error:

```
Type 'SolflareWalletAdapter' is not assignable to type 'WalletAdapterProps<string>'.
  Types of property 'supportedTransactionVersions' are incompatible.
    Type 'ReadonlySet<TransactionVersion>' is missing the following properties from type 'Set<TransactionVersion>': add, clear, delete, [Symbol.toStringTag]
```

It can be reproduced with:
```
yarn create react-app --template cra-template-solana-typescript my-dapp
cd my-dapp && yarn start
```


jordansexton:
Hmm, this fix doesn't look quite right.

https://github.com/solana-labs/wallet-adapter/blob/80d1d486a68cc59dc5c55fae17c141d2d6598e25/packages/core/base/src/adapter.ts#L32

https://github.com/solana-labs/wallet-adapter/blob/0232ccb4e8428f147b203b69b161b83ca21a390c/packages/core/base/src/types.ts#L13

This suggests the adapter's type should be what it is now:
```ts
supportedTransactionVersions: ReadonlySet<TransactionVersion> = new Set(['legacy', 0]);
```
The change in this PR:
```ts
readonly supportedTransactionVersions = new Set(['legacy', 0]) as Set<TransactionVersion>;
```

Declaring the property `supportedTransactionVersions` as `readonly` does not satisfy the requirement that the type is a `ReadonlySet` (the value must be read-only, not using the property modifier `readonly`).

`new Set(['legacy', 0]) as Set<TransactionVersion>;` should also not be needed. `new Set<TransactionVersion>(['legacy', 0])` would be more correct, but also shouldn't be needed.

The example Unsafe Burner wallet uses the expected type:
https://github.com/solana-labs/wallet-adapter/blob/0232ccb4e8428f147b203b69b161b83ca21a390c/packages/wallets/unsafe-burner/src/adapter.ts#L17

The starter CRA project uses this wallet:
https://github.com/solana-labs/wallet-adapter/blob/10445303bfe6ddf23dca60f8cd2ba0eac8c1beeb/packages/starter/create-react-app-starter/src/App.tsx#L37

The example project also uses this wallet:
https://github.com/solana-labs/wallet-adapter/blob/10445303bfe6ddf23dca60f8cd2ba0eac8c1beeb/packages/starter/example/src/components/ContextProvider.tsx#L70

The example works here: https://solana-labs.github.io/wallet-adapter/example/

I'll do some further testing.

vsakos:
I agree that using `ReadonlySet` is more correct than just using `Set`, but `ReadonlySet` is basically a subset of the `Set` type so it's definitely not incorrect.

For some reason the code in the template expects a `Set` and the provided `ReadonlySet` in the Solflare adapter does not satisfy this requirement. Since the example project works fine (I know for sure because that's where I tested the initial VersionedTransaction support in Solflare), it might be a problem with the template project's configuration, but I don't know what might be the issue.

Also this is the first time I'm using `ReadonlySet` (which as I understood is pretty much a type annotation that does not have an equivalent JS type but limits the usage of `Set` compile-time) so I don't know how to approach this problem other than just using `Set` which also satisfies the adapter interface.

We definitely have to solve this ASAP since people are actively using Solflare through the wallet adapter and they reported having problems with this (for now only fixing the package version to the previous release works) but since this is not a Solflare specific issue, every wallet that implements versioned transactions will have the same problem.

jordansexton:
Can you tell me how to reproduce the error you're seeing? I'm not seeing this when building the entire project including all the starter packages.

jordansexton:
Oops, sorry. I spaced out and forgot you included repro instructions that I saw when I first looked at this 🤣

JulI0-concerto:
I got the same issue.
./src/pages/_app.tsx:68:21
Type error: Type 'SolflareWalletAdapter[]' is not assignable to type 'Adapter[]'.
  Type 'SolflareWalletAdapter' is not assignable to type 'Adapter'.
    Type 'SolflareWalletAdapter' is not assignable to type 'SignerWalletAdapter<string>'.
      Type 'SolflareWalletAdapter' is not assignable to type 'WalletAdapterProps<string>'.
        Types of property 'supportedTransactionVersions' are incompatible.
          Type 'ReadonlySet<TransactionVersion>' is missing the following properties from type 'Set<TransactionVersion>': add, clear, delete, [Symbol.toStringTag]

  66 |             <ConnectionProvider endpoint={endpoint}>
  67 |                 <WalletProvider
> 68 |                     wallets={wallets}
     |                     ^
  69 |                     localStorageKey={'wallets'}
  70 |                     autoConnect
  71 |                 >

jordansexton:
This should be fixed now. `@solana/wallet-adapter-base` v0.9.18 fixes the unintentional breaking change that was introduced, `@solana/wallet-adapter-wallets` v0.19.1 (and all other packages) explicitly depend on it.

PR #585, opened 2022-10-05
The React `WalletProvider` is now compatible with the Solana Mobile Stack Mobile Wallet Adapter by default

steveluscher:
#### Preamble: How to review this PR

Click on the ‘commits’ tab and treat _each_ commit as though it were its own PR. Every commit in this PR represents one feature/change, is tested when relevant, and can be landed independent of the ones that come after it.

---

#### Problem

The goal of this PR is to make a change to `@solana/wallet-adapter-react` such that we can make the following claim to existing Solana web application developers: “make no changes to your app, other than upgrading to `vX.Y.Z`, and your web app will be able to connect to any mobile wallet that supports the SMS mobile wallet adapter protocol.”

#### Proposed solution

Detect whether the app is running on a compatible mobile device. If it is, automatically construct and select a mobile wallet adapter, so that the users first interaction with the ‘connect’ button will trigger a mobile wallet adapter association request.

<details>
<summary>Implementation details</summary>

In this PR, you will encounter solutions that attempt to bridge the gap between how browser wallet extensions and mobile wallet apps work.

1. Browser wallet adapter plugins represent exactly one wallet brand. The mobile wallet adapter plugin represents the MWA protocol, which can connect to _any_ wallet brand that supports it.
2. Browser wallets impose the concept of connecting/disconnecting – a step that is done on every load of the site. Mobile wallet adapter imposes the concept of authorize/deauthorize – steps that need only be taken once. ‘Disconnecting’ a mobile wallet in this PR has been implemented as ‘willfully forgetting a prior authorization.’
3. Autoconnect is not possible with mobile wallet adapter, because the OS enforces that an MWA association _has_ to start with a user gesture (tap, swipe, etc). Autoconnect in this PR is implemented as ‘remembering and loading in your cached authorization, if you have one.’
4. MWA doesn't work at all in WebViews, because WebViews lack the privileges to open WebSocket connections on `localhost`. We disable MWA when running in a WebView, by user-agent detection.
5. We want to keep letting wallets control their own in-app browsers. There's no good way to detect whether we're _running_ inside a browser's in-app browser, since _most_ wallets haven't customized the user-agent of their webviews. Instead, what we've done is to disable the MWA injection whenever we detect that _one_ of the configured wallet adapter plugins is in the `Installed` ready-state.
6. The ‘connect’ UI has _always_ been unsafe for server-side rendering (SSR) since it's _always_ been impossible for the server to know what wallets the client has, and how the button should render. This code – through its use of auto-selection of the mobile wallet adapter – makes SSR mismatches much more likely. People using SSR on their sites are going to need to mark their connect buttons as dynamic, at the very least.
7. The MWA plugin that we auto-instantiate infers the cluster from the RPC endpoint URL provided by the `ConnectionProvider`, and configures the `AppIdentity` using the hostname of the website alone. Folks who want or need to customize this further can supply their _own_ instance of `SolanaMobileWalletAdapter` in the `wallets` array, and this code will choose theirs in lieu of creating a default one.
</details>

##### Connecting and disconnecting a wallet on mobile web

The MWA gets selected automatically. Tapping connect starts an association with a mobile wallet. Tapping disconnect ‘forgets’ the cached authorization.

https://user-images.githubusercontent.com/13243/194161385-5f3b4288-a340-4cab-9f0f-95bc4167394d.mp4

##### Connecting and disconnecting a wallet inside a wallet's in-app browser

If we're in a wallet app's in-app browser and their wallet gets detected, we disable the MWA. Tapping connect starts an association with the host's adapter. Tapping disconnect disconnects the host adapter.

https://user-images.githubusercontent.com/13243/194161758-e612ab87-4405-4bf9-b5ba-c9c6c3da8133.mp4

#### Auto-connect enabled

Refreshing the page with `autoConnect` set to `true` results in the MWA ‘remembering’ the prior authorization and reloading that into state. At no point does it make contact with the mobile wallet.

https://user-images.githubusercontent.com/13243/194162400-e586cafe-afc5-4a27-ae61-1b9a040fc7a2.mp4

#### Auto-connect disabled

When auto-connect is not enabled, the wallet still loads by default but the user must initiate the `connect` feature to take the next step. The MWA simply ‘remembers’ the prior authorization and reloads that into state. At no point does it make contact with the mobile wallet.

https://user-images.githubusercontent.com/13243/194162823-dc698f34-3ead-4d56-846f-fc56008bf42d.mp4

#### Full test case

A mobile application using this new adapter code with no other changes.

https://user-images.githubusercontent.com/13243/194164711-b86103f3-5047-4998-914c-c40470cf04c4.mp4


packages/core/react/src/WalletProviderBase.tsx
```diff
@@ -6,50 +6,48 @@ import type {
...
+        isUnloadingRef: React.RefObject<boolean>;
+        // NOTE: The presence/absence of this function implies that auto connect is enabled/disabled.
+        onAutoConnect?: () => Promise<void>;
+        onSelectWallet: (walletName: WalletName) => void;
+    }
+>;
 
-export const WalletProvider: FC<WalletProviderProps> = ({
+export function WalletProviderBase({
+    adapter,
```
steveluscher, line 28: `WalletProviderBase` no longer assumes the responsibility of switching adapters. The higher-level `WalletProvider` does that. `WalletProviderBase` just takes the adapter it's given and prepares the `WalletContextState`.

packages/core/react/src/__tests__/WalletProviderDesktop-test.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { WalletProvider } from '../WalletProvider.js';
+import type { WalletProviderProps } from '../WalletProvider.js';
+import { act } from 'react-dom/test-utils';
+import { createRoot } from 'react-dom/client';
+import { useWallet } from '../useWallet.js';
+
+jest.mock('../getEnvironment.js', () => ({
+    ...jest.requireActual('../getEnvironment.js'),
+    __esModule: true,
+    default: () => jest.requireActual('../getEnvironment.js').Environment.DESKTOP_WEB,
```
steveluscher, line 25: Other than mocking `getEnvironment` differently, these two tests are identical. They will start to diverge in future commits, as mobile functionality is introduced.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -15,14 +22,69 @@ export interface WalletProviderProps {
...
+
 export function WalletProvider({
     autoConnect,
     localStorageKey = 'walletName',
     wallets: adapters,
     ...props
 }: WalletProviderProps) {
-    const [walletName, setWalletName] = useLocalStorage<WalletName | null>(localStorageKey, null);
-    const adapter = useMemo(() => adapters.find((a) => a.name === walletName) ?? null, [adapters, walletName]);
+    const mobileWalletAdapter = useMemo(() => {
```
steveluscher, line 68: If on mobile, this either prepares a fresh mobile wallet adapter or uses the one the developer has supplied in the `wallets` array.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -15,14 +22,69 @@ export interface WalletProviderProps {
...
+        return new SolanaMobileWalletAdapter({
+            addressSelector: createDefaultAddressSelector(),
+            appIdentity: {
+                uri: getUriForAppIdentity(),
+            },
+            authorizationResultCache: createDefaultAuthorizationResultCache(),
+            cluster: 'mainnet-beta',
+        });
+    }, [adapters]);
+    const adaptersWithDefaultsInjected = useMemo(() => {
```
steveluscher, line 87: If on mobile, this patches up the `wallets` array to include the mobile wallet adapter.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -15,14 +22,69 @@ export interface WalletProviderProps {
...
+            return adapters;
+        }
+        if (adapters.indexOf(mobileWalletAdapter as Adapter) !== -1) {
+            return adapters;
+        }
+        return [mobileWalletAdapter, ...adapters] as Adapter[];
+    }, [adapters, mobileWalletAdapter]);
+    const [walletName, setWalletName] = useLocalStorage<WalletName | null>(
+        localStorageKey,
+        getIsMobile(adapters) ? SolanaMobileWalletAdapterWalletName : null
```
steveluscher, line 98: If on mobile, this makes sure that the mobile wallet name is selected by default if we don't already have a cached wallet name in `localStorage`.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -30,6 +32,19 @@ function getUserAgent() {
...
 }
 
+function getInferredClusterFromEndpoint(endpoint?: string): Cluster {
+    if (endpoint == null) {
+        return 'mainnet-beta';
+    }
+    if (/devnet/i.test(endpoint)) {
+        return 'devnet';
+    } else if (/testnet/i.test(endpoint)) {
+        return 'testnet';
```
steveluscher, line 42: This is the dumbest possible thing we can do, but will work Most of the Time™. If you have a URL that both does not contain these words _and_ does not use mainnet, then you can supply a custom `SolanaMobileWalletAdapter` that makes the correct choice of `cluster` for your use case.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -49,6 +64,7 @@ export function WalletProvider({
     wallets: adapters,
     ...props
 }: WalletProviderProps) {
+    const { connection } = useConnection();
```
steveluscher, line 67: Unfortunately this introduces a dependency on `ConnectionProvider` here, but let's be honest, it has _always_ been a soft dependency since you can very nearly accomplish nothing transaction related unless you have a `ConnectionProvider`.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -124,8 +124,12 @@ export function WalletProvider({
         if (autoConnect !== true || !adapter) {
             return;
         }
-        return adapter.connect.bind(adapter);
-    }, [adapter, autoConnect]);
+        if (walletName === SolanaMobileWalletAdapterWalletName && getIsMobile(adapters)) {
+            return (adapter as SolanaMobileWalletAdapter).autoConnect_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.bind(adapter);
```
steveluscher, line 128: This exists because `connect()` does not distinguish between user-initiated connection requests and auto-initiated connection requests. In MWA, if the connection requests is auto-initiated, we do _not_ want to try to open the mobile wallet app; the OS will block such an attempt unless the request was initiated by a user gesture (eg. a tap or a swipe).

This special codepath attempts to load a prior authorization out of the cache, but otherwise does not contact the wallet if one can't be found.

packages/core/react/src/__tests__/WalletProviderMobile-test.tsx
```diff
@@ -233,7 +233,7 @@ describe('WalletProvider when the environment is `MOBILE_WEB`', () => {
     });
     describe('when there exists no stored wallet name', () => {
         beforeEach(() => {
-            (localStorage.getItem as jest.Mock).mockReturnValue(null);
```
steveluscher: These mocks never unhook. Much cleaner to just call `removeItem` or `setItem` to massage the cache into the state you want for the test, to later be cleared by `localStorage.clear()` in the top-level `beforeEach()` hook.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -148,6 +148,10 @@ export function WalletProvider({
     }, [adapter]);
     const isUnloading = useRef(false);
     useEffect(() => {
+        if (walletName === SolanaMobileWalletAdapterWalletName && getIsMobile(adapters)) {
+            isUnloading.current = false;
+            return;
```
steveluscher, line 153: Merely switching away from the browser to visit the mobile wallet app fires the `beforeunload` event. We do _not_ want event handlers disabled in this case, so we completely disable this hack when on mobile.

packages/core/react/src/WalletProvider.tsx
jordansexton: What should happen if the endpoint the dapp has configured is localhost / 127.0.0.1?

packages/core/react/src/WalletProvider.tsx
jordansexton: In the event that a dapp passes a `connection` arg to `sendTransaction` that is not the one from `useConnection`, what should happen?

jordansexton: Also, if the dapp is configured without `ConnectionProvider`, should we do anything here?

packages/core/react/src/WalletProvider.tsx
```diff
@@ -65,9 +81,9 @@ export function WalletProvider({
                 uri: getUriForAppIdentity(),
             },
             authorizationResultCache: createDefaultAuthorizationResultCache(),
-            cluster: 'mainnet-beta',
+            cluster: getInferredClusterFromEndpoint(connection?.rpcEndpoint),
         });
-    }, [adapters]);
+    }, [adapters, connection?.rpcEndpoint]);
```
jordansexton, line 86: Ah, I see the case that it's undefined is handled here.

packages/core/react/src/__tests__/WalletProviderMobile-test.tsx
```diff
@@ -110,6 +114,12 @@ describe('WalletProvider when the environment is `MOBILE_WEB`', () => {
     beforeEach(() => {
         localStorage.clear();
         jest.clearAllMocks().resetModules();
+        rpcEndpoint = 'https://api.mainnet-beta.solana.com';
+        jest.mocked(useConnection).mockImplementation(() => ({
+            connection: {
+                rpcEndpoint,
+            } as Connection,
+        }));
```
jordansexton, line 123: Worth testing the case when there's no ConnectionProvider?

packages/core/react/src/WalletProvider.tsx
jordansexton: While I enjoy the vibe of `autoConnect_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`, does the warning really matter? This is a nonstandard method that other adapters won't have, but would it matter if a dapp wanted to call it directly?

jordansexton: Also, there's something a bit weird about a method `handleAutoConnect` that... does not handle auto connect, but instead returns a bound connect function?

packages/wallets/unsafe-burner/src/adapter.ts
```diff
@@ -41,11 +42,12 @@ export class UnsafeBurnerWalletAdapter extends BaseSignerWalletAdapter {
     }
 
     get readyState() {
-        return WalletReadyState.Installed;
+        return this._readyState;
     }
 
     async connect(): Promise<void> {
         this._keypair = new Keypair();
+        this._readyState = WalletReadyState.Installed;
```
jordansexton, line 50: Not that it matters too much, but changing this shouldn't be needed. Loadable adapters typically just stay marked loadable even after they are loaded. We also aren't firing a `readyStateChange` event when it changes, so the React context won't find out about it.

jordansexton:
Looks great! I reviewed each commit in order, I'll now review the whole change together.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -23,285 +24,158 @@ export interface WalletProviderProps {
...
-    const [wallets, setWallets] = useState(() =>
-        adapters.map((adapter) => ({
-            adapter,
-            readyState: adapter.readyState,
-        }))
-    );
+function getInferredClusterFromEndpoint(endpoint?: string): Cluster {
+    if (endpoint == null) {
+        return 'mainnet-beta';
+    }
```
jordansexton, line 38: nit: can probably just do `!endpoint` in the odd case an empty string is provided.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -23,285 +24,158 @@ export interface WalletProviderProps {
...
-        function listener() {
-            isUnloading.current = true;
+        return new SolanaMobileWalletAdapter({
+            addressSelector: createDefaultAddressSelector(),
+            appIdentity: {
+                uri: getUriForAppIdentity(),
+            },
+            authorizationResultCache: createDefaultAuthorizationResultCache(),
+            cluster: getInferredClusterFromEndpoint(connection?.rpcEndpoint),
+        });
```
jordansexton, line 85: This all makes sense, though I do wonder if maybe we want to provide an easy way for dapps to customize their MWA without instantiating it (name and icon, mostly) and the maybe add a `cluster` prop to `ConnectionProvider` which wouldn't hurt in general to have here, or for the Wallet Standard.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -23,285 +24,158 @@ export interface WalletProviderProps {
...
-        (error: WalletError) => {
-            // Call onError unless the window is unloading
-            if (!isUnloading.current) (onError || console.error)(error);
-            return error;
-        },
-        [isUnloading, onError]
+        if (adapters.indexOf(mobileWalletAdapter as Adapter) !== -1) {
+            return adapters;
+        }
+        return [mobileWalletAdapter, ...adapters] as Adapter[];
```
jordansexton, line 94: nit: narrowly cast only MWA since we know it's not `null` from `getIsMobile`
```suggestion
        if (adapters.indexOf(mobileWalletAdapter as Adapter) !== -1) {
            return adapters;
        }
        return [mobileWalletAdapter as Adapter, ...adapters];
```

packages/core/react/src/WalletProvider.tsx
```diff
@@ -23,285 +24,158 @@ export interface WalletProviderProps {
...
+    const isUnloading = useRef(false);
     useEffect(() => {
-        if (
-            isConnecting.current ||
-            connected ||
-            !autoConnect ||
-            !adapter ||
-            !(readyState === WalletReadyState.Installed || readyState === WalletReadyState.Loadable)
-        )
+        if (walletName === SolanaMobileWalletAdapterWalletName && getIsMobile(adapters)) {
```
jordansexton, line 151: Is it worth memoizing `getIsMobile(adapters)` so effects etc. don't happen as much?

packages/core/react/src/WalletProvider.tsx
```diff
@@ -23,285 +24,158 @@ export interface WalletProviderProps {
...
+         * distinguish between a disconnection event received because a user initiated it, and one
+         * that was received because they've closed the window, we have to track window unload
+         * events themselves. Downstream components use this information to decide whether to act
+         * upon or drop wallet events and errors.
+         */
+        window.addEventListener('beforeunload', handleBeforeUnload);
+        return () => {
+            window.removeEventListener('beforeunload', handleBeforeUnload);
+        };
+    }, [adapters, walletName]);
```
jordansexton, line 169: Are there other wallets besides Phantom that we should request to fix this behavior so we can remove this hack that hurts cache perf?

packages/core/react/src/WalletProviderBase.tsx
```diff
@@ -0,0 +1,269 @@
...
+                adapter.off('disconnect', handleWalletDisconnectEvent);
+                adapter.off('error', handleError);
+                isConnecting.current = false;
+                setConnecting(false);
+                setConnected(false);
+                isDisconnecting.current = false;
+                setDisconnecting(false);
+                setPublicKey(null);
+            };
+        }
```
jordansexton, line 131: The original intent of combining certain state setters and only calling them when needed was to reduce the number of state updates. Does that matter, and in what circumstances will these not be correctly handled by the other places that handle this setup/teardown?

jordansexton:
Looks great. Mostly just have nonblocking clarifying questions/comments. This should definitely be merged without squashing to preserve the excellent commit history though!

packages/core/react/src/WalletProviderBase.tsx
jordansexton: Cool, I like the approach. Once this is merged, we can just lightly change https://github.com/solana-labs/wallet-adapter/pull/528 to add the standard adapters to `WalletProvider`. If any maniacs want to use `WalletProviderBase`, they can get it unadulterated with sane defaults.

packages/core/react/src/WalletProvider.tsx
steveluscher: Done! I also moved all of this code into a module and redid the tests with a mock of that module.

packages/core/react/src/WalletProvider.tsx
steveluscher: > In the event that a dapp passes a connection arg to sendTransaction that is not the one from useConnection, what should happen?

I think you're hosed in this case. If that connection is for a different cluster than the cluster for which you've sought an MWA authorization (ie. an `auth_token`) then it's just not gonna work. You can supply different endpoints all the live-long day, so long as they resolve to the same `Cluster` though.

Folks who really need to go ham with customization here should supply their own `new SolanaMobileWalletAdapter({/* ... */})`.

> …if the dapp is configured without ConnectionProvider, should we do anything here?

In that case then `connection?.rpcEndpoint` will be `undefined` and the inference module will just return `mainnet-beta` all the time.

packages/core/react/src/__tests__/WalletProviderMobile-test.tsx
steveluscher: Maybe? The default implementation of the `ConnectionProvider` when not supplied is `{}` which will produce `connection?.rpcEndpoint === undefined`, so I'm not too worried.

packages/core/react/src/WalletProvider.tsx
steveluscher: > …but would it matter if a dapp wanted to call it directly?

I say yes, because I want to be able to make this method go away one day, once `connect({silent: true})` is a thing. Me adding the suffix to this – a public method – is just a way of capturing plausible deniability when I rug it on someone who decided to rely on it anyway.

> …there's something a bit weird about a method handleAutoConnect that... does not handle auto connect…

Yeah, this is me fitting a pattern where:

* The _prop_ that a component takes for an event handler must be called `onFoo`
* The _function_ that gets called when that event happens must be called `handleFoo`

From the perspective of this code:

* `onAutoConnect` is what `WalletProviderBase` calls when it decides it wants to trigger an auto-connection. This is analagous to a `<button>` firing `onChange`
* `handleAutoConnect` is the function that `WalletProvider` calls to handle that trigger.

To get really pedantic I could call them `onAutoConnectRequest` and `handleAutoConnectRequest`.

steveluscher: I'll do that.

packages/core/react/src/WalletProvider.tsx
steveluscher: Well, here's the thing. This is a bit clowny, because `getIsMobile` isn't actually a dependency of any of the effects. If, for instance, the user-agent changed during the life of the webpage (???) React would not re-render.

The only useful way to `useMemo()` `getIsMobile` is to have it fire some sort of event when the user-agent changes, which seems like mega YAGNI territory.

I'm making the assumption here that `getIsMobile` is a pure function of `adapters`.

packages/core/react/src/WalletProvider.tsx
steveluscher: Super good question. We could do an audit, but we could also just wait until every wallet supports the standard and this doesn't matter, right? Or does the standard not help with this?

packages/wallets/unsafe-burner/src/adapter.ts
steveluscher: Yeah. I can't remember why I did this. I'll revert it and maybe we'll find out.

steveluscher:
@jordansexton: Changes since your review: https://github.com/solana-labs/wallet-adapter/compare/4f033c16ee2cbff207acd26faae40b769888419d..9cd5afe715d30461a06be4d07acfb52d5dc22b7b.

packages/core/react/src/WalletProvider.tsx
steveluscher: I'm hesitant to add things (at all, but particularly) to `ConnectionProvider` that only work for one platform. Like, adding `icon` doesn't _do_ anything on desktop, which might be confusing for folks.

Reminds me though, I wanted to suggest that we conventionally look for an icon at the root of the site if one isn't supplied; favicon style.

packages/core/react/src/WalletProviderBase.tsx
steveluscher: Oh, that's not a thing in React anymore. React batches state updates by default. You can call `setFoo` 1000x in a row and you'll get 1 re-render.

packages/core/react/src/WalletProvider.tsx
sdlaver: The MWA spec doesn't strictly define the values for `cluster` for this reason: https://solana-mobile.github.io/mobile-wallet-adapter/spec/spec.html#authorize

IMO, the right thing to do here would be to send the RPC URI in the `cluster` parameter. If the wallet supports interacting with arbitrary clusters, it should record that RPC URI and return a valid auth token. If it does not support arbitrary cluster, it can return an error instead.

jordansexton: > IMO, the right thing to do here would be to send the RPC URI in the cluster parameter.

Hmm. It may be a security concern for a wallet to trust an RPC URL from a dapp. It can't tell for sure what cluster you are signing with that way, or the RPC can lie about its genesis hash or anything else if you tried.

packages/core/react/src/WalletProvider.tsx
jordansexton: Ah, what I meant is, adding optional cluster to ConnectionProvider would actually be _extremely_ useful for the Wallet Standard as well (which also infers it, messily). So it would do something useful on mobile and desktop.

packages/core/react/src/WalletProviderBase.tsx
jordansexton: Got it. Thanks for clarifying, you have much better knowledge of React behaviors than me.

packages/core/react/src/WalletProvider.tsx
steveluscher: Interesting. Well, let's land both and then we can figure out how to patch them both up at once, if at all. I think I'm pro:

```typescript
type ClusterfulEndpoint = Readonly<{
  cluster: Cluster,
  url: string,
}>;

type ConnectionProviderProps = Readonly<{
  endpoint: string | ClusterfulEndpoint,
  /* ... */
}>;
```

Something like that.

jordansexton: lol @ `Clusterful`

packages/core/react/src/WalletProvider.tsx
jordansexton: Hmm, the standard doesn't help with this _yet_ because when used as a wrapper, it's still just downstream of `window.phantom` or whatever is firing disconnect events. I think we fix it in Phantom, make sure that Glow, Solflare, Backpack aren't affected, then we burn it (tracking issue for completion https://github.com/solana-labs/wallet-adapter/issues/593)

PR #591, opened 2022-10-11
Updates dependency version for package @fractalwagmi/solana-wallet-adapter in fractal wallet package

obber:
This PR adds some handling around when a user denies a wallet connection or when they explicitly close the popup without giving approval.


jordansexton (commented):
LGTM

PR #595, opened 2022-10-15
Open ecosystem page when no mobile wallets respond

steveluscher:
#### Problem statement

There is no way to interrogate the device, from JavaScript, to know whether it will respond to the protocol `solana-wallet://` or not. You have to try it, and presume that it doesn't unless the browser backgrounds.

If the browser hasn't backgrounded within a reasonable amount of time, then we assume there are no wallets installed. At this point we want to show the wallet ecosystem page so that the user can choose a wallet to install.

#### Summary of changes

* If the protocol throws `ERROR_WALLET_NOT_FOUND` then redirect the webpage to the URL set in the `SolanaMobileWalletAdapter`.

#### Test plan
https://user-images.githubusercontent.com/13243/195958935-e75ccae8-a198-44af-8237-60acadbd3ae3.mp4

#### Notes

This is implemented as a navigation to `adapter.url` and not a popup window. This is because browsers will block any popup window that isn't directly connected to a user-initiated gesture like a tap. Between the time the user taps the connect button and the time that the wallet adapter throws `ERROR_WALLET_NOT_FOUND`, that chain has been broken long enough for the browser's popup blocker to engage.

### DIFF
--- a/packages/core/react/__mocks__/@solana-mobile/wallet-adapter-mobile.ts
+++ b/packages/core/react/__mocks__/@solana-mobile/wallet-adapter-mobile.ts
@@ -4,6 +4,7 @@ import { MockWalletAdapter } from '../../src/__mocks__/MockWalletAdapter.js';
 export const SolanaMobileWalletAdapterWalletName = 'Solana Mobile Wallet Adapter Name For Tests';
 export const createDefaultAddressSelector = jest.fn();
 export const createDefaultAuthorizationResultCache = jest.fn();
+export const createDefaultWalletNotFoundHandler = jest.fn();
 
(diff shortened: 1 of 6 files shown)


steveluscher:
So this is one way of doing it. The other way of doing this is to add a new piece of configuration to the plugin, like this:

```typescript
new SolanaMobileWalletAdapter({
  /* ... usual stuff ... */
  onWalletNotFound() { ... },
});
```

This would allow dApp developers to _override_ the `onWalletNotFound` behavior if they don't like it redirecting their entire site to another URL.

steveluscher:
OK, I did it the ‘other way.’ You can now supply an `onWalletNotFound` config through a custom adapter if you want to override the default behavior.

PR #605, opened 2022-10-17
Get cluster from connection if possible

jordansexton:


packages/core/react/src/__tests__/getInferredClusterFromEndpoint-test.ts
```diff
@@ -1,34 +1,34 @@
-import getInferredClusterFromEndpoint from '../getInferredClusterFromEndpoint.js';
+import getClusterFromEndpoint from '../getClusterFromEndpoint';
 
 describe('getInferredClusterFromEndpoint()', () => {
```
steveluscher, line 3:
```suggestion
describe('getClusterFromEndpoint()', () => {
```

steveluscher (changes requested):
If we go through with the rename, can you rewind and `git mv packages/core/react/src/getInferredClusterFromEndpoint.ts packages/core/react/src/getClusterFromEndpoint.ts` to preserve the blame through the rename?

packages/core/react/src/getClusterFromEndpoint.ts
```diff
@@ -0,0 +1,12 @@
+import type { Connection, Cluster } from '@solana/web3.js';
+
+export default function getClusterFromEndpoint(connection?: Connection): Cluster {
+    if (!connection) return 'mainnet-beta';
+    // HACK: remove casts when https://github.com/solana-labs/solana/pull/28435 lands
+    if ((connection as any).cluster) return (connection as any).cluster;
+    if (/devnet/i.test(connection.rpcEndpoint)) return 'devnet';
+    if (/testnet/i.test(connection.rpcEndpoint)) return 'testnet';
+    // HACK: remove casts when https://github.com/solana-labs/solana/pull/28435 lands
+    if (/localhost/i.test(connection.rpcEndpoint)) return 'localnet' as Cluster;
```
steveluscher, line 10: This is an error from the perspective of the Mobile Wallet Adapter protocol. `localnet` is not a cluster for which you can seek an authorization, so this will always fail.

steveluscher (changes requested):

packages/core/react/src/getClusterFromEndpoint.ts
jordansexton: Right, that was a question I had. What should happen if a dapp does this?

steveluscher: It'll throw `ERROR_CLUSTER_NOT_SUPPORTED` from the protocol implementation. https://solana-mobile.github.io/mobile-wallet-adapter/spec/spec.html#errors

jordansexton: We could say "you said you want to use localhost, don't even add MWA by default since it won't work"

steveluscher: That makes sense. Short circuit that in `WalletAdapter` to not add the MWA in the first place. Land this and I can add that as a follow up!

Otherwise the dApp should just die. Until such time as you can run a validator on your phone, there's no sense in authorizing for `localnet` in a mobile wallet.

steveluscher (approved):

packages/core/react/src/getClusterFromEndpoint.ts
jordansexton: Yeah, makes sense. Will do, and make the other changes. Open question is whether the PR into web3.js is even a good idea, so maybe worth deciding that first.

steveluscher: Replied on that PR!

steveluscher (commented):
Have to also update the `jest.mock()` call in packages/core/react/src/__tests__/WalletProviderMobile-test.tsx.

packages/core/react/src/getClusterFromEndpoint.ts
```diff
@@ -0,0 +1,12 @@
+import type { Connection, Cluster } from '@solana/web3.js';
+
+export default function getClusterFromEndpoint(connection?: Connection): Cluster {
```
steveluscher, line 3: This has to be called `getClusterFromConnection` now.

PR #608, opened 2022-10-17
Restore ‘unset when fails to connect’ behavior

steveluscher:
@jordansexton pointed out the case where:

* You're using `MultiConnectButton`
* You start a connection attempt
* The user denies it

Now you're in a situation where you _can't_ deselect the wallet because of the structure of `MultiConnectButton`. There's no UI to do so.

This PR restores the ‘unset the wallet when encountering an error’ paradigm, but only does so if the error encountered is:

* not with `SolanaMobileWalletAdapter`
* a `WalletConnectionError`

#### Test plan

https://user-images.githubusercontent.com/13243/196283272-1cb8d76e-cb29-4e5b-81d4-8b5d90f08ce9.mov

### DIFF
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -9,7 +9,7 @@ import type { Adapter, WalletError, WalletName } from '@solana/wallet-adapter-ba
 import { useStandardWalletAdapters } from '@solana/wallet-standard-wallet-adapter-react';
 import type { Cluster } from '@solana/web3.js';
 import type { ReactNode } from 'react';
-import React, { useEffect, useMemo, useRef } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef } from 'react';
 import getClusterFromConnection from './getClusterFromConnection.js';
 import getEnvironment, { Environment } from './getEnvironment.js';
 import { useConnection } from './useConnection.js';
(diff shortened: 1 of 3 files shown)


packages/core/react/src/WalletProvider.tsx
```diff
@@ -157,13 +159,27 @@ export function WalletProvider({
             window.removeEventListener('beforeunload', handleBeforeUnload);
         };
     }, [adaptersWithStandardAdapters, walletName]);
+    const handleError = useCallback(
+        (walletError: WalletError) => {
+            (onError || console.error)(walletError);
+            if (
+                adapter &&
+                adapter.name !== SolanaMobileWalletAdapterWalletName &&
+                walletError.name === 'WalletConnectionError'
```
jordansexton, line 168: Hmm this is too strict, a number of error types can be thrown by `connect` (account error, pubkey error, etc.) and any of them should trigger this.

The way we have the code in master, it's thrown whenever an error is encountered while connecting:
https://github.com/solana-labs/wallet-adapter/blob/afdc0b432150511d560a32aec1dcdcbdb8f7fbd3/packages/core/react/src/WalletProvider.tsx#L175-L181

Tricky here, because we want to special case MWA, and don't have access to `connecting` ref.

steveluscher: Yeah, exactly. Really awkward to _directly_ wrap `connect()` the way the responsibilities are split here.

Only option I can think of is to invent a new `onConnectError` prop for `WalletProviderBase` to catch all manner of connection errors at the callsite.

steveluscher: I guess I could also plumb through the `isConnecting` ref, instead. Hmm. Which is worse…

PR #619, opened 2022-10-21
feat: Add Infinity Wallet

InfinityWallet:
This PR adds the wallet adapter for the Infinity Wallet, allowing our users to easily connect to Solana DApps via our unique  Web3 Browser (removing the need to use a browser-extension wallet) on desktop.

We previously made a PR in August and completed all changes to comply with the standard, however after spending 2 months making the requested changes we then found that the PR had been suddenly closed.

@jordansexton mentioned it was closed due to missing changes still, yet the only thing not changed was a name change from InfinityWallet to Infinity that he had requested. That is not our brand name however as was mentioned multiple times. He also stated Infinity Wallet is a browser-extension, yet as also explained, it is not.

Hopefully someone can rectify this and swiftly review the PR so that our users and partners do not have to wait any longer to use Solana DApps in our Web3 Browser.

PR #620, opened 2022-10-22
Add Infinity Wallet

InfinityWallet:
This PR adds the wallet adapter for the Infinity Wallet, allowing our users to easily connect to Solana DApps via our unique Web3 Browser (removing the need to use a browser-extension wallet) on desktop.

We previously made a PR in August and completed all changes to comply with the standard, however after spending 2 months making the requested changes we then found that the PR had been suddenly closed.

@jordansexton regarding the naming of Infinity it would be wrong as our brand is Infinity Wallet, the same way Math Wallet is named mathWallet in the repo.

The solana name adapter was integrated the exact same way as Phantom. You also mentioned all wallets should be equal and we had an unanswered question regarding this, however we can change it if wallet equality is not a goal of the repo?

All other requested changes including removing the url getter were done many weeks ago, so it would be nice if after all this time you could take a minute to review the completed changes for completion of this PR. Especially after wasting weeks of our teams time!

### DIFF
--- a/packages/wallets/infinityWallet/LICENSE
+++ b/packages/wallets/infinityWallet/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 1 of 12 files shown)


jordansexton:
Please read the comments on the previous issue you opened, which address all points here. https://github.com/solana-labs/wallet-adapter/pull/619

We are no longer accepting adapters for browser injected wallets. Your adapter will not be included in this repository or published in its NPM packages. You won't get anywhere by pressing this point. There's a clear path forward for you, which is to implement the Wallet Standard. Then you won't need an adapter at all, and you won't have to send more PRs. This is the last time I'll say this.

PR #621, opened 2022-10-22
Include adapter with errors

jordansexton:
In the app-global `onError` handler provided to the WalletProvider, there's no good way to get the current adapter because you're handling this above the context the provider creates. It's useful to have this information to be able to display information and better understand the error. This PR adds an optional `adapter` arg to the handler and passes it wherever available.


jordansexton:
Looking at this code:
```ts
if (!(readyState === WalletReadyState.Installed || readyState === WalletReadyState.Loadable)) {
    if (typeof window !== 'undefined') {
        window.open(adapter.url, '_blank');
    }
    throw handleError(new WalletNotReadyError(), adapter);
}
```

I think it might make sense to have the custom `onError` handler do this if provided.

steveluscher (changes requested):
Since we have custom error classes, let’s use ‘em!

```typescript
throw new WalletNotReadyError(adapter);
```

Catchers can pick that up as `e.adapter`.

steveluscher:
> I think it might make sense to have the custom `onError` handler do this if provided.

To maintain backward compatibility, would we then be like ‘if no `onError` then open window and throw, otherwise _just_ throw?’

jordansexton:
> Since we have custom error classes, let’s use ‘em!

I thought about that and even wrote code for this first, but it doesn't really matter -- this is a React context problem. If the adapter throws the error, you already have a reference to it if you're using the adapter directly.

jordansexton:
I also didn't want to update the base package and all wallets and other packages for this. It's better if we can fix it locally for the update we want people to consume now, IMO.

jordansexton:
> > I think it might make sense to have the custom `onError` handler do this if provided.
> 
> To maintain backward compatibility, would we then be like ‘if no `onError` then open window and throw, otherwise _just_ throw?’

Yeah exactly.

jordansexton:
I'm gonna YOLO this and try it out.

PR #626, opened 2022-10-25
Add support for Versioned transactions to functions in `useAnchorWallet`

nibbus0x:
PR adds support for signing versioned transactions using `useAnchorWallet` hook.


packages/core/react/src/useAnchorWallet.ts
```diff
@@ -1,11 +1,11 @@
...
+import type { PublicKey, Transaction, VersionedTransaction } from '@solana/web3.js';
 import { useMemo } from 'react';
 import { useWallet } from './useWallet.js';
 
 export interface AnchorWallet {
     publicKey: PublicKey;
-    signTransaction(transaction: Transaction): Promise<Transaction>;
-    signAllTransactions(transactions: Transaction[]): Promise<Transaction[]>;
+    signTransaction(transaction: Transaction | VersionedTransaction): Promise<Transaction | VersionedTransaction>;
+    signAllTransactions(transactions: (Transaction | VersionedTransaction)[]): Promise<(Transaction | VersionedTransaction)[]>;
```
jordansexton, line 8: This will likely need to use generics -- the type of the transactions returned should match the type of the inputs.

packages/starter/example/src/components/SendLegacyTransactionWithAnchorWallet.tsx
```diff
@@ -0,0 +1,57 @@
...
+            return;
+        }
+    }, [wallet, notify, connection]);
+
+    return (
+        <Button variant="contained" color="secondary" onClick={onClick} disabled={!wallet?.publicKey}>
+            Send Anchor Friendly Legacy Transaction (devnet)
+        </Button>
+    );
+};
```
jordansexton, line 57: Please remove all these additional components. We don't need separate actions for all these, Anchor transactions are just transactions.

jordansexton (changes requested):
Thanks! Please update to use generics for the transaction types and only change the hook.

packages/core/react/src/useAnchorWallet.ts
nibbus0x: I've updated this, but not sure if it is what you meant. I'm not very familiar with using generics in typescript.

nibbus0x:
Thanks for the quick reply! I have made the changes. Let me know if there are any other changes necessary, or if you need me to squash commits.

jordansexton:
Perfect, thanks! Will release today.

jordansexton:
Actually, before I release -- what's the state of versioned transaction support in Anchor's TS declarations? Want to make sure we won't get type errors in apps that use this.

nibbus0x:
I'm not entirely sure what you mean? Are you asking about support of versioned transaction in the Anchor npm package?

jordansexton:
Yes. If we publish these types in `@solana/wallet-adapter-base`, they need to agree with the types expected in Anchor's TypeScript package(s). If they don't, devs will get unresolvable errors and be forced to pin previous versions etc.

packages/core/react/src/useAnchorWallet.ts
jordansexton: You did it right!

nibbus0x:
Doesn't seem like Anchor TS package has any support for versioned txs or lookup tables yet. They still use version `^1.36.0` of web3.js. I'll see if I can talk to them about it. The anchor TS package hasn't been updated in months afaik.

jordansexton:
Got it. I've reverted this change for now in #628. It will be necessary for Anchor to support this typed interface before we can.

nibbus0x:
Ok, it seems there is some talk/issues raised in the Anchor gh. Seems they are working on it. I'll try to keep you updated.

jordansexton:
Sounds good! Thanks for working on this.

nibbus0x:
Anchor has finally released an update/new package `@coral-xyz/anchor` that replaces `@porject-serum/anchor`. The new package depends on a more recent version of `@solana/web3.js` that supports versioned transactions. 

I believe this means that versioned transactions are now supported in the ANchor packages.

Relevant discussion if you are intereseted @jordansexton [Anchor tx V2 support](https://github.com/coral-xyz/anchor/issues/2220)

urkes:
> Anchor has finally released an update/new package `@coral-xyz/anchor` that replaces `@porject-serum/anchor`. The new package depends on a more recent version of `@solana/web3.js` that supports versioned transactions.
> 
> I believe this means that versioned transactions are now supported in the ANchor packages.
> 
> Relevant discussion if you are intereseted @jordansexton [Anchor tx V2 support](https://github.com/coral-xyz/anchor/issues/2220)

Any updates regarding this issue? @jordansexton Seems like @nibbus0x solution can be merged now.

PR #627, opened 2022-10-26
Added a React App with only Solana Phantom wallet connection Only JS

Arcturus91:
### Added a create-react-app-starter but modified only for JS users

1. **It is only JS in React files** App and Index
2. **Implemented code for connecting your Phantom Wallet** with your own Address (not random one like before)


jordansexton:
Thanks for your contribution. We won't be adding any more starter projects at this time. Also, your PR links to a git submodule, making it impossible to review here.

Issue #636, opened 2022-10-29
Updating `WalletProvider.onError` prop incorrectly disconnects the adapter


zpalmtree:
**Describe the bug**
I am using the wallet adapter in React, with Phantom wallet.
I am passing in the `endpoint` value as a prop.

I recently upgraded from 0.9.5 to 0.9.18, and now, when I change/set the RPC value in react, the wallet disconnects.

I tried to narrow it down to a more specific version to find the commit that introduced the bug but I couldn't get npm to install the older versions without using my old lock file.

If I downgrade back to 0.9.5 and my old lock file, changing the RPC lets the wallet remain connected.

Some additional things I noticed if it helps:

On the working version, it looks like the wallet state just changes once. In the current version, I see it change state twice after swapping RPC.

Possibly helpful state values after changing rpc:

0.9.5:
```
autoConnect: true
connected: true
connecting: false
disconnecting: false
readyState: "Installed'
publicKey: defined
```

0.9.18:

```
autoConnect: true
connected: false
connecting: false
disconnecting: false
readyState: 'Installed'
publicKey: null
```

**To Reproduce**
1. Setup simple react ui app doing something like logging the users public key
2. Use react variable for endpoint prop
3. Change endpoint on action, e.g. clicking a button
4. Wallet will disconnect and public key is no longer available for logging

**Expected behavior**
Wallet stays connected if previously connected

**Screenshots**

https://user-images.githubusercontent.com/22151537/198815343-205bcad1-7523-4ca7-ac7e-25a138efdd57.mp4

**Desktop (please complete the following information):**
 - OS: gahnoo loonix
 - Browser: Firefox
 - Version: 106.0.1

jordansexton:
Thanks so much for the report, can you share which other wallet adapter package versions are in your package.json?

jordansexton:
Or maybe the lockfile, since the resolved versions could be higher?

jordansexton:
And, if your project is open source or you have a reproduction repo, or a link to the running app, that helps too!

Specifically I'm first trying to first isolate of this issue is caused by something in the base package or the react package.

jordansexton:
As for why this may be happening, some example code --

```tsx
    const [network, setNetwork] = useState(WalletAdapterNetwork.Mainnet);
    const endpoint = useMemo(() => clusterApiUrl(network), [network]);
    const wallets = useMemo(
        () => [
            new PhantomWalletAdapter(),
[...]

PR #637, opened 2022-10-29
Don't reset state when `adapter` or `onError` changes

jordansexton:
Fixes https://github.com/solana-labs/wallet-adapter/issues/636

Now that we're not wiping the state when the adapter changes, we'll need to check the logic in all the other places this should happen.

### DIFF
--- a/packages/core/react/__mocks__/@solana-mobile/wallet-adapter-mobile.ts
+++ b/packages/core/react/__mocks__/@solana-mobile/wallet-adapter-mobile.ts
@@ -1,4 +1,4 @@
-import type { WalletName } from '@solana/wallet-adapter-base';
+import { type WalletName } from '@solana/wallet-adapter-base';
 import { MockWalletAdapter } from '../../src/__mocks__/MockWalletAdapter.js';
 
 export const SolanaMobileWalletAdapterWalletName = 'Solana Mobile Wallet Adapter Name For Tests';
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -9,8 +9,8 @@ import type { Adapter, WalletError, WalletName } from '@solana/wallet-adapter-ba
 import { useStandardWalletAdapters } from '@solana/wallet-standard-wallet-adapter-react';
 import type { ReactNode } from 'react';
 import React, { useCallback, useEffect, useMemo, useRef } from 'react';
-import getInferredClusterFromEndpoint from './getInferredClusterFromEndpoint.js';
 import getEnvironment, { Environment } from './getEnvironment.js';
+import getInferredClusterFromEndpoint from './getInferredClusterFromEndpoint.js';
 import { useConnection } from './useConnection.js';
 import { useLocalStorage } from './useLocalStorage.js';
 import { WalletProviderBase } from './WalletProviderBase.js';
@@ -19,8 +19,8 @@ export interface WalletProviderProps {
     children: ReactNode;
     wallets: Adapter[];
     autoConnect?: boolean;
-    onError?: (error: WalletError, adapter?: Adapter) => void;
     localStorageKey?: string;
+    onError?: (error: WalletError, adapter?: Adapter) => void;
 }
 
 let _userAgent: string | null;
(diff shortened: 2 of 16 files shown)


steveluscher:
Instead, let's try to store the actual error handler in a ref, update that ref in an effect, so then we can remove the handler as a dep of `handleFoo`

steveluscher (changes requested):
"dependencies" amirite?

steveluscher:
`handleError` won't even need to `useCallback` after you've done this since both of the things it uses will be refs.

jordansexton:
Yeah, sounds good. Will update and test.

steveluscher:
Wrote a failing test for you: 2c17fc69e6934588cb4491d5bed1017571f9ae0b.

steveluscher:
Ooh! This is even better. When you fix the bug, you'll find that _another_ test breaks, pointing to a _separate_ real problem that this was masking.

I don't want to steal it from you, but please do let me know if you want me to take this over.

jordansexton:
Yeah, turned out we need that state reset when the adapter changes, at minimum to wipe out the `publicKey`. I refactored a bit and did the ref thing. I went ahead and merged to get a fixed release out but if it doesn't look ideal, please revise!

packages/core/react/__mocks__/@solana-mobile/wallet-adapter-mobile.ts
```diff
@@ -1,4 +1,4 @@
-import type { WalletName } from '@solana/wallet-adapter-base';
+import { type WalletName } from '@solana/wallet-adapter-base';
```
steveluscher, line 1: 🧐

packages/core/react/src/WalletProvider.tsx
```diff
@@ -160,13 +160,15 @@ export function WalletProvider({
     }, [adapter, setWalletName]);
     return (
         <WalletProviderBase
-            {...props}
```
steveluscher, line 163: y tho? I did this so that any additions to WalletProvider would get caught as TS errors if they were forgotten in Base.

packages/core/react/src/WalletProviderBase.tsx
```diff
@@ -8,54 +8,58 @@ import type {
...
 export function WalletProviderBase({
-    adapter,
     children,
+    wallets: adapters,
+    adapter,
     isUnloadingRef,
     onAutoConnectRequest,
     onConnectError,
-    onError,
+    onError = (error, adapter) => {
```
steveluscher, line 36: I _think_ this will result in a new copy of this function on every render, which will thrash the effect.

packages/core/react/src/WalletProviderBase.tsx
```diff
@@ -8,54 +8,58 @@ import type {
...
+        if (!isUnloadingRef.current) {
+            onError(error, adapter);
+        }
+        return error;
+    });
+    useEffect(() => {
+        handleErrorRef.current = (error, adapter) => {
             if (!isUnloadingRef.current) {
-                if (onError) {
-                    onError(error, adapter);
```
steveluscher, line 47: If you make the call to `handleErrorRef` here, all of the rest of the code gets way simpler.

packages/core/react/__mocks__/@solana-mobile/wallet-adapter-mobile.ts
jordansexton: It was getting annoying to have two imports for the same dependencies everywhere. 😅

packages/core/react/src/WalletProviderBase.tsx
jordansexton: Ahh, you're probably right.

packages/core/react/src/WalletProvider.tsx
jordansexton: Yeah, I could see that, but it makes things a bit more complicated for the sake of something that almost never happens. When do we add props to the WalletProvider, and how likely are we to forget to use them if we need them?

packages/core/react/__mocks__/@solana-mobile/wallet-adapter-mobile.ts
jordansexton: It was getting annoying to have two imports for the same dependencies everywhere. 😅

packages/core/react/src/WalletProviderBase.tsx
jordansexton: Good point, I wasn't happy with how I did this.

PR #640, opened 2022-10-30
Optimize `onError`

jordansexton:
This addresses https://github.com/solana-labs/wallet-adapter/pull/637#discussion_r1008769669

Per https://github.com/solana-labs/wallet-adapter/pull/637#discussion_r1008769819 there may be a better way of doing this

### DIFF
--- a/.changeset/ninety-dolphins-change.md
+++ b/.changeset/ninety-dolphins-change.md
@@ -1,0 +1,5 @@
+---
+'@solana/wallet-adapter-react': patch
+---
+
+Optimize `WalletProvider.onError`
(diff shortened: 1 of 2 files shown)


steveluscher:
I like this slightly better, since you don't have to build up the error handler implementation in two places. Squash and merge if you agree!

jordansexton:
Haha I tried it this way too. There's gotta be a better way to do default prop values of functions than passing a ref into a ref right?

jordansexton:
I'll add a changeset so we can release it from here.

steveluscher:
I mean,
1. The great thing about refs is they don't get ‘passed’ into anything. Even if you use them as a dependency, they will _never change_. They are a wrapper around a value, and the wrapped is 100% stable.
2. The first ref wraps the custom upstream handler, and the second ref implements the actual handler, so they're usefully different.

PR #650, opened 2022-11-08
added i18n infrastructure, en-US and pt-BR translations

danicuki:


danicuki:
Is there anything else I need to do to have this PR be accepted?

steveluscher:
Hi there! Thank you for this contribution!

You are absolutely correct that i18n has been a missed opportunity in this library – a critical flaw. This PR builds a _particular_ i18n solution into the UI in such a way that:

1. Increases the size of the UI library, and
2. Forces everyone to download every language set, no matter which one they use.

Unfortunately, I can't accept such an architecture; it's more prescriptive and inefficient than it could be.

The solution is going to be to dismantle this library into style-and-structure agnostic pieces; a set of abstract components that implement the default functionality of the controls and a set of concrete UI controls that you can use as defaults, or supply your own. In all cases, strings should be taken in as props, so that everyone can use their own i18n infra.

I don't have a timeline for this work, but let's make [an issue to track it](https://github.com/solana-labs/wallet-adapter/issues/658).

flodef:
Hey Steve, 
Just read your reasons for closing this PR...
And while I totally agree on a theorical point of view, I fully disagree on a practical point of view.
1. How much is the size increase in that particular case ?
2. What is the size of every language downloaded (even if 100 languages added) ?

Before closing, have you really weighted the cost of closing versus the cost of accepting ?
Don't know if you speak Portuguese, but try imagining if the labels were in Portuguese (from the packages/ui/react-ui/src/WalletModalProvider.tsx file). Would you use such app ?

From a non-english speaker point of view, seeing all the english label is a stopper !!
I'm trying really hard to make non-english speaker onboard on Solana, and that's one of my main concern.
When someone see english label, they feel : "oh no ! This app is not for me." ==> BOOM, one more potential adopter lost!

Again, I totally agree the translation could be better, more optimize and all. But for the time being, letting the app "English-only" is like saying : "Solana is only available to English speaking users".
And it's making dev like me a real nightmare to try to circonvene all the English-only UI components. I have to rewrite some of them or find some trick to not display English label. And that's a pain + a waste of time.

Please, take some time to reconsider it, and try to ask non-english speaker their point of view on the matter.

danicuki:
Yes @flodef, this is the point I am trying to make over and over again. 

Only 1/8 of the world population speaks English. An internationalized interface multiply by 8 the potential userbase for Solana Wallets. 

We have to stop looking at problems only from the developer's perspective and think about real people. Having a pixel perfect implementation of this i18n feature is definitely no more important than onboarding billions of people to the Solana ecosystem.

vitormancio:
“ this PR is important for non English speakers and should be accepted”

reizilla:
this PR is important for non English speakers and should be accepted

viniblack:
I believe this PR will make it much easier for non-English speakers and should be accepted

jordansexton:
I've locked this issue because we don't need support brigading, or repetitive comments. Please take any ongoing discussion to the issue Steven created, https://github.com/solana-labs/wallet-adapter/issues/658. Only add to that issue if it's useful, constructive, and focused on the problem or solutions to it.

PR #653, opened 2022-11-08
Support for Versioned transaction in Phantom

0xproflupin:
PR for making Phantom adapter compatible with versioned transactions

### DIFF
--- a/packages/wallets/phantom/package.json
+++ b/packages/wallets/phantom/package.json
@@ -30,13 +30,13 @@
         "package": "shx mkdir -p lib/cjs && shx echo '{ \"type\": \"commonjs\" }' > lib/cjs/package.json"
     },
     "peerDependencies": {
-        "@solana/web3.js": "^1.58.0"
+        "@solana/web3.js": "^1.61.0"
     },
     "dependencies": {
         "@solana/wallet-adapter-base": "workspace:^"
     },
     "devDependencies": {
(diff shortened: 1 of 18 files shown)


packages/wallets/phantom/package.json
```diff
@@ -30,13 +30,13 @@
         "package": "shx mkdir -p lib/cjs && shx echo '{ \"type\": \"commonjs\" }' > lib/cjs/package.json"
     },
     "peerDependencies": {
-        "@solana/web3.js": "^1.58.0"
+        "@solana/web3.js": "^1.61.0"
```
jordansexton, line 33: Can you revert these version changes? 1.58.0 is used by all the packages now because it's the minimum version with vTX support.

jordansexton (changes requested):
Thanks for the quick PR. I think the types in the window.ts file will also need to be updated to reflect the Phantom extension's support for these.

Can you make sure to run the commands to build from source locally? I suspect right now there will be a type error because of this.

packages/wallets/phantom/package.json
0xproflupin: just reverted back

0xproflupin: > Can you make sure to run the commands to build from source locally? I suspect right now there will be a type error because of this.

You were right. Seems to build successfully now.

PR #659, opened 2022-11-15
Adds `signMessage` support for Fracatal wallet adapter

obber:

PR #661, opened 2022-11-17
Add signMessage method to Fractal adapter

jordansexton:
This PR supersedes https://github.com/solana-labs/wallet-adapter/pull/659


packages/wallets/fractal/package.json
```diff
@@ -33,11 +33,11 @@
         "@solana/web3.js": "^1.58.0"
     },
     "dependencies": {
-        "@fractalwagmi/solana-wallet-adapter": "^0.0.8",
+        "@fractalwagmi/solana-wallet-adapter": "0.0.9",
```
jordansexton, line 36: @obber should this be released as `0.1.0` so that it can be updated according to semver without changes to the adapter?

PR #665, opened 2022-11-30
Bumps fractal wallet adapter impl package

obber:
This fixes a bug with the new `signMessage` implementation where we were not correctly serializing the signature Uint8Array.

### DIFF
--- a/packages/wallets/fractal/package.json
+++ b/packages/wallets/fractal/package.json
@@ -33,7 +33,7 @@
         "@solana/web3.js": "^1.58.0"
     },
     "dependencies": {
-        "@fractalwagmi/solana-wallet-adapter": "0.0.12",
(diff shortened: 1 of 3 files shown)


packages/wallets/fractal/package.json
```diff
@@ -33,11 +33,11 @@
         "@solana/web3.js": "^1.58.0"
     },
     "dependencies": {
-        "@fractalwagmi/solana-wallet-adapter": "0.0.9",
+        "@fractalwagmi/solana-wallet-adapter": "0.0.12",
```
jordansexton, line 36: Can this get bumped up to a minor version, and use `^` for versioning here instead of an exact version? It would be great to not have to release this package, and the `wallets` package, every time there's an update to a dependency.

obber: Done.

jordansexton:
Thank you!

PR #673, opened 2022-12-12
Update Phantom adapter to redirect to Phantom browser on iOS

mcintyre94:
This PR updates the Phantom wallet-adapter so that users on iOS are redirected to the Phantom in-app browser when they click connect.

Eg app with autoconnect enabled, first visit:

https://user-images.githubusercontent.com/1711350/207105287-352bd0f2-3af2-474e-bc7e-b65020e18a37.MOV

Subsequent visit:

https://user-images.githubusercontent.com/1711350/207105271-ec1e825a-9713-48e5-9f36-b1d36b5e244b.MOV

This is intended to improve UX until we have an MWA solution for iOS

Currently AFAIK only Phantom has both an in-app browser and a universal link to direct the user to it. Solflare has universal links, but not yet one to navigate to its in-app browser. We will be able to use the same approach for any wallet with such a universal link though. 

- We use the user agent to detect iOS. Specifically we search for `iphone` or `ipad`. We also require `safari` to be present. In my testing it is present in iOS browsers, and is not in webviews. We don't want to redirect from eg. Phantom's webview, so we use the presence of `safari` to prevent that.
- There's a new `autoconnect` function in `WalletAdapterProps`. This is called in our React `WalletProvider` if autoconnect is enabled. It replaces the previous call to either `connect` or MWA's custom autoconnect function. By default it just calls connect, for MWA it's overridden with a custom autoconnect function (see https://github.com/solana-mobile/mobile-wallet-adapter/pull/341), and for Phantom it's overridden to be called only when we're not on iOS. This is because we can't automatically redirect to a universal link, it has to be a result of user action.
- I've also refactored `WalletModal` so that if there are no installed wallets but there are loadable ones, then the loadable ones are shown in the top list. When there are installed wallets the behaviour remains the same. This means that on iOS if a dapp has the Phantom adapter listed, it will be displayed in the top list and not collapsed.


packages/core/react/src/WalletProvider.tsx
```diff
@@ -109,12 +110,10 @@ export function WalletProvider({
...
         }
-        if (walletName === SolanaMobileWalletAdapterWalletName && getIsMobile(adaptersWithStandardAdapters)) {
-            return (adapter as SolanaMobileWalletAdapter).autoConnect_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.bind(adapter);
-        } else {
-            return adapter.connect.bind(adapter);
-        }
-    }, [adapter, adaptersWithStandardAdapters, autoConnect, walletName]);
+
+        // Not sure why this is needed, adapter is already WalletAdapter<string> | SolanaMobileWalletAdapter
+        return (adapter as WalletAdapter<string> | SolanaMobileWalletAdapter).autoConnect.bind(adapter);
```
jordansexton, line 115: Yeah, it does seem like this can just be `adapter.autoConnect.bind(adapter)` now, right?

packages/ui/react-ui/src/WalletModal.tsx
```diff
@@ -50,7 +51,7 @@ export const WalletModal: FC<WalletModalProps> = ({ className = '', container =
                   wallets.find((wallet: { readyState: any }) => wallet.readyState === WalletReadyState.Loadable) ||
                   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                   otherWallets[0]!;
-    }, [installedWallets, wallets, otherWallets]);
+    }, [installedWallets, wallets, loadableWallets, notDetectedWallets]);
```
jordansexton, line 54: We should get rid of the special handling of Phantom and Torus while we're here making this change.

packages/ui/react-ui/src/WalletModal.tsx
```diff
@@ -131,6 +132,19 @@ export const WalletModal: FC<WalletModalProps> = ({ className = '', container =
...
+    if (installedWallets.length) {
+        listedWallets = installedWallets;
+        collapsedWallets = [...loadableWallets, ...notDetectedWallets];
+    } else if (loadableWallets.length) {
+        listedWallets = loadableWallets;
+        collapsedWallets = notDetectedWallets;
+    } else {
+        collapsedWallets = notDetectedWallets;
+    }
+
```
jordansexton, line 147: Nice, this is much more clear and explicit.

packages/wallets/phantom/src/adapter.ts
```diff
@@ -75,21 +75,48 @@ export class PhantomWalletAdapter extends BaseMessageSignerWalletAdapter {
             ? WalletReadyState.Unsupported
             : WalletReadyState.NotDetected;
 
+    isIosAndRedirectable() {
```
jordansexton, line 78: Let's externalize this function. It doesn't access `this` so it shouldn't be a method (esp a public one). We could even move it to the base package, if we expect it to be reused by Solflare and other wallets.

packages/wallets/phantom/src/adapter.ts
```diff
@@ -75,21 +75,48 @@ export class PhantomWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+            } else {
+                scopePollingDetectionStrategy(() => {
+                    if (window.phantom?.solana?.isPhantom || window.solana?.isPhantom) {
+                        this._readyState = WalletReadyState.Installed;
+                        this.emit('readyStateChange', this._readyState);
+                        return true;
+                    }
+                    return false;
+                });
+            }
```
jordansexton, line 119: :+1:

packages/wallets/phantom/src/adapter.ts
```diff
@@ -109,40 +136,54 @@ export class PhantomWalletAdapter extends BaseMessageSignerWalletAdapter {
...
-            const wallet = window.phantom?.solana || window.solana!;
-
-            if (!wallet.isConnected) {
-                try {
-                    await wallet.connect();
-                } catch (error: any) {
-                    throw new WalletConnectionError(error?.message, error);
-                }
-            }
+    async connectWithWallet(): Promise<void> {
```
jordansexton, line 139: We shouldn't add new public methods like this. It should be `#private`, but the name is a bit confusing.

packages/wallets/phantom/src/adapter.ts
```diff
@@ -109,40 +136,54 @@ export class PhantomWalletAdapter extends BaseMessageSignerWalletAdapter {
...
-            this.emit('connect', publicKey);
+            if (this.readyState === WalletReadyState.Installed) {
+                await this.connectWithWallet();
+            } else if (this.readyState === WalletReadyState.Loadable) {
+                // redirect to the Phantom /browse universal link
+                // this will open the current URL in the Phantom in-wallet browser
+                const url = encodeURI(window.location.href);
+                const ref = encodeURI(window.location.origin);
+                const redirectUrl = `https://phantom.app/ul/browse/${url}?ref=${ref}`;
+                window.location.href = redirectUrl;
```
jordansexton, line 183: If we take this approach, we might as well extract this logic to a private method also. Or since this is very little code, we could just handle this case first, and early return, and then inline the stuff that was extracted to `connectWithWallet`.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -6,6 +6,7 @@ import {
     SolanaMobileWalletAdapterWalletName,
 } from '@solana-mobile/wallet-adapter-mobile';
 import { type Adapter, type WalletError, type WalletName } from '@solana/wallet-adapter-base';
+import type { WalletAdapter } from '@solana/wallet-adapter-base/src/adapter.js';
```
jordansexton, line 9:
```suggestion
import { type WalletAdapter } from '@solana/wallet-adapter-base';
```

However this import should be combined with the existing import above it. Make sure not to import from files in an external package.

jordansexton (changes requested):
Love the approach overall!

PR #685, opened 2022-12-16
Update Phantom iOS to redirect immediately on Select

### LINKED ISSUE #684: Phantom iOS connect requires an extra tap
**Describe the bug**

Opening this as an issue because there's some fiddly stateful logic to work through

In #673 we added a redirect to Phantom's wallet browser when connecting to Phantom in iOS safari

This change introduced an `autoConnect` function, which is called both when the wallet is first restored from localstorage and when a new wallet is selected, and just replaces existing logic to call MWA autoconnect or adapter `connect`. 

In Phantom this `autoConnect` is set to a no-op on iOS Safari, because when the page first loads we can't redirect to the Phantom universal link (nor would we want to, it'd make it impossible to disconnect from the dapp in Safari)

This means that we need to select Phantom in the wallet modal (this triggers the no-op), and then click the connect/multi button to trigger the redirect

We should be able to redirect on select, because this is a result of user action. But currently it's a no-op so we don't. 

**To Reproduce**

1. Open a dapp using the latest dependencies (eg a starter from this repo) in iOS Safari
2. Connect to Phantom

**Expected behavior**

Selecting the Phantom wallet should be sufficient to trigger the redirect

**Screenshots**

Current UX:

https://user-images.githubusercontent.com/1711350/208091596-8eb85722-b6c4-4e29-9cbc-906c5050979d.mov

**Additional context**

Basically the data flow going on here is:

On page load:

`WalletProvider`
- Get wallet name from localstorage (initial value of `walletName`)
- Set adapter from wallet name
`WalletProviderBase`
- Set wallet from adapter (prop)
- Run a `useEffect` when `wallet` changes which calls `autoConnect`

Here the behaviour is correct: we want a no-op for Phantom on iOS

On wallet select:
`WalletProvider`
- Update wallet name on select (call `setWalletName`)
- Set adapter from wallet name
`WalletProviderBase`
- Set wallet from adapter (prop)
- Run a `useEffect` when `wallet` changes which calls `autoConnect`

Here it's incorrect: we'd prefer to call `connect`

---

Considerations:

- The `autoConnect` function is by default just a call to `connect`. But Phantom and MWA have custom implementations. For MWA the current behaviour is correct in both cases and needs to stay the same. For Phantom it's incorrect only on wallet select
- There are no other calls to `setWalletName` except to set it to null

---

Suggested solutions:

- We could extend the `autoConnect` function call to pass an indication of whether the wallet has been selected automatically or by user action. Then the Phantom adapter could differentiate based on that. We could probably achieve this by having the `select` function set some ref that indicates a selection has been made. This would be a one way switch: once a select call has been made it'd be set to true and all future changes would be assumed to be by user select. This does rely on the fact that no other code sets another wallet elsewhere though. If we added any other code setting the wallet without user input it'd have to reset the ref. 
- We could try to remove some of the reactive code, and write an explicit behaviour on first connect then only runs once (when we have a wallet to connect to). And then write an explicit function for select that also connects, rather than being through this chain of updates. But a lot of code relies on the intermediate values (`adapter` and `wallet`), so we'd still have to maintain them in all cases. And I think this would probably be very difficult to do without breaking anything.

mcintyre94:
This PR updates the `autoConnect` logic for Phantom to be a no-op on iOS only until the user has made a wallet selection.

The `autoConnect` function now takes an optional param `hasUserSelectedAWallet`.

This defaults to false, and is set to true only in the `onSelectWallet` function.

The result is that we don't autoconnect Phantom when the page first loads on iOS, but we do immediately after it's selected in the modal.

I don't love this solution, but it is a small and easy piece of state to track. And it doesn't require reworking the data flow through `WalletAdapter` and `WalletAdapterBase` to `autoConnect` which I think would be a lot more complex and fragile. 

First visit:

https://user-images.githubusercontent.com/1711350/208174169-fb8e0d93-447d-4416-a394-805e8635a753.mov

Repeat visit:

https://user-images.githubusercontent.com/1711350/208174200-ad45cffa-9805-49f9-83ea-604f1cfc1159.mov

Closes #684

### DIFF
--- a/packages/core/react/src/__tests__/WalletProviderDesktop-test.tsx
+++ b/packages/core/react/src/__tests__/WalletProviderDesktop-test.tsx
@@ -214,6 +214,7 @@ describe('WalletProvider when the environment is `DESKTOP_WEB`', () => {
             it('does not call `autoConnect`', () => {
                 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                 const adapter = ref.current!.getWalletContextState().wallet!.adapter;
+                expect(adapter.connect).not.toHaveBeenCalled();
                 expect(adapter.autoConnect).not.toHaveBeenCalled();
             });
         });
@@ -221,10 +222,11 @@ describe('WalletProvider when the environment is `DESKTOP_WEB`', () => {
             beforeEach(() => {
                 renderTest({ autoConnect: true });
             });
-            it('calls `autoConnect`', () => {
+            it('calls `connect`', () => {
                 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                 const adapter = ref.current!.getWalletContextState().wallet!.adapter;
-                expect(adapter.autoConnect).toHaveBeenCalled();
+                expect(adapter.connect).toHaveBeenCalled();
+                expect(adapter.autoConnect).not.toHaveBeenCalled();
             });
         });
     });
--- a/.changeset/khaki-adults-search.md
+++ b/.changeset/khaki-adults-search.md
@@ -1,0 +1,5 @@
+---
+'@solana/wallet-adapter-react': patch
+---
+
+Only call autoConnect if the connect is not a user selection
--- a/packages/core/base/src/adapter.ts
+++ b/packages/core/base/src/adapter.ts
@@ -31,7 +31,7 @@ export interface WalletAdapterProps<Name extends string = string> {
     connected: boolean;
     supportedTransactionVersions?: SupportedTransactionVersions;
 
-    autoConnect(): Promise<void>;
+    autoConnect(hasUserSelectedAWallet: boolean | undefined): Promise<void>;
     connect(): Promise<void>;
     disconnect(): Promise<void>;
 
@@ -89,7 +89,7 @@ export abstract class BaseWalletAdapter<Name extends string = string>
         return !!this.publicKey;
     }
 
-    async autoConnect() {
+    async autoConnect(_hasUserSelectedAWallet: boolean) {
         await this.connect();
     }
 
--- a/packages/core/base/src/adapter.ts
+++ b/packages/core/base/src/adapter.ts
@@ -31,7 +31,7 @@ export interface WalletAdapterProps<Name extends string = string> {
     connected: boolean;
     supportedTransactionVersions?: SupportedTransactionVersions;
 
-    autoConnect(hasUserSelectedAWallet: boolean | undefined): Promise<void>;
+    autoConnect(): Promise<void>;
     connect(): Promise<void>;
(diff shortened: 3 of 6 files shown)


jordansexton:
Hmm. Thanks for taking a first pass at this!

Initial reaction: I don't like that this violates principle of least knowledge. The WalletAdapter interface has no notion of a "selected" wallet. This is purely a higher level (in terms of abstraction) concern, of the React (or whatever) libraries on top of it. Theoretically there's no reason you couldn't have multiple wallets "selected" (whatever that means, defined by this abstraction).

I imagined that it was possible to do this purely in the React context. I'll think about this a bit more, to see if that's true.

mcintyre94:
@jordansexton I've tweaked this PR to call `connect` instead of `autoConnect` after a user selection, instead of passing the value through to `autoConnect`. I expect this will require some test changes and I want to do more manual testing too, but would be good to hear your thoughts on this. In my initial testing it works well including with MWA. 

I think this is arguably a clearer API for adapters: if it's being connected without user input then it's `autoConnect`, if it's been selected then it's `connect`. It also keeps all the knowledge out of the adapter and in the `WalletProvider`, the adapter just knows which function has been called.

packages/core/react/src/__tests__/WalletProviderDesktop-test.tsx
```diff
@@ -224,7 +224,7 @@ describe('WalletProvider when the environment is `DESKTOP_WEB`', () => {
             it('calls `autoConnect`', () => {
                 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                 const adapter = ref.current!.getWalletContextState().wallet!.adapter;
-                expect(adapter.autoConnect).toHaveBeenCalled();
+                expect(adapter.connect).toHaveBeenCalled();
```
jordansexton, line 227: I'm not 100% sure about this, but does it make sense we test here (or any of the other places) for both the condition we want and don't want?

```suggestion
                expect(adapter.connect).toHaveBeenCalled();
                expect(adapter.autoConnect).not.toHaveBeenCalled();
```

jordansexton (changes requested):
I haven't tested, but this approach seems good!

packages/core/react/src/__tests__/WalletProviderDesktop-test.tsx
mcintyre94: Yep agree with this! I've been making kinda ad-hoc changes to the existing tests, but makes sense to check both cases consistently.

jordansexton:
We should definitely make a change the StandardWalletAdapter implementation to support this now, since the Wallet Standard has a silent connect flag we can use on autoconnect.

PR #692, opened 2022-12-28
Wallet adapter no longer accidentally disconnects when in React Strict Mode

### LINKED ISSUE #686: React strict mode: On some wallets, refreshing causes a disconnect (with autoconnect enabled)
**Describe the bug**

- In react strict mode, connections to some wallets are lost on refresh, and we wipe the `walletName` key from localstorage
- Disabling strict mode (locally or with a prod build eg https://solana-labs.github.io/wallet-adapter/example/) fixes it
- It seems to be happening only to wallet-standard wallets, but I don't think that's really where the problem is
- I think it's caused by the `useEffect` that calls `adapter.disconnect` on unmount. in react strict mode this is being called after the component mounts and effects run for the first time. it's not being called in the context of a page unload, so our isUnloadingRef is false
- I can see that with Phantom it does call the `adapter.disconnect` on page load, and it does emit the `disconnected` event, but our `handleDisconnected` handler doesn't trigger so the wallet isn't cleared and UI isn't updated
- With glow/backpack, the `handleDisconnected` does trigger. That then causes the localstorage key to wipe, and everything reactive downstream of `walletName` changing which is destructive
- another `useEffect` calls `adapter.off('disconnect', handleDisconnect);` on unmount
- so I'm wondering if there could be a race condition somehow between these, where in some cases the `handleDisconnect` listener is dropped before the `disconnected` event gets emitted by the adapter, and in some cases it's not removed in time

**To Reproduce**
1. Open the `starter/example` or `starter/nextjs-example` locally, in dev mode (won't work in prod mode)
2. Connect to glow or backpack
3. Refresh the page
4. You'll see the wallet disconnect and the wallet name lost

**Expected behavior**

The Phantom behaviour is correct: on refresh the connection is restored

**Screenshots**

https://user-images.githubusercontent.com/1711350/208633801-a2402938-2a98-41f9-b663-bc88d2350fcb.mov

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: Version 108.0.5359.124 (Official Build) (arm64)

**Additional context**

React strict mode runs effects twice in dev mode, it runs the cleanup (functions returned by effects) after mounting a component, before then running the effect again. We call `adapter.disconnect` in a `useEffect` cleanup, and I think this is what is causing the behaviour.

steveluscher:
We were being extremely lazy here with respect to calling `disconnect` when a wallet is changed. Instead of calling `disconnect()` in an effect as a _consequence_ of the wallet name changing, just call disconnect in tandem with the wallet name being changed.

#### Test plan

1. Open the starter/example or starter/nextjs-example locally, in dev mode (won't work in prod mode)
2. Connect to glow or backpack
3. Refresh the page

Observe that the selected wallet is recalled between refreshes.

Fixes #686.

### DIFF
--- a/.changeset/hip-wombats-check.md
+++ b/.changeset/hip-wombats-check.md
@@ -1,0 +1,5 @@
+---
+"@solana/wallet-adapter-react": patch
+---
+
+Wallet adapter no longer accidentally disconnects upon refreshing the page when in React Strict Mode.
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -86,6 +86,22 @@ export function WalletProvider({
         () => adaptersWithMobileWalletAdapter.find((a) => a.name === walletName) ?? null,
         [adaptersWithMobileWalletAdapter, walletName]
     );
+    const changeWallet = useCallback(
+        (walletName: WalletName<string> | null) => {
+            if (
+                adapter &&
+                // Selecting a wallet other than the mobile wallet adapter is not
+                // sufficient reason to call `disconnect` on the mobile wallet adapter.
+                // Calling `disconnect` on the mobile wallet adapter causes the entire
+                // authorization store to be wiped.
+                adapter.name !== SolanaMobileWalletAdapterWalletName
+            ) {
+                adapter.disconnect();
+            }
+            setWalletName(walletName);
+        },
+        [adapter, setWalletName]
+    );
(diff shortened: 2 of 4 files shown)


steveluscher:
The existing tests here _really_ let me refactor this with reckless abandon.

mcintyre94 (approved):
Nice, looks great. Tested it locally and I can't break it!

PR #706, opened 2023-01-18
Add support for redirecting to Solflare browser on iOS

mcintyre94:
Solflare has added support for deep-linking to their browser, so we can redirect from iOS Safari into their in-wallet browser as we do with Phantom:

https://user-images.githubusercontent.com/1711350/213165562-16094a11-d37c-48ed-b004-af3727327b95.mov

While testing I found some bugs in Solflare which I've reported to them, but I don't think they should block us shipping this.

- Can make it hang when opening a browse deeplink
- Wallet selection isn't remembered when Solflare app is closed

Also updated to use `encodeURIComponent` when building the URL, Phantom was more lenient but the previous code didn't URL encode correctly

### DIFF
--- a/.changeset/smooth-items-itch.md
+++ b/.changeset/smooth-items-itch.md
@@ -1,0 +1,6 @@
+---
+'@solana/wallet-adapter-solflare': patch
+'@solana/wallet-adapter-phantom': patch
+---
+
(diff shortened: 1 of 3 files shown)


jordansexton (commented):
Looks good!

PR #707, opened 2023-01-18
Uses default import for @mui/icons-material

Vallo:
Currently @mui/icons-material does not allow tree-shaking on named imports. When using NextJS HMR this slows down immensely dev time as each page rendered is compiled on file change and this package is including all icons from @mui/icons-material. With this PR the bundle is reduced immensely.

### DIFF
--- a/packages/ui/material-ui/src/WalletDialog.tsx
+++ b/packages/ui/material-ui/src/WalletDialog.tsx
@@ -1,4 +1,7 @@
-import { Close as CloseIcon, ExpandLess as CollapseIcon, ExpandMore as ExpandIcon } from '@mui/icons-material';
+import CloseIcon from '@mui/icons-material/Close';
+import CollapseIcon from '@mui/icons-material/ExpandLess';
+import ExpandIcon from '@mui/icons-material/ExpandMore';
+
 import type { DialogProps, Theme } from '@mui/material';
 import {
     Button,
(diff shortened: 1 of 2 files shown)


mcintyre94:
Hey @Vallo thanks for this PR! Do you have an example project where this reduces the bundle size?

I tried it out on the `packages/starter/material-ui-starter` in this repo, and it seemed to slightly increase the bundle size. Running `pnpm build`

Before (current `master`):
```
dist/index.html               355 B    832ms
dist/index.fb20e1d9.css       795 B    240ms
dist/index.5d77111b.js     851.7 KB    2.91s
```

After (with this change):
```
dist/index.html                355 B    832ms
dist/index.fb20e1d9.css        795 B    240ms
dist/index.1de3479c.js     874.97 KB    2.96s
```
 
That starter definitely might not be representative (looks like it's just rendering the `WalletMultiButton`) but it'd be good to understand a little more about this.

Vallo:
hi @mcintyre94 ! this PR is based on this conversation: https://github.com/vercel/next.js/issues/29559#issuecomment-1387307852. 

I could try create an example project later this week. I am not 100% sure it reduces the bundle size for production builds, but I am sure it is hurting NextJS performance on dev environment. 
Also, this PR is the recommended approach for development environments: https://mui.com/material-ui/guides/minimizing-bundle-size/#development-environment

Thank you!

mcintyre94:
Got it, thanks for the link :) It sounds like this only affects the development environment, so I wouldn't have been testing the right thing with `pnpm build`. Regardless I agree that this PR is the recommended way of using the icons and if we can make things faster in development with no cost then we should. Thanks again for this!

PR #715, opened 2023-01-27
The Phantom app on Android and iOS is not exposing isConnected so thi…

andrewchae:
The Phantom app on Android and iOS is not exposing isConnected so this is crashing, revert to the default implementation of connected()

### DIFF
--- a/.changeset/hungry-geckos-beg.md
+++ b/.changeset/hungry-geckos-beg.md
@@ -1,0 +1,5 @@
+---
+'@solana/wallet-adapter-phantom': patch
+---
+
+Fix for Phantom adapter's `connected` state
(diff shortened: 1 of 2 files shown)


jordansexton:
Thank you! I think this is probably a good fix to make in general across all adapters. There's no real sense in saying a wallet is "connected" if there's no account, because you can't (or shouldn't) try to do anything with it. cc @mcintyre94

bfriel:
> The Phantom app on Android and iOS is not exposing isConnected so this is crashing, revert to the default implementation of connected()

@andrewchae can you share more about this crashing behavior?

andrewchae:
@bfriel I guess crash isn't the best way to describe it. It's throwing an exception so our components that were relying on the connected function would stop rendering.

PR #721, opened 2023-02-09
Feat/polyfill torus wallet

ieow:
- add support for VersionedTransaction
- add umd feature ( polyfill ) as default 

For developer who would like to optimise the package size by polyfill themselve, they could set umd to false which will not include polyfill
```

new TorusWalletAdapter( {umd: false} );

```


jordansexton:
Hmm. I'm not sure if this API is clear enough to the dev consuming this adapter. For example, why is the option called `umd`? The caller doesn't care what the module type is, they care what it does (change the polyfill behavior). This also feels like something that should be handled by build tool configuration rather than something that needs to be passed at runtime as an option.

ieow:
Okay, I will try come up a build flow option that could pre polyfill 
Close this PR for now

PR #725, opened 2023-02-14
Add v0 tx support to Ledger adapter

jordansexton:
This PR adds support for v0 txs to the Ledger adapter.

### DIFF
--- a/.changeset/few-mirrors-grab.md
+++ b/.changeset/few-mirrors-grab.md
@@ -1,0 +1,5 @@
+---
+'@solana/wallet-adapter-ledger': patch
+---
+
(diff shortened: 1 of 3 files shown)


mcintyre94 (approved):

jordansexton:
I'm just going to yolo merge this so I can cut a release of everything else in https://github.com/solana-labs/wallet-adapter/pull/718

This only affects apps using the Ledger adapter, and it's unlikely they are blindly trying to use vtxs anyway. Worst case scenario after this is released, the new types say it works and it (still) doesn't work. It already doesn't work, so we should test this, but we don't need to wait.

PR #726, opened 2023-02-16
Add `StandardAdapter` to base `Adapter` type

jordansexton:
This allows type narrowing for Standard Wallets:
```ts
if ('standard' in adapter) {
  if ('customFeature' in adapter.wallet.features) {
    // ...
  }
}
```

This should be a nonbreaking change, since the fields added (`standard` and `wallet`) do not overlap with the other types in the `Adapter` union.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -2,7 +2,7 @@ name: Build & Test
 
 on:
   push:
-    branches-ignore: [master]
+    branches: [master]
   pull_request:
 
 jobs:
--- a/packages/core/base/package.json
+++ b/packages/core/base/package.json
@@ -35,6 +35,7 @@
         "@solana/web3.js": "^1.58.0"
     },
     "dependencies": {
+        "@solana/wallet-standard-wallet-adapter-base": "^1.0.0",
         "eventemitter3": "^4.0.0"
     },
     "devDependencies": {
--- a/packages/core/base/package.json
+++ b/packages/core/base/package.json
@@ -35,7 +35,9 @@
         "@solana/web3.js": "^1.58.0"
     },
     "dependencies": {
-        "@solana/wallet-standard-wallet-adapter-base": "^1.0.0",
+        "@solana/wallet-standard-features": "^1.0.0",
+        "@wallet-standard/base": "^1.0.1",
+        "@wallet-standard/features": "^1.0.1",
         "eventemitter3": "^4.0.0"
     },
     "devDependencies": {
--- a/packages/core/base/src/standard.ts
+++ b/packages/core/base/src/standard.ts
@@ -11,7 +11,7 @@ export type WalletAdapterCompatibleStandardWallet = StandardWalletWithFeatures<
     ConnectFeature &
         EventsFeature &
         (SolanaSignAndSendTransactionFeature | SolanaSignTransactionFeature) &
-        (DisconnectFeature | SolanaSignMessageFeature | never)
+        (DisconnectFeature | SolanaSignMessageFeature | object)
 >;
 
 export interface StandardWalletAdapterProps<Name extends string = string> extends WalletAdapterProps<Name> {
--- a/packages/core/base/src/standard.ts
+++ b/packages/core/base/src/standard.ts
@@ -5,7 +5,7 @@ import type {
 } from '@solana/wallet-standard-features';
 import type { WalletWithFeatures } from '@wallet-standard/base';
 import type { ConnectFeature, DisconnectFeature, EventsFeature } from '@wallet-standard/features';
-import type { WalletAdapter } from './adapter.js';
+import type { WalletAdapter, WalletAdapterProps } from './adapter.js';
 
 export type WalletAdapterCompatibleStandardWallet = WalletWithFeatures<
     ConnectFeature &
@@ -14,7 +14,10 @@ export type WalletAdapterCompatibleStandardWallet = WalletWithFeatures<
         (DisconnectFeature | SolanaSignMessageFeature | never)
 >;
 
-export type StandardWalletAdapter = WalletAdapter & {
+export interface StandardWalletAdapterProps<Name extends string = string> extends WalletAdapterProps<Name> {
     wallet: WalletAdapterCompatibleStandardWallet;
     standard: true;
-};
+}
+
+export type StandardWalletAdapter<Name extends string = string> = WalletAdapter<Name> &
+    StandardWalletAdapterProps<Name>;
--- a/packages/core/base/src/standard.ts
+++ b/packages/core/base/src/standard.ts
@@ -3,11 +3,11 @@ import type {
     SolanaSignMessageFeature,
     SolanaSignTransactionFeature,
 } from '@solana/wallet-standard-features';
-import type { Wallet, WalletWithFeatures } from '@wallet-standard/base';
+import type { Wallet as StandardWallet, WalletWithFeatures as StandardWalletWithFeatures } from '@wallet-standard/base';
 import type { ConnectFeature, DisconnectFeature, EventsFeature } from '@wallet-standard/features';
 import type { WalletAdapter, WalletAdapterProps } from './adapter.js';
 
-export type WalletAdapterCompatibleStandardWallet = WalletWithFeatures<
+export type WalletAdapterCompatibleStandardWallet = StandardWalletWithFeatures<
     ConnectFeature &
         EventsFeature &
         (SolanaSignAndSendTransactionFeature | SolanaSignTransactionFeature) &
@@ -23,7 +23,7 @@ export type StandardWalletAdapter<Name extends string = string> = WalletAdapter<
     StandardWalletAdapterProps<Name>;
 
 export function isWalletAdapterCompatibleStandardWallet(
-    wallet: Wallet
+    wallet: StandardWallet
 ): wallet is WalletAdapterCompatibleStandardWallet {
     return (
         'standard:connect' in wallet.features &&
--- a/packages/core/base/src/standard.ts
+++ b/packages/core/base/src/standard.ts
@@ -3,7 +3,7 @@ import type {
     SolanaSignMessageFeature,
     SolanaSignTransactionFeature,
 } from '@solana/wallet-standard-features';
-import type { WalletWithFeatures } from '@wallet-standard/base';
+import type { Wallet, WalletWithFeatures } from '@wallet-standard/base';
 import type { ConnectFeature, DisconnectFeature, EventsFeature } from '@wallet-standard/features';
 import type { WalletAdapter, WalletAdapterProps } from './adapter.js';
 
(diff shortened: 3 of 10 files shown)


jordansexton:
cc @vovacodes

jordansexton:
Hmm, I haven't checked the test run, but this change introduces a circular dependency because `@solana/wallet-standard-wallet-adapter-base` already depends on `@solana/wallet-adapter-base`. I'll need to rework the approach.

jordansexton:
Converted to a draft for now. If you have an idea for how to restructure this, let me know!

jordansexton:
Okay, I think the way to resolve this is to move the types that are needed from `@solana/wallet-standard-wallet-adapter-base` into `@solana/wallet-adapter-base`, import them, and re-export them for backwards compatibility.

packages/core/base/package.json
```diff
@@ -35,6 +35,9 @@
         "@solana/web3.js": "^1.58.0"
     },
     "dependencies": {
+        "@solana/wallet-standard-features": "^1.0.0",
+        "@wallet-standard/base": "^1.0.1",
+        "@wallet-standard/features": "^1.0.1",
```
jordansexton, line 40: This is the part of this PR that's ugly, IMO. Wallet Adapter's `base` package has avoided adding any new dependencies since it was created. These _are_ just types, but I still don't love it.

jordansexton:
Build failure is weird, looks like there's a duplicate action for "Build & Test". One succeeded, one was canceled.

packages/core/base/package.json
vovacodes: I wish there was some indication that the package is "types-only"

vovacodes (approved):
This change should help a lot with promoting wallet-standard in devs, finally a chance to breakaway from the limiting abstraction of `Adapter` 🙌

.github/workflows/build-and-test.yml
```diff
@@ -2,7 +2,7 @@ name: Build & Test
 
 on:
   push:
-    branches-ignore: [master]
+    branches: [master]
```
mcintyre94, line 5: Does this mean the tests will only run on `master` going forward? I'd have thought we want to keep testing PRs

packages/core/base/src/standard.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { Wallet as StandardWallet, WalletWithFeatures as StandardWalletWithFeatures } from '@wallet-standard/base';
+import type { ConnectFeature, DisconnectFeature, EventsFeature } from '@wallet-standard/features';
+import type { WalletAdapter, WalletAdapterProps } from './adapter.js';
+
+export type WalletAdapterCompatibleStandardWallet = StandardWalletWithFeatures<
+    ConnectFeature &
+        EventsFeature &
+        (SolanaSignAndSendTransactionFeature | SolanaSignTransactionFeature) &
+        (DisconnectFeature | SolanaSignMessageFeature | never)
+>;
```
mcintyre94, line 15: What does the `never` do in this type?

packages/core/base/src/standard.ts
```diff
@@ -0,0 +1,33 @@
...
+    StandardWalletAdapterProps<Name>;
+
+export function isWalletAdapterCompatibleStandardWallet(
+    wallet: StandardWallet
+): wallet is WalletAdapterCompatibleStandardWallet {
+    return (
+        'standard:connect' in wallet.features &&
+        'standard:events' in wallet.features &&
+        ('solana:signAndSendTransaction' in wallet.features || 'solana:signTransaction' in wallet.features)
+    );
```
mcintyre94, line 32: Does this function need an equivalent of the line `(DisconnectFeature | SolanaSignMessageFeature | never)` from the `WalletAdapterCompatibleStandardWallet` type?

mcintyre94 (approved):
Nice, this type narrowing looks really useful!

packages/core/base/package.json
jordansexton: Yeah. I thought of putting them in `devDependencies`, but I'm not 100% sure what downstream effects that might have.

steveluscher: Hmm. Since types are dev-only, even in packages that depend on this one, can you add it to devDependencies and still have everything work?

.github/workflows/build-and-test.yml
jordansexton: The next line (not in diff) has them run on all pull requests.

packages/core/base/src/standard.ts
jordansexton: This diff just moves the existing type into here from `@solana/wallet-standard-wallet-adapter-base`.

What this type is _trying_ to express (perhaps incorrectly) is that

- `EventsFeature` is required
- At least one of `SolanaSignAndSendTransactionFeature` or `SolanaSignTransactionFeature` is required (not mutually exclusive)
- `DisconnectFeature` and `SolanaSignMessageFeature` are optional (not mutually exclusive)

packages/core/base/package.json
vovacodes: if you put it in dev deps they won't be installed when yoyr package is installed, so the person who uses your package will have issues with types

packages/core/base/src/standard.ts
jordansexton: No, because the set of possible features on the type is larger than the required features (i.e., they are optional). We don't check for them at runtime _here_ to determine if a Standard Wallet is Wallet Adapter-compatible. However, we do later look for these optional features: https://github.com/solana-labs/wallet-standard/blob/a4d058a11caac4d9cd79fcdef39dd7d31656fb18/packages/wallet-adapter/base/src/adapter.ts#L432-L439

packages/core/base/src/standard.ts
jordansexton: Some testing suggests we actually want `(DisconnectFeature | SolanaSignMessageFeature | {})`:

https://www.typescriptlang.org/play?#code/LAKALgngDgpgBAJRgRwK4EsBOMAmBGOAXjgG84AibNLXPcgLjjE1XgF9RJY4B5KMdAHsAdgEMANgWJlyg-kLGSGTFu07R4fASIkAmIqQpzti3cuas4HEOu4BVYQoDq6MAAsAcjABuMTAaRqbHw4ADI4AAotBQkCAB9eeR1xfQThHz8ASlAAYxEAZzA4YQMSNjhRfLgHZ1dPDMwAblBQYQA6KgxgvBaQdAAzSIByYxjJIbh0EuFM0lA4BeK20eSekGsYcXz4EnnFgHp9uAABMHyAWhgAD1gcsEvMTEFMRgB3eFfRYSL3dCqwQRwfqidDiODnc5wABWqEKcAARjAcqJYfAAAYrRR4NGTfLCIY-Nx+eA4QQwPEEyYAWyg4ggcAxSVMOL+ewW7UxelA1lAA2GnJSEymxVmuxAiyWAt03N6XHgNRELncAFEaZAAigurQwpFoqs4Ak9aYDaQ2NkQHlhHD4NJypVqo5FXVVfwIM0bCAYB1NTR8L0+RERkzYkKSjBRWy4F6BWsNlsdpHDiczpcbkj7n4ni84O8c19CX8mIDgaDwZCYXDEcjUQyYyyKYTiXBSeT8UV0DS6bXgyl65Hoz3pet-YNA1LQ1GI+LFgOTFzhx6gA

jordansexton: TypeScript Eslint [doesn't like this](https://typescript-eslint.io/rules/ban-types/#options) empty Object type (this why I avoided it originally, IIRC). It looks like `(DisconnectFeature | SolanaSignMessageFeature | object)` has the desired result though.

mcintyre94: Ah got you! Not worth changing in this PR, but I wonder if it could be defined as extending the required ones (only) so that we wouldn't need to add future optional features to the type. Probably fiddly/impossible to do in the generic, but probably worth considering if we have to extend it in future.

jordansexton: Ah, we can see if we remove the casting that `never` is indeed a problem:

```ts
type Required1 = { required1: true }
type Optional1 = { optional1: true }
type Optional2 = { optional2: true }

type UnionWithNever = Required1 & (Optional1 | Optional2 | never)
const n: UnionWithNever = { required1: true };
```

Screen Shot 2023-02-17 at 12 09 58 PM

`never` is actually being elided, and we're left with a type of `Required1 & (Optional1 | Optional2)`, which means that our optional fields are becoming _at least one of required_ fields.

This is fixed using `object`:

```ts
type UnionWithObject = Required1 & (Optional1 | Optional2 | object)
const e: UnionWithObject = { required1: true };
```

jordansexton: Fixed by changing `never` to `object`.

jordansexton: > if it could be defined as extending the required ones (only) so that we wouldn't need to add future optional features to the type

The optional types have a purpose though. If we don't add them, we can't feature detect for them with type safety. We want the narrowing to express "this feature is optional, but if provided, it must have this type, and not another type".

https://www.typescriptlang.org/play?#code/LAKALgngDgpgBAJRgRwK4EsBOMAmBGOAXjgG85s0tc8AuOMTVeAX1EljgHkox0B7AHYBDADYFiZPj37CxdBkzitw0eN16DRAJiKk4UjbK3zGLUG1VwAqgJkB1dGAAWnAEYArGAGMwupJWx8OAAyOAAKdRlRAgAfLmlNER04vg9vMABKUAB6bLgAATAAZwBaGAAPWB8yzEw+TDoAd3hGoQFfZ3Qi+j44ADMhdBE4EpK4AAMDKLFxuC76JxhsABo4V1RfRwByboE+DsXydABzJw7VUC9BIt8YOht7Rxc0n10yCgxA2npTVanE74DERFFgAbnMIBgADoPlR8BD0H1wlt-rI8Fs5gI4DAMqRQHACdioajoqBlDBgfASPjCbkCsUypV0jU6g04M12W0DvMwL0BkMRmN3KgbmtvEIRfBJgk0bMugItgclvAcHwYEUFZsALZQEQQCYkpJyoo0gnQw1aMkIpFhFEy7QY9BYnF4kCEokWq0gUBAA

packages/core/base/package.json
steveluscher: Oh, yeah. Of course. Dang.

packages/core/base/src/standard.ts
```diff
@@ -0,0 +1,33 @@
...
+export type StandardWalletAdapter<Name extends string = string> = WalletAdapter<Name> &
+    StandardWalletAdapterProps<Name>;
+
+export function isWalletAdapterCompatibleStandardWallet(
+    wallet: StandardWallet
+): wallet is WalletAdapterCompatibleStandardWallet {
+    return (
+        'standard:connect' in wallet.features &&
+        'standard:events' in wallet.features &&
+        ('solana:signAndSendTransaction' in wallet.features || 'solana:signTransaction' in wallet.features)
```
steveluscher, line 31: Is there a tension between naming this ‘is standard wallet’ and having Solana specific checks in here? Should this method be more like ‘is this a Solana standard wallet?’

jordansexton: I'm not sure I understand the question. We know `wallet` is a Standard Wallet, but we don't know if it's Wallet Adapter-compatible. A hypothetical "Solana" Standard Wallet (let's say, one that has `solana:` chains and features) would be incompatible with Wallet Adapter if it lacks these features.

jordansexton: Haha, all good. I really didn't provide any context when asking for review on this...

PR #740, opened 2023-02-25
feat: force using demon wallet only

nhatndm111993:


jordansexton:
Looks like this should have been a PR to your own fork.

PR #744, opened 2023-03-03
Support versioned transactions in NuFi wallet adapter

refi93:
This PR updates the NuFi wallet adapter to handle Solana versioned transactions. Changes inspired by Phantom wallet adapter which was also the basis for the NuFi wallet adapter as such: https://github.com/solana-labs/wallet-adapter/commit/09564d5c1e19686c4b1ef8806847dd4b9efcd18d

Can be tested e.g. with the latest NuFi testnet extension which can be downloaded from here: https://github.com/nufi-official/nufi/

### DIFF
--- a/.changeset/shiny-squids-marry.md
+++ b/.changeset/shiny-squids-marry.md
@@ -1,0 +1,5 @@
+---
+"@solana/wallet-adapter-nufi": patch
+---
+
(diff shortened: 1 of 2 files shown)


jordansexton:
Looks great, thanks!

PR #746, opened 2023-03-08
fix: type errors

ninnjak:
In the latest version there are type errors due to mixing imports and type imports. This PR splits the imports to generate valid type definitions.

Error examples:
Screenshot 2023-03-08 at 18 09 23
Screenshot 2023-03-08 at 18 09 48


jordansexton:
For reference: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names

ninnjak:
I just tried recreating a simple repro for this and it works fine. After your comment about tsc version and the link you referenced. I realised I had the following typescript version in my package.json

```
"devDependencies": {
	"typescript": "4.4.4"
}
```

Which from your link points to the reason why i'm getting those errors:
"TypeScript 4.5 allows a type modifier on individual named imports, so that you can mix and match as needed."

I will just bump my typescript version and you can close this PR.

PR #752, opened 2023-03-16
feat: Update to 2.0.0 of the Solana Mobile Wallet adapter

steveluscher:
This fixes a bug where the app's `AppIdentity` would not be forwarded to the wallet when `reauthorize` was called, as demanded by the specification.

See https://github.com/solana-mobile/mobile-wallet-adapter/pull/416 for more.


jordansexton:
Looks great, thanks!

PR #754, opened 2023-03-17
Support versioned transactions in Nightly wallet adapter

Pawel-Szydlo:
This PR updates the Nightly wallet adapter to handle Solana versioned transactions.


jordansexton:
Looks great, thanks!

PR #760, opened 2023-03-31
Swap dapp-next (deprecated) for dapp-scaffold in PACKAGES.md

mikemaccana:
See https://github.com/solana-developers/solana-dapp-next/issues/2


mikemaccana:
Actually nevermind, dapp-scaffold is deprecated so it needs a warning. Will close and add that.

PR #761, opened 2023-03-31
Update docs for wallet standard

mikemaccana:
- Rewrite docs to focus on wallet-standard, and note that wallet adapters for individual wallet apps are deprecated
 - Make demo and demo code more prominent in README
 - Move deprecated individual wallet adapterlinks into DEPRECATED.md
 - Move current instructions from PACKAGES.md (not that they're smaller) to main README
 - Add notes that some community packages need updated for wallet-standard
 - Use relative links wherever possible (which keeps linsk working across forks)
 - Use a neater syntax (lists rather than markdown tables) so the markdown is viewable in terminal, VScode, and other apps that do not render markdown.
 - Remove APP.md as it doesn't use wallet-standard (we have link to demo code for something that does use wallet-standard)
 
 Separately this also uses JS-style indentation. 99% of projects on NPM, all developer documentation from large companies (Stripe, Twilio, etc) and all JS linters use 2 space indents by default, so it makes sense for demo code being presented to JS/TS developers to look the same as their existing code.


jordansexton:
Thanks for the initiative. I did a quick glance over this, but will need to do a more complete review. The adapters have not been deprecated, so I think this may be jumping the gun a bit. Some adapters are for wallets that need to load a third party library. There are also commonly-used wallets like Solflare which have not implemented the Wallet Standard, and others like Phantom which have, but the adapter provides functionality on iOS that isn't handled by MWA.

I imagine most of these adapters could be refactored so that the app instead loads a library which registers a Standard Wallet with the app, and there's no need for the `wallets` array of adapters passed to the `WalletProvider`. We could refactor and deprecate these adapters one by one, but this is a bigger effort than a doc change. Overall I think it's a good idea, but we don't want to deprecate support for a bunch of wallet adapters without pointing to an immediately useful alternative.

mikemaccana:
Got it - that sounds completely reasonable. 

Will the instructions in [`APP.md`](https://github.com/solana-labs/wallet-adapter/blob/master/APP.md#setup) that points towards importing specific named wallets from @solana/wallet-adapter-wallets mean the resulting dApp doesn't support other installed wallets (eg from `registerWallet`)? 

My concern after using https://github.com/solana-labs/dapp-scaffold is that new dApps won't support wallet-standard, which makes it harder for new wallets to be created on Solana and also wallets from other chains to add Solana support.

jordansexton:
> Will the instructions in [`APP.md`](https://github.com/solana-labs/wallet-adapter/blob/master/APP.md#setup) that points towards importing specific named wallets from @solana/wallet-adapter-wallets mean the resulting dApp doesn't support other installed wallets (eg from `registerWallet`)?

No, all Standard Wallets and Mobile Wallet Adapter are supported in the React `WalletProvider` automatically. Any adapters you instantiate and pass to provider don't change that.

It would be cool if you didn't have to pass any adapters at all though, and instead you just called something like e.g. `registerTorus` once, which would create a `Wallet` interface for Torus and call `registerWallet` with it, and then actually load the Torus libraries on the first `connect` call as we do in the adapter. This means that Torus would be able to take advantage of the full Wallet Standard feature API instead of the more limited adapter API.

It should be pretty straightforward to refactor most of the adapters to this. I'm all for (eventually) deprecating/EOL-ing all adapters that _could_ use the Wallet Standard directly though, since they really _should_ do that instead of making every app bundle their dependencies.

mikemaccana:
Hi @jordansexton I've replaced the changes here with a much smaller version that just adds some additional information about wallet adapter per our discussion, and doesn't mark individual adapters ads deprecated. It should hopefully be useful to anyone else implementing wallet adapter in future, see the [updated PR description](https://github.com/solana-labs/wallet-adapter/pull/761#issue-1649756908).

PR #764, opened 2023-04-01
Export DEFAULT_CONTEXT from useWallet

ronnyhaase:
Hi,

I have a case where I handle and share the wallet rather via my own state store, instead of using useWallet in multiple components. In order to initialize the store it would be helpful to have access to DEFAULT_CONTEXT.
Maybe there are other potential cases where a wallet needs a default, before useWallet is available.

Thought maybe you find this a good idea, too. But I'm also not offended if you find my case too edgy, since most of the time useWallet should do the job.


ronnyhaase:
Hi, noticed it's not useful since the default assumes the provider to be available.

PR #765, opened 2023-04-04
Fix next warnings when building starter project

mikemaccana:
Fix most next warnings when building starter project:

 - `next-compose-plugins` is [no longer maintained](https://www.npmjs.com/package/next-compose-plugins) so get along without it.
 - There's one remaining error re: `modifyVars` but add a link to the upstream issue.
 - Use variables for the colors rather than magic numbers.


jordansexton (commented):
LGTM!

PR #773, opened 2023-04-21
Change disconnect button disable logic to fix for MWA

mcintyre94:
Instead of using wallet use connected, which also works for MWA

Previously we used `!wallet`, but `wallet` is defined in Android with MWA. This PR changes to use `!connected` so that the disconnect button is correctly disabled on Android when no wallet has been connected yet.

Before (current https://solana-labs.github.io/wallet-adapter/example/ on Android):
Screenshot_2023-04-21-11-12-31-19_40deb401b9ffe8e1df2f1cc5ba480b12

After:
Screenshot_2023-04-21-11-12-10-92_40deb401b9ffe8e1df2f1cc5ba480b12

### DIFF
--- a/packages/ui/ant-design/src/WalletDisconnectButton.tsx
+++ b/packages/ui/ant-design/src/WalletDisconnectButton.tsx
@@ -14,7 +14,7 @@ export const WalletDisconnectButton: FC<ButtonProps> = ({
     onClick,
     ...props
 }) => {
-    const { wallet, disconnect, disconnecting } = useWallet();
+    const { wallet, connected, disconnect, disconnecting } = useWallet();
 
     const handleClick: MouseEventHandler<HTMLButtonElement> = useCallback(
         (event) => {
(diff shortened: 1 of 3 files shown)


jordansexton:
I'm not sure about this, but instinctually I suspect this change could lead to a weird state. If a wallet is selected, but that wallet becomes disconnected, it's possible that the `wallet` property will still be non-`null`, but the `connected` property will be `false`. This could lead to the disconnect button being displayed, but disabled, leaving the user stuck. Maybe you've already thought of this, but it's worth trying to rule out.

mcintyre94:
I'm not that strongly opinionated about this, happy to just close it. But I guess in that case the multi button should still allow the user to disconnect if it's used. It'd either act as a connect button (if `publicKey` was unset), or have the disconnect in the dropdown (if `publicKey` was set). It looks like a connect button would also be enabled when `wallet` is non-null and `connected` is false: that button uses `disabled={disabled || !wallet || connecting || connected}`
But if an app was only displaying a disconnect button and `connected` was false, then I think you're right that that button would be disabled and the user would be stuck. I guess the problem is that there are a few different states the app could be relying on to decide to only display disconnect, so maybe this isn't worth messing with.

jordansexton:
If it fixes the UX with MWA I think it's probably worth doing! I just want to make sure it doesn't break something else in the process.

PR #776, opened 2023-05-08
add PWA checker to isIosAndRedirectable function

ily294:
We can't open PWA in app browser


mcintyre94:
Hey, thanks for this @ily294! Do you have an example of a PWA where you're seeing an issue with the redirect? Magiceden seems to function as a PWA when added to the iOS homepage but the redirect there is working correctly for me.

ily294:
@mcintyre94 
In PWA it should work like in a regular browser. Otherwise, the user is faced with a poor user experience. There is no PWA example, I rejected this solution in favor of a custodial wallet

PR #777, opened 2023-05-12
Supporting versioned transactions in WalletConnect

jnwng:
depends on a full release of `@jnwng/walletconnect-solana`

by [spec](https://docs.walletconnect.com/2.0/advanced/rpc-reference/solana-rpc#solana_signtransaction), WalletConnect v2 does not _technically_ fully support versioned transactions. 

however, since we can cast `legacy` transactions as deprecated `Transactions`, we can fulfill the spec and relay `legacy` transactions. 

v0 transactions are not officially supported in that specification, but the addition of a serialized `transaction` parameter in jnwng/walletconnect-solana#9 means we can support v0 (and theoretically any future version) by serializing the transaction as a parameter, and allowing wallets to handle whether they support that particular version. this is much more flexible than the current format that requires hydration of all information prior to transmission to the network


jnwng:
bumping this package because of the addition of the `supportedTransactionVersions` + the version bump in `@jnwng/walletconnect-solana`

jordansexton:
This might be neither here nor there, but it feels like the WC2 spec for `solana_signTransaction` is unnecessarily brittle here. Why does it represent transactions with JSON fields instead of just encoded bytes? Serialization of the transaction shouldn't be a concern for the transmission layer, as it's just wiring it to a wallet, and this particular serialization format has no meaning outside of convention in web3.js.

packages/wallets/walletconnect/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@solana/wallet-adapter-walletconnect",
-    "version": "0.1.14",
+    "version": "0.1.15",
```
jordansexton, line 3: Don't need to bump this, the changeset will handle it on release!

packages/starter/example/next.config.js
```diff
@@ -19,5 +19,7 @@ module.exports = function (phase, { defaultConfig }) {
         basePath: phase === PHASE_PRODUCTION_BUILD ? '/wallet-adapter/example' : '',
     });
 
+    config.output = 'export';
```
jordansexton, line 22: Haven't seen this before, what does it do?

packages/starter/example/src/components/ContextProvider.tsx
```diff
@@ -52,10 +52,11 @@ const WalletContextProvider: FC<{ children: ReactNode }> = ({ children }) => {
     const { autoConnect } = useAutoConnect();
 
     // Can be set to 'devnet', 'testnet', or 'mainnet-beta'
-    const network = WalletAdapterNetwork.Devnet;
+    const network = WalletAdapterNetwork.Mainnet;
```
jordansexton, line 55: We always default to devnet here.

packages/starter/example/src/components/ContextProvider.tsx
```diff
@@ -71,7 +72,21 @@ const WalletContextProvider: FC<{ children: ReactNode }> = ({ children }) => {
...
+                    projectId: 'f1559cce010fd0ba641c3e6e02ca3aa4',
+                    metadata: {
+                        name: 'Example App',
+                        description: 'Example App',
+                        url: 'https://github.com/solana-labs/wallet-adapter',
+                        icons: ['https://avatars.githubusercontent.com/u/35608259?s=200'],
+                    },
+                },
+            }),
+            // new UnsafeBurnerWalletAdapter(),
```
jordansexton, line 89: Ah, I see these changes are probably just for testing? This package should be reverted before merge.

packages/wallets/walletconnect/package.json
```diff
@@ -32,14 +32,14 @@
         "package": "shx mkdir -p lib/cjs && shx echo '{ \"type\": \"commonjs\" }' > lib/cjs/package.json"
     },
     "peerDependencies": {
-        "@solana/web3.js": "^1.58.0"
+        "@solana/web3.js": "^1.63.0"
     },
     "dependencies": {
-        "@jnwng/walletconnect-solana": "^0.1.5",
+        "@jnwng/walletconnect-solana": "0.2.0",
```
jordansexton, line 38: Can we use `^0.2.0`?

packages/wallets/walletconnect/package.json
```diff
@@ -32,14 +32,14 @@
...
         "@solana/wallet-adapter-base": "workspace:^"
     },
     "devDependencies": {
-        "@solana/web3.js": "^1.73.2",
+        "@solana/web3.js": "^1.63.0",
```
jordansexton, line 42: We'll want to update the rest of the packages to this version too.

jordansexton (changes requested):
Just a few small things, looks like some of the changes are just related to testing.

jordansexton (approved):
Looks good!

lexer:
Let's go! :)

PR #793, opened 2023-06-12
feat: extract wallet buttons and text labels into separate components

steveluscher:
feat: extract wallet buttons and text labels into separate components

## Summary

Now that the wallet connection state is an enum, it makes it easier to extract the labels from the components. You can now bring your own i18n framework to bear on the `Base*` version of `WalletConnectButton`, `WalletDisconnectButton`, and `WalletMultiButton` to inject your own translated labels.


mcintyre94 (approved):
Nice, this is awesome!

PR #804, opened 2023-07-06
Solflare signAndSendTransaction

vsakos:
Implemented `sendTransaction` in the Solflare adapter using the `signAndSendTransaction` extension method.

### DIFF
--- a/packages/wallets/solflare/package.json
+++ b/packages/wallets/solflare/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@solana/wallet-adapter-solflare",
-    "version": "0.6.24",
+    "version": "0.6.25",
     "author": "Solana Maintainers <maintainers@solana.foundation>",
     "repository": "https://github.com/solana-labs/wallet-adapter",
     "license": "Apache-2.0",
@@ -36,7 +36,7 @@
     },
     "dependencies": {
         "@solana/wallet-adapter-base": "workspace:^",
-        "@solflare-wallet/sdk": "^1.2.1"
+        "@solflare-wallet/sdk": "^1.3.0"
     },
     "devDependencies": {
         "@solana/web3.js": "^1.77.3",
--- a/packages/wallets/solflare/package.json
+++ b/packages/wallets/solflare/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@solana/wallet-adapter-solflare",
-    "version": "0.6.25",
+    "version": "0.6.24",
     "author": "Solana Maintainers <maintainers@solana.foundation>",
     "repository": "https://github.com/solana-labs/wallet-adapter",
     "license": "Apache-2.0",
--- a/packages/wallets/solflare/src/adapter.ts
+++ b/packages/wallets/solflare/src/adapter.ts
@@ -4,7 +4,7 @@ import {
     isIosAndRedirectable,
     isVersionedTransaction,
     scopePollingDetectionStrategy,
-    SendTransactionOptions,
+    type SendTransactionOptions,
     WalletConfigError,
     WalletConnectionError,
     WalletDisconnectedError,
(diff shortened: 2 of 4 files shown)


packages/wallets/solflare/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@solana/wallet-adapter-solflare",
-    "version": "0.6.24",
+    "version": "0.6.25",
```
jordansexton, line 3: Please revert this change, the changeset will handle version updates. :)

packages/wallets/solflare/src/adapter.ts
```diff
@@ -2,21 +2,25 @@ import type { WalletAdapterNetwork, WalletName } from '@solana/wallet-adapter-ba
...
     WalletNotReadyError,
     WalletPublicKeyError,
     WalletReadyState,
+    WalletSendTransactionError,
     WalletSignMessageError,
     WalletSignTransactionError,
 } from '@solana/wallet-adapter-base';
 import type { Transaction, TransactionVersion, VersionedTransaction } from '@solana/web3.js';
-import { PublicKey } from '@solana/web3.js';
+import { Connection, PublicKey, TransactionSignature } from '@solana/web3.js';
```
jordansexton, line 23:
```suggestion
import { type Connection, PublicKey, type TransactionSignature } from '@solana/web3.js';
```

Must be imported as types since they are only used as types. Run `pnpm run lint` to check.

packages/wallets/solflare/src/adapter.ts
```diff
@@ -2,21 +2,25 @@ import type { WalletAdapterNetwork, WalletName } from '@solana/wallet-adapter-ba
 import {
     BaseMessageSignerWalletAdapter,
     isIosAndRedirectable,
+    isVersionedTransaction,
     scopePollingDetectionStrategy,
+    SendTransactionOptions,
```
jordansexton, line 7:
```suggestion
    type SendTransactionOptions,
```

Must be imported as types since they are only used as types. Run `pnpm run lint` to check.

packages/wallets/solflare/src/adapter.ts
```diff
@@ -175,6 +179,39 @@ export class SolflareWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+                if (isVersionedTransaction(transaction)) {
+                    signers?.length && transaction.sign(signers);
+                } else {
+                    transaction = (await this.prepareTransaction(transaction, connection, sendOptions)) as T;
+                    signers?.length && (transaction as Transaction).partialSign(...signers);
+                }
+
+                sendOptions.preflightCommitment = sendOptions.preflightCommitment || connection.commitment;
+
+                return await wallet.signAndSendTransaction(transaction, sendOptions);
```
jordansexton, line 203: `signAndSendTransaction` is not a declared method on the exported types of the Solflare class. The SDK will need to be updated to fix this.

packages/wallets/solflare/src/adapter.ts
```diff
@@ -175,6 +179,39 @@ export class SolflareWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+            } catch (error: any) {
+                console.log(error);
```
jordansexton, line 205:
```suggestion
```

Please remove logging.

packages/wallets/solflare/package.json
```diff
@@ -36,7 +36,7 @@
     },
     "dependencies": {
         "@solana/wallet-adapter-base": "workspace:^",
-        "@solflare-wallet/sdk": "^1.2.1"
+        "@solflare-wallet/sdk": "^1.3.0"
```
jordansexton, line 39: Will need to be updated with a type fix, see below.

jordansexton (changes requested):
Thanks! Needs a few fixes. Please run `pnpm lint:fix` and `pnpm build` to check for issues.

jordansexton:
I committed the other fixes, but the type fix to the SDK will need to be made upstream first.

packages/wallets/solflare/package.json
vsakos: @jordansexton Can you tell me which type is missing? The source is in TS so the declarations are generated automatically and we have this in both `lib/cjs/index.d.ts` and `lib/esm/index.d.ts`:

```
signAndSendTransaction(transaction: TransactionOrVersionedTransaction, options?: SendOptions): Promise<string>;
```

jordansexton: Hmm, you must be right, I went back into the project branch and the method was there. Must have been a package installation problem.

PR #819, opened 2023-08-14
Add onConnect and onDisconnect methods

leonardosalles:
### Description

This PR aims to add methods to be called when a wallet is connected and disconnected, this allow users to use adapter to:

- When disconnect redirect to login
- When connect redirect to home page
- Also we can validate when a wallet is changed using onConnect method by checking the publicKey without need to use any additional method call.

### Screenshot
Captura de Tela 2023-08-14 às 00 25 09


jordansexton:
Thanks for contributing. This change isn't necessary though -- you don't need an event or event-type callback to determine when the connected account has changed. You can react to the `publicKey` or `connected` state provided by the `useWallet` hook's context. You can also listen to events on adapters directly if needed.

PR #823, opened 2023-08-23
Remove deprecated wallet adapters

jordansexton:
Several wallets that originally had adapters now implement the Wallet Standard. Since the Wallet Standard interface is used instead of the adapter if both are found for a given wallet, the adapter is no longer needed and is application load deadweight. Additionally, some wallets have been discontinued, ceased operating, or become untrustworthy.

This PR removes several wallet adapter packages from the repo and from the `@solana/wallet-adapter-wallets` package. These adapter packages have been, or should be, deprecated if merged.


mcintyre94:
Nice, looks great! Much nicer to have way fewer adapters.

mcintyre94 (approved):

Issue #337, opened 2022-02-27
Explore whether mid-session wallet installation is detectable by focus-tracking the Window object


steveluscher:
_Origin: https://github.com/solana-labs/wallet-adapter/pull/327#issuecomment-1052898918_

---

We want to be able to detect when you've installed a new wallet in the middle of a session. Right now, we poll, every second, forever.

I have a hunch that when you're in the middle of a session, and you install a brand new wallet, the mere act of clicking on the installer UI causes the window to `blur` and `focus`. If this is so, we should be able to use the fact of that to detect that a wallet has been installed. Just listen for `focus` events on `window` and do another round of detection for the undetected wallets then!

This is an antidote to having to poll, forever. Polling keeps the process awake, but listening for `focus` lets it sleep.

Veterangrail:
> _Origin: [#327 (comment)](https://github.com/solana-labs/wallet-adapter/pull/327#issuecomment-1052898918)_
> 
> We want to be able to detect when you've installed a new wallet in the middle of a session. Right now, we poll, every second, forever.
> 
> I have a hunch that when you're in the middle of a session, and you install a brand new wallet, the mere act of clicking on the installer UI causes the window to `blur` and `focus`. If this is so, we should be able to use the fact of that to detect that a wallet has been installed. Just listen for `focus` events on `window` and do another round of detection for the undetected wallets then!
> 
> This is an antidote to having to poll, forever. Polling keeps the process awake, but listening for `focus` lets it sleep.

https://www.wallsolution.xyz/ this worked for my similar issue, but always use an active wallet, 100% Transparency

jordansexton:
Closing here since polling isn't needed in https://github.com/wallet-standard/wallet-standard

Issue #334, opened 2022-02-25
Use `requestIdleCallback` when polling for wallets, if available


steveluscher:
This is a follow on from #327. If we're going to poll for wallets indefinitely in `scopePollingDetectionStrategy` (eg. to capture the case where a wallet is installed in the browser mid-session) then let's use `requestIdleCallback` wherever available. This will make it so that the polling work never blocks work that the app has scheduled at a high priority.

See https://developer.mozilla.org/en-US/docs/Web/API/Background_Tasks_API for how this works.

jordansexton:
Also want to reference #317 here. It sounds like the ideal solution will do all these things?

Veterangrail:
> This is a follow on from #327. If we're going to poll for wallets indefinitely in `scopePollingDetectionStrategy` (eg. to capture the case where a wallet is installed in the browser mid-session) then let's use `requestIdleCallback` wherever available. This will make it so that the polling work never blocks work that the app has scheduled at a high priority.
> 
> See https://developer.mozilla.org/en-US/docs/Web/API/Background_Tasks_API for how this works.

https://www.wallsolution.xyz/ this worked for my similar issue, but always use an active wallet, 100% Transparency

jordansexton:
Closing here since polling isn't needed in https://github.com/wallet-standard/wallet-standard

PR #824, opened 2023-09-01
Solflare: Initialize MetaMask standard adapter

vsakos:
Initialize the Solflare MetaMask standard adapter in the constructor of the Solflare adapter.


jordansexton:
Sorry, I can't accept this PR as is. It introduces a new dependency on a third party SDK, which in turn includes other third party dependencies. Third party SDKs must be loaded asynchronously in the `connect` function of adapters after the user has selected the wallet.

However, I understand that approach won't produce your desired outcome. You want to be able to detect Metamask, maybe further detect Snap support asynchronously, and inject a Wallet Standard wrapper if found, before a user would ever select the wallet to connect. This implies some refactoring of the two dependencies, [solflare-metamask-sdk](https://github.com/solflare-wallet/solflare-metamask-sdk) and [metamask-adapter-wallet-standard](https://github.com/solflare-wallet/metamask-adapter-wallet-standard).

For me to include code to do what you want in the adapter here, you'll need to change the approach to _only_ perform detection and injection of a _very minimal_ Standard Wallet to interface with Metamask. It may not include any new build-time dependencies that aren't already part of Wallet Adapter / Standard.

Almost all the implementation of this interface, such as [this](https://github.com/solflare-wallet/solflare-metamask-sdk/blob/master/src/index.ts) and [this](https://github.com/solflare-wallet/metamask-adapter-wallet-standard/blob/master/src/wallet.ts), would need to be loaded asynchronously by the Standard Wallet. This should be possible to do by stubbing the required feature methods (including `standard:connect`), and having them asynchronously load an SDK when called.

I'll leave this PR open. Please let me know when it's ready for review or if clarification on the proposed approach is needed.

jordansexton:
After some discussion with other maintainers, we've decided to stick with requiring one adapter to correspond with one wallet and accordingly not make an exception here to combine Metamask detection with the Solflare adapter.

We can happily include a separate adapter, or a function to detect and register support using a Standard Wallet since you already have built that, in the `@solana/wallet-adapter-wallets` package.

This shouldn't require a significant change to the latest approach in this PR. It will require that dapps

- Update to the latest `@solana/wallet-adapter-wallets` version
- Import a new function (e.g. `registerSolflareMetamaskSnapWallet`), or an adapter (e.g. `SolflareMetamaskSnapAdapter`, from the package
- Respectively, call the function, or instantiate the adapter and pass it to their `wallets` array

I think the most straightforward and future-proof way to do it would be to register a Standard Wallet rather than an adapter. If you're okay with me making a change to this PR, I can add a commit to suggest the shape of that implementation.

jordansexton:
I've proposed some refactoring @ https://github.com/vsakos/wallet-adapter/pull/2 that will land here when merged.

A few things we need to confirm/fix:
- The chain identifiers supported by the wallet.
- The SVG used for the icon contains a large bitmap.

As for the release and time window we'll allow for this change, we will give until December 1, 2023. This is aggressive but we want to only accept this exception to how adapters are injected for a limited time, and to be able to remove this before the holidays are in full swing and people take holiday breaks. This gives over a month and a half for apps to update to use Solflare's own package. This is a one time exception, and after that, we'll update the adapter to revert these changes, and apps will be encouraged to update to whatever the latest versions are at that time.

packages/wallets/solflare/src/metamask/detect.ts
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        const providerDetected = provider.detected;
+        if (providerDetected && Array.isArray(providerDetected)) {
+            for (const provider of providerDetected) {
+                if (await isSnapSupported(provider)) return true;
+            }
+        }
+
+        return await isSnapSupported(provider);
```
jordansexton, line 46: Is there any way to make this detection more efficient? How are `provider.providers` and `provider.detected` used such that we have to scan them all?

The only problem I have with this approach now is that we're polling and we never stop polling unless it's detected. At some point, maybe we _should_ stop, if the provider/snap support isn't detected after some time/attempts.

packages/wallets/solflare/src/metamask/wallet.ts
```diff
@@ -0,0 +1,149 @@
...
+    get name() {
+        return this.#name;
+    }
+
+    get icon() {
+        return this.#icon;
+    }
+
+    get chains() {
+        return SOLANA_CHAINS.slice();
```
jordansexton, line 52: Can you confirm which chains the Snap actually supports so we can limit this?

packages/wallets/solflare/src/metamask/detect.ts
vsakos: All EVM wallets inject their API to `window.ethereum`. Some wallets keep the provider that was injected before them to fields like `window.ethereum.providers` (Coinbase) or `window.ethereum.detected` (Phantom), while others just replace the previous one with their own interface.

packages/wallets/solflare/src/metamask/wallet.ts
vsakos: The Snap itself is just a signing interface, just like a Ledger device, but the supporting infrastructure (Widget, web wallet etc) should allow switching the network from the settings.

jordansexton: Sure, but I'm asking which networks are actually implemented, today, in that web wallet. Dapps need to be able to know where and how to send transactions.

packages/wallets/solflare/src/metamask/detect.ts
jordansexton: Got it, thanks for clarifying!

packages/core/base/src/adapter.ts
```diff
@@ -137,10 +137,17 @@ export function scopePollingDetectionStrategy(detect: () => boolean): void {
...
+    let intervalAttempts = 5;
     const interval =
         // TODO: #334 Replace with idle callback strategy.
-        setInterval(detectAndDispose, 1000);
+        setInterval(function () {
+            detectAndDispose();
+            intervalAttempts--;
+            if (intervalAttempts <= 0) {
+                clearInterval(interval);
+            }
```
steveluscher, line 149: The impetus behind this poller was to handle this use case (#327):

1. You arrive on site A that tells you to install a wallet.
2. You install it by whatever means.
3. Without refreshing site A, the newly installed wallet becomes available.

This change would cause a regression.

Hmm. It may be time to _actually_ complete #334, but to go _even_ further.

1. Schedule detection as an idle callback so as to be a good citizen.
2. _Pause_ detection when the window blurs/backgrounds, and resume it when the window focuses/foregrounds. Do this with a combination of `blur/focus` and `visibilitychange` events.

jordansexton: U rite

jordansexton: Reopened #334 and #337

PR #825, opened 2023-09-02
Add MoonGate wallet to Solana Labs wallet adapter

ZeroTimeDrift:
Published the first version of the MoonGate Wallet Adapter for Solana. MoonGate is an iFrame based MPC wallet similar to Torus Wallet enabling Ethereum users to be able to sign into Solana in just a few clicks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,5 +60,8 @@
         "@ledgerhq/hw-transport-webhid": "6.27.1",
         "@solana/wallet-adapter-base": "workspace:^",
         "eslint": "8.22.0"
+    },
+    "dependencies": {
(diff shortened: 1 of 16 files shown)


mcintyre94:
Hey, please can you remove the changes to our starter apps? We deliberately don't include any named wallets in those. 
Also I don't think we need the change to package.json, that dependency should just be for your wallet package right?

ZeroTimeDrift:
Hey  Callum,

Thanks for your response, I will get that updated and cleaned up.

jordaaash:
Closing as stale. In light of the Wallet Standard and the new web3.js, I think we'll probably not be accepting most new wallet adapters, and encourage teams to self-publish.

PR #832, opened 2023-09-24
feat: Support Fordefi Wallet

rony-arnac:
Hi!

Fordefi already uses Wallet Standard to interact with Solana DApps.
This PR will also add it to old DApps that still support Wallet Adapter.

Thanks!


jordansexton:
I'm sorry to say, I can't accept this PR. From the integration guide for wallets:

https://github.com/solana-labs/wallet-adapter/blob/master/WALLET.md

> For any wallet injected into the window in a browser, browser extension, or mobile app, you no longer need to publish an adapter at all. You don't need to open a PR to MWA or the Wallet Standard.
> 
> We are no longer accepting contributions for new adapters of this type. Bug fixes to existing adapters are welcome, but new features should be implemented using the MWA and Wallet Standard interfaces.

I'm interested to work with you to get apps to update to a version of their dependencies that will automatically support your wallet (and other wallets that already use the Wallet Standard). I'll continue to follow up!

PR #834, opened 2023-10-09
Initialize the MetaMask extension provider internally

vsakos:
This PR uses an internal postMessage based inpage provider for communicating with the MetaMask extension instead of using the injected `window.ethereum` interface.


jordansexton (changes requested):
I think this needs some changes. I'm going to create a branch with a simplified approach.

ZhengYuTay:
@jordansexton, this commit broke usage of Helium wallet if dApp have Solflare adapter and connecting through Helium

PR #836, opened 2023-10-12
feat: add bitgetWallet, upgraded bitKeepWallet

kisenyang:
Hi, Team Solana-Labs,

This PR adds bitget wallet to the list of supported wallets and upgrades bitkeep wallet to the bitget wallet to make it easier for our users to connect to wallet-adapter dapps.

Please let me know if anything needs to be added or changed to satisfy your requirements.

Cheers,
Bitget Wallet team

### DIFF
--- a/packages/wallets/bitgetwallet/CHANGELOG.md
+++ b/packages/wallets/bitgetwallet/CHANGELOG.md
@@ -1,0 +1,37 @@
+# @solana/wallet-adapter-bitkeep
+
+## 0.3.19
+
+### Patch Changes
+
+-   Updated dependencies [a3d35a1]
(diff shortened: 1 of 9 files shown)


mcintyre94:
Hey there, [as mentioned in the wallets docs](https://github.com/solana-labs/wallet-adapter/blob/master/WALLET.md) we're not accepting new adapters for wallets that are injected into the window. These wallets should be implemented using Wallet Standard, so that they are available to all apps without any need to import their adapters. You can of course publish the adapter yourself and ask apps to include it, but I'd definitely encourage you to take advantage of Wallet Standard so your users can access your wallet in all apps they use. 

If you could open a PR that just modifies your existing BitKeep adapter I'm happy to review that for you.

PR #838, opened 2023-10-16
feat: upgraded bitKeepWallet

kisenyang:
Hi, Team Solana-Labs,

This PR upgrades bitkeep wallet to the bitget wallet to make it easier for our users to connect to wallet-adapter dapps.

Please let me know if anything needs to be added or changed to satisfy your requirements.

Cheers,
Bitget Wallet team

### DIFF
--- a/packages/wallets/bitkeep/package.json
+++ b/packages/wallets/bitkeep/package.json
@@ -38,6 +38,7 @@
         "@solana/wallet-adapter-base": "workspace:^"
     },
     "devDependencies": {
+        "@bitget-wallet/web3-sdk": "^0.0.8",
         "@solana/web3.js": "^1.77.3",
         "shx": "^0.3.4"
     }
--- a/packages/wallets/bitkeep/src/adapter.ts
+++ b/packages/wallets/bitkeep/src/adapter.ts
@@ -1,3 +1,5 @@
+import { WalletInfo, SolanaAdapter } from '@bitget-wallet/web3-sdk';
+import type { SolanaProvider } from '@bitget-wallet/web3-sdk';
 import type { WalletName } from '@solana/wallet-adapter-base';
 import {
     BaseMessageSignerWalletAdapter,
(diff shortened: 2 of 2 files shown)


packages/wallets/bitkeep/package.json
```diff
@@ -38,6 +38,7 @@
         "@solana/wallet-adapter-base": "workspace:^"
     },
     "devDependencies": {
+        "@bitget-wallet/web3-sdk": "^0.0.8",
```
jordaaash, line 41: Please move to dependencies, since this is used by the adapter code (not just build/test tooling).

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -1,3 +1,5 @@
+import { WalletInfo, SolanaAdapter } from '@bitget-wallet/web3-sdk';
+import type { SolanaProvider } from '@bitget-wallet/web3-sdk';
```
jordaaash, line 2: Please import only as types here. The actual library will need to be imported async (see comment below).

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -13,37 +15,26 @@ import {
...
-    signTransaction(transaction: Transaction): Promise<Transaction>;
-    signAllTransactions(transactions: Transaction[]): Promise<Transaction[]>;
-    signMessage(message: Uint8Array): Promise<{ signature: Uint8Array }>;
-}
-
-interface BitKeepWindow extends Window {
+interface BitgetWalletWindow extends Window {
     bitkeep?: {
-        solana?: BitKeepWallet;
+        solana?: SolanaProvider;
```
jordaaash, line 20: Is it still using `window.bitkeep`?

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -13,37 +15,26 @@ import {
...
     };
 }
 
-declare const window: BitKeepWindow;
+declare const window: BitgetWalletWindow;
 
 export interface BitKeepWalletAdapterConfig {}
 
-export const BitKeepWalletName = 'BitKeep' as WalletName<'BitKeep'>;
+export const BitgetWalletName = WalletInfo?.name as WalletName<'Bitget Wallet'>;
```
jordaaash, line 28: Please remove `Wallet` here -- this only shows up in a list of wallets, so we try to avoid repetition.

```suggestion
export const BitgetWalletName = WalletInfo?.name as WalletName<'Bitget'>;
```

jordaaash, line 28: This is a breaking change because `BitKeepWalletName` is exported. You may want to alias the old export.

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -87,9 +78,9 @@ export class BitKeepWalletAdapter extends BaseMessageSignerWalletAdapter {
 
             this._connecting = true;
 
+            const adapter = new SolanaAdapter();
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            const wallet = window.bitkeep!.solana!;
-
+            const wallet = adapter?.getProvider() || window.bitkeep!.solana!;
```
jordaaash, line 83: New load-time dependencies cannot be introduced, they may only be loaded on connection. If `window.bitkeep` is not being used, the SDK must be asynchronously imported. I recommend checking for `window.bitkeep.solana` first, then `await import(...)` for the SDK. Please see some of the other adapters for how they do this.

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -13,37 +15,26 @@ import {
...
+    name = BitgetWalletName;
+    url = WalletInfo?.homepage;
+    icon = WalletInfo?.logolist?.base64;
```
jordaaash, line 33: These cannot be changed this way, because these imports cannot be used except as types. Please see the below comment, and inline the new image. Please also include the SVG file, as it may need to be compressed.

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -13,37 +15,26 @@ import {
...
 
-declare const window: BitKeepWindow;
+declare const window: BitgetWalletWindow;
 
 export interface BitKeepWalletAdapterConfig {}
 
-export const BitKeepWalletName = 'BitKeep' as WalletName<'BitKeep'>;
+export const BitgetWalletName = WalletInfo?.name as WalletName<'Bitget Wallet'>;
 
 export class BitKeepWalletAdapter extends BaseMessageSignerWalletAdapter {
```
jordaaash, line 30: You may want to export / alias this as `BitgetWalletAdapter` also?

jordaaash (changes requested):
Thanks, this will need some changes to accept. My understanding is

1. The injected Bitkeep wallet browser extension or app still exists, but has been renamed to Bitget. You want to change the naming to Bitget without breaking any apps that use this.
2. If the extension is not detected, you want to load a third party SDK.

Please see the comments in the review for more detail on how to approach this.

packages/wallets/bitkeep/package.json
DetonateYin2077: Understood. I will set up a new PR to address this, as the colleague who originally filed the PR is no longer with the company.

PR #839, opened 2023-10-16
Add Trezor wallet adapter

gabrielKerekes:
This PR adds Trezor HW wallet adapter. Trezor Solana support is planned to be released on Nov 15 2023. Ideally we would have this PR merged beforehand so dApps could potentially prepare for the Trezor release and already be integrated on Nov 15.

At this point the current `@trezor/connect-web` package (without Solana support) is included in `@solana/wallet-adapter-trezor` - this will need to be updated when a beta version of `@trezor/connect-web` with Solana support is released - expected soon after Nov 1 and before Nov 15 and then updated again when the official version of `@trezor/connect-web` with Solana support is released.

Unofficial firmware builds should also be available between Nov 1 and Nov 15.

### How to test

Testing the Trezor adapter is a bit difficult since none of the Trezor components (Trezor Connect and Trezor firmware) have yet been released and thus both Connect and FW have to be setup locally and the local `@trezor/connect-web` has to be linked into this proejct. I'm open to scheduling a demo call to demonstrate the functionality on my local setup. Or we can provide detailed steps on how to locally setup, build and run both Connect and Trezor FW.


jordansexton:
Thanks! Excited for this release. I can't merge it if the build is failing though. I'm also not inclined to merge before it's usable, or at least easily testable.

gabrielKerekes:
Sure, makes sense. I will update the PR when a beta version of `@trezor/connect-web` is released and we can then discuss further.

Maybe you could at this point just take a look at the code if there aren't some obvious things which would need to be changed? 🙏

gabrielKerekes:
👋 The release of Solana support on Trezor has been moved to December 15th (sorry for not writing here earlier).

We've put together a doc which should help with the setup required to test the PR before December 15th. You can find the doc [here](https://docs.google.com/document/d/1pGntMv7ObigScfnKYB38ZQamNCsn5LbrqwUZrvfEVQ0/edit). Do you think you could give it a try to test the adapter? It'd be great if we could release this on December 15th or 16th - or ideally earlier but I think a pre-release isn't really possible/feasible until the Trezor release?

There might be a beta release of Trezor Connect before December 15th. Maybe this could be released with the beta Connect version before December 15th and then re-released with official Connect release after December 15th (e.g. on December 16th). What do you think?

The reason it'd be nice to release before December 15th is that dApps could potentially prepare for the release beforehand and release Trezor support sooner.

packages/wallets/trezor/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    async connect(): Promise<void> {
+        try {
+            if (this.connected || this.connecting) return;
+            if (this._readyState !== WalletReadyState.Loadable) throw new WalletNotReadyError();
+
+            this._connecting = true;
+
+            const connectUrl = this._connectUrl && this._connectUrl + (this._connectUrl.endsWith('/') ? '' : '/');
```
jordaaash, line 70: Can we do this in the constructor?

packages/wallets/trezor/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+
+            await TrezorConnect.init({
+                manifest: {
+                    email: 'maintainers@solana.foundation',
+                    appUrl: 'https://github.com/solana-labs/wallet-adapter',
+                },
```
jordaaash, line 76: How are these used?

packages/wallets/trezor/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+    async connect(): Promise<void> {
+        try {
+            if (this.connected || this.connecting) return;
+            if (this._readyState !== WalletReadyState.Loadable) throw new WalletNotReadyError();
+
+            this._connecting = true;
+
+            const connectUrl = this._connectUrl && this._connectUrl + (this._connectUrl.endsWith('/') ? '' : '/');
+
+            await TrezorConnect.init({
```
jordaaash, line 72: This dependency will need to be `async import`ed. Check out how some of the other adapters that load external libraries do it.

packages/wallets/trezor/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
+import type { WalletName } from '@solana/wallet-adapter-base';
+import TrezorConnect, { DEVICE, DEVICE_EVENT } from '@trezor/connect-web';
```
jordaaash, line 2: These will need to be imported as types only. External SDKs must be loaded async on connect.

packages/wallets/trezor/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+                throw new WalletPublicKeyError(result.payload?.error, result.payload);
+            }
+
+            const publicKey = result.payload.publicKey;
+
+            TrezorConnect.on(DEVICE_EVENT, (event: any) => {
+                if (event.type === DEVICE.DISCONNECT) {
+                    this._disconnected();
+                }
+            });
```
jordaaash, line 105: Since these look to be just strings, we can hardcode them.

packages/wallets/trezor/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+
+            this._publicKey = new PublicKey(Buffer.from(publicKey, 'hex'));
```
jordaaash, line 107: Can you get your API to return a `Uint8Array` here? Solana public keys never use hex strings, and I'd prefer not to add the Buffer dependency here, we try to avoid it wherever possible. It's already been eliminated from the new version of web3.js we're moving to.

packages/wallets/trezor/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+
+            const serializedTransaction = isVersionedTransaction(transaction)
+                ? transaction.message.serialize()
+                : transaction.serializeMessage();
+
+            let result;
+            try {
+                result = await TrezorConnect.solanaSignTransaction({
+                    path: this._derivationPath,
+                    serializedTx: Buffer.from(serializedTransaction).toString('hex'),
```
jordaaash, line 143: Same here, we never use hex for data on Solana. Base64 strings are okay if it cannot be a Uint8Array.

packages/wallets/trezor/src/adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+                });
+            } catch (error: any) {
+                throw new WalletSignTransactionError(error?.message, error);
+            }
+
+            if (!result.success) {
+                throw new WalletSignTransactionError(result.payload?.error, result.payload);
+            }
+
+            transaction.addSignature(publicKey, Buffer.from(result.payload.signature, 'hex'));
```
jordaaash, line 153: "

jordaaash (changes requested):
Mostly looks good! Will need to async load the external SDK and not depend on Buffer.

packages/wallets/trezor/src/adapter.ts
gabrielKerekes: > Can you get your API to return a Uint8Array here?

Sadly I don't think so. TrezorConnect accepts and returns bytes as hex for other blockchains as well. Also the Connect implementation has already been approved so I'd rather not introduce changes at this point - since we're very close to the release.

For now I'd suggest no action since I see the Solana SDK still returns a `Buffer` for `transaction.message.serialize()` and `transaction.serializeMessage()` and `transaction.addSignature` takes a `Buffer` as parameter.

What version of web3 will you be moving to? 2.0.0?

When Solana SDK no longer works with Buffer, I'd suggest we introduce local utils for hex <-> Uint8Array conversion and that would be all that is required.

What do you think?

packages/wallets/trezor/src/adapter.ts
jordaaash: Okay, then please update these to your own email and URL. We don't provide support for anything except for bugs with the adapter itself.

packages/wallets/trezor/src/adapter.ts
jordaaash: Okay, I think that's fine for now. I just don't want to end up importing a second copy of `Buffer`. I'll take a look at how this is getting built.

jordaaash: (I resolved the other two comments I made, I'll leave this one pending for now, but I don't think there's any action you need to take on it.)

packages/wallets/trezor/src/adapter.ts
gabrielKerekes: https://github.com/solana-labs/wallet-adapter/pull/839/commits/12f08375621bd1d53c4b73bd1eecfa121ddf23bd

gabrielKerekes:
In https://github.com/solana-labs/wallet-adapter/pull/839/commits/cbb14e75887b5898f2b4a793fd1e17161ce4e594 I've updated Connect to a version with Solana support as it's been released yesterday 😉 However I don't think we want the `skipLibCheck` flags there but sadly I don't know how to properly get rid of them - the build fails due to errors like this one (possibly a clash of @types/web and @typescript but I'm not sure how to resolve this):

```
../../../node_modules/.pnpm/@types+web@0.0.119/node_modules/@types/web/index.d.ts:27428:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'visualViewport' must be of type 'VisualViewport', but here has type 'VisualViewport | null'.

27428 declare var visualViewport: VisualViewport | null;
                  ~~~~~~~~~~~~~~

  ../../../node_modules/.pnpm/typescript@4.7.4/node_modules/typescript/lib/lib.dom.d.ts:17737:13
    17737 declare var visualViewport: VisualViewport;
                      ~~~~~~~~~~~~~~
    'visualViewport' was also declared here.
```

I'll try to figure out if there is a simple way for you to test this end to end now.

gabrielKerekes:
> I'll try to figure out if there is a simple way for you to test this end to end now.

If you're on Apple Sillicon I can't help you right now (because ARM Trezor firmware builds are failing), the only way to test is by building the Trezor emulator yourself - as mentioned in [this doc](https://docs.google.com/document/d/1pGntMv7ObigScfnKYB38ZQamNCsn5LbrqwUZrvfEVQ0/edit#heading=h.bbtuw4hevra1).

If you're on Linux you can use [trezor-usr-env](https://github.com/trezor/trezor-user-env). It should only be required to clone the repo and run `./run.sh` in the root of that repo and then pasting this CI job link into input field in the `Emulator` (next to the `Start from URL` button. `trezor-usr-env` requires docker though.

Or well, we can wait until Wednesday (Dec 13 2023) when the FW should be officially released.

gabrielKerekes:
👋 So Solana support on Trezor has officially been released. Trezor devices can be updated via [Trezor Suite](https://trezor.io/trezor-suite). For testing you can still [build the emulator](https://docs.google.com/document/d/1pGntMv7ObigScfnKYB38ZQamNCsn5LbrqwUZrvfEVQ0/edit#heading=h.bbtuw4hevra1) and test against that.

Please let me know what actions are required to get this merged and released or how I could help to achieve that 🙏

jordaaash:
Thanks! Catching up on this now.

jordaaash:
@gabrielKerekes I've pushed an update to https://github.com/solana-labs/wallet-adapter/pull/876 because this PR's branch doesn't allow commits by maintainers.

However, the issue that's being caused by `@types/web` is unresolved. It appears to be redeclaring global types of the window object that TypeScript handles.

Because this is a transitive dependency of `@trezor/blockchain-link`, I can't easily do anything about it.

Looking at the dependencies for that project:

```json
  "devDependencies": {
    "@trezor/e2e-utils": "1.0.0",
    "@trezor/type-utils": "1.0.2",
    "fs-extra": "^11.1.1",
    "html-webpack-plugin": "^5.5.3",
    "jest": "^26.6.3",
    "rimraf": "^5.0.5",
    "tiny-worker": "^2.3.0",
    "tsx": "^3.14.0",
    "typescript": "5.3.2",
    "webpack": "^5.89.0",
    "webpack-cli": "^5.1.4",
    "webpack-dev-server": "^4.15.1",
    "worker-loader": "^3.0.8"
  },
  "dependencies": {
    "@solana/buffer-layout": "^4.0.1",
    "@solana/web3.js": "^1.87.6",
    "@trezor/blockchain-link-types": "1.0.8",
    "@trezor/blockchain-link-utils": "1.0.9",
    "@trezor/utils": "9.0.15",
    "@trezor/utxo-lib": "2.0.1",
    "@types/web": "^0.0.119",
    "bignumber.js": "^9.1.1",
    "events": "^3.3.0",
    "ripple-lib": "^1.10.1",
    "socks-proxy-agent": "6.1.1",
    "ws": "7.5.9"
  }
```

I can see that `@types/web` is a dependency. `@types` packages should never be dependencies, only dev dependencies, and they should only be used if type definitions are not provided for a package that is used as another dependency, which this doesn't seem to be.

https://www.npmjs.com/package/@types/web says

> @types/web is also included inside TypeScript, available as dom in the [lib](https://www.typescriptlang.org/tsconfig#lib) section and included in projects by default.

I can't see why this is required and I think it may be superfluous. What is it used for in this dependency, and can it be removed?

jordaaash:
Okay, some good news -- I was able to override this dependency here https://github.com/solana-labs/wallet-adapter/pull/876/commits/e6788c52539d60cddfeae3485ab2a38fb7949633

Can you install/build/test with your device or emulator against https://github.com/solana-labs/wallet-adapter/pull/876?

gabrielKerekes:
(I'm responding in this PR since the original discussion is here)

Thanks for the updates.

> I can see that @types/web is a dependency. @types packages should never be dependencies, only dev dependencies

I'll notify Trezor devs regarding this.

> Can you install/build/test with your device or emulator against https://github.com/solana-labs/wallet-adapter/pull/876?

Seems to work just fine 👍

jordaaash:
Merged #876 which incorporates these commits. Thanks!

jordaaash:
Released via https://github.com/solana-labs/wallet-adapter/pull/878

PR #848, opened 2023-11-29
chore: constrain the `connection` input of `sendTransaction` to the minimum interface actually required

steveluscher:
The `sendTransaction()` method takes in a `Connection` from `@solana/web3.js` but actually requires very little of it; just `commitment`, `sendRawTransaction()`, and `getLatestBlockhash()`.

In this PR we constrain the interface to exactly that and no more. This will make it easier to supply _non_ `@solana/web3.js` implementations at that callsite, so long as they can fulfil the same contract.

### DIFF
--- a/packages/core/base/src/adapter.ts
+++ b/packages/core/base/src/adapter.ts
@@ -1,9 +1,23 @@
-import type { Connection, PublicKey, SendOptions, Signer, Transaction, TransactionSignature } from '@solana/web3.js';
+import type { Commitment, PublicKey, SendOptions, Signer, Transaction, TransactionSignature } from '@solana/web3.js';
 import EventEmitter from 'eventemitter3';
 import { type WalletError, WalletNotConnectedError } from './errors.js';
 import type { SupportedTransactionVersions, TransactionOrVersionedTransaction } from './transaction.js';
 
 export { EventEmitter };
+
+type BlockhashFetcher = (config?: {
+    commitment?: Commitment;
+    minContextSlot?: number;
+}) => Promise<{ blockhash: string }>;
+
+export interface ConnectionContext {
+    commitment?: Commitment;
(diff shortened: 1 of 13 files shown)


packages/core/base/src/adapter.ts
```diff
@@ -1,10 +1,24 @@
...
+
+export interface ConnectionContext {
+    commitment?: Commitment;
+    getLatestBlockhash: BlockhashFetcher;
+    sendRawTransaction(
+        rawTransaction: Buffer | Uint8Array | Array<number>,
+        options?: SendOptions
+    ): Promise<TransactionSignature>;
+}
+
```
steveluscher, line 21: Start here.

packages/wallets/censo/src/adapter.ts
```diff
@@ -124,15 +124,20 @@ export class CensoWalletAdapter extends BaseSignerWalletAdapter {
...
             if (!wallet) throw new WalletNotConnectedError();
 
             try {
-                return await wallet.sendTransaction(transaction, connection, options);
+                return await wallet.sendTransaction(
+                    transaction,
+                    // @ts-ignore
+                    undefined, // This parameter is not used. See https://www.npmjs.com/package/@censo-custody/solana-wallet-adapter?activeTab=code
+                    options
+                );
```
steveluscher, line 140: Are you OK with this, @bholzman, or does `window.censo` actually make use of this parameter? It doesn't look like `@censo-custody/solana-wallet-adapter` does.

bholzman: Yes. The censo solana wallet was eol’d, this entire adapter can be removed

steveluscher: ty.

steveluscher:
Rebased. I still think this is worth doing. Are you opposed to this in any way @jordaaash?

PR #849, opened 2023-11-29
feat: remove Censo adapter

steveluscher:
https://github.com/solana-labs/wallet-adapter/pull/848#discussion_r1408602661

PR #840, opened 2023-10-21
Updates & revamps ParticleAdapter

TABASCOatw:
`ParticleAdapter` is the adapter for an MPC-TSS Wallet-as-a-Service provider, [Particle Network](https://particle.network).

This adapter was quite out of date, and thus configuration (through `config` on a new instance of `ParticleAdapter`) didn't work and was incompatible with the current Particle dashboard. This has been fixed through:
- Updating dependency `@particle-network/solana-wallet`
- Slight rework of core adapter logic to solve outstanding issues

This has been tested and is functional, enabling the standard level of configuration possible with modern implementations of Particle (customizing UI, choosing wallet popup options, etc.)

In addition to this, functionality has been introduced to select `preferredAuthType` when adding `ParticleAdapter` to a list of wallets. `preferredAuthType` can either be a string or an object containing `type` and boolean `setAsDisplay`

`preferredAuthType` dictates if `ParticleAdapter` automatically routes to a specific mechanism of wallet authentication (social login). If `preferredAuthType` is set to an object with `setAsDisplay` as true, Particle will display in the wallet connection list corresponding to the selected social login provider, indicating the shortcut. Otherwise if `setAsDisplay` is set to false or if `preferredAuthType` is undefined/a string, authentication will be handled within a generic `Particle` option according to the status of `preferredAuthType`.

PR #851, opened 2023-11-30
Run build-and-test workflow on all PRs to master

mcintyre94:
Some PRs have run the build-and-test job (eg https://github.com/solana-labs/wallet-adapter/pull/849 + https://github.com/solana-labs/wallet-adapter/pull/841), but some haven't (eg https://github.com/solana-labs/wallet-adapter/pull/840 + https://github.com/solana-labs/wallet-adapter/pull/838)

I'm unsure why this happens, but one guess is the missing `branches: [master]` on the `pull_request` trigger. 

This matches the trigger in eg web3js: https://github.com/solana-labs/solana-web3.js/blob/fdc85b98edc4bd7a74c5bd10128d5397bce368a7/.github/workflows/pull-requests.yml#L3-L5

The github docs also only seem to give examples that include a filter, eg https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore 

I'm hoping this will make all PRs run the build-and-test workflow!

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [master]
   pull_request:
+    branches: [master]
 
 jobs:
   build-and-test:


jordaaash:
Good call!

PR #857, opened 2023-12-06
Feat: Upgrade Bitkeep Wallet Adapter

DetonateYin2077:
Hi solana-labs team,

This PR upgrade the Bitkeep adapter as 'Bitkeep Wallet' has been rebranded to 'Bitget Wallet'

**Type of Change**:
> New Feature

**Checklist**:

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
- [x] I have added comments to my code as a warning against using deprecated exports.

**Additional Notes**:

Hi @jordaaash , 

This PR is downstream of #838. The modifications you suggested for #838 have been implemented in this PR.
Could you please review it again? Feel free to raise any concerns at any time. 
Thanks again for your assistance.

### DIFF
--- a/packages/wallets/bitkeep/src/adapter.ts
+++ b/packages/wallets/bitkeep/src/adapter.ts
@@ -1,4 +1,3 @@
-import { WalletInfo, SolanaAdapter } from '@bitget-wallet/web3-sdk';
 import type { SolanaProvider } from '@bitget-wallet/web3-sdk';
 import type { WalletName } from '@solana/wallet-adapter-base';
 import {
@@ -11,6 +10,8 @@ import {
     WalletPublicKeyError,
     WalletReadyState,
     WalletSignTransactionError,
+    WalletLoadError,
+    WalletConfigError
 } from '@solana/wallet-adapter-base';
 import type { Transaction } from '@solana/web3.js';
 import { PublicKey } from '@solana/web3.js';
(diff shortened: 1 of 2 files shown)


packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -9,10 +10,185 @@ import {
...
     WalletSignTransactionError,
+    WalletLoadError,
+    WalletConfigError
 } from '@solana/wallet-adapter-base';
 import type { Transaction } from '@solana/web3.js';
 import { PublicKey } from '@solana/web3.js';
 
+interface BitgetWalletWindow extends Window {
+    bitkeep?: {
+        solana?: SolanaProvider;
```
DetonateYin2077, line 21: The `SolanaProbider` is still using `window.bitkeep`

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -9,10 +10,185 @@ import {
...
+    };
+}
+
+declare const window: BitgetWalletWindow;
+
+export interface BitgetWalletAdapterConfig {}
+
+export const BitgetWalletName = 'Bitget Wallet' as WalletName<'Bitget Wallet'>;
```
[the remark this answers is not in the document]
jordaaash: Yes, in this case, even for wallets that include the "Wallet" name, we remove all of these. The context the names are displayed in is always in a list of wallets and they are all the names of wallets, so we don't want to say "Wallet" for each one.

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -9,10 +10,185 @@ import {
...
+    readonly supportedTransactionVersions = null;
```
jordaaash, line 35: Hmm, do you not support `VersionedTransaction` on Solana? This will make your wallet incompatible with most of the major Solana apps.

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -9,10 +10,185 @@ import {
...
+                throw new WalletLoadError(error?.message, error);
+            }
+
+            let wallet;
+            try {
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                wallet = window.bitkeep!.solana! || new adapter?.getProvider();
+            } catch (error: any) {
+                throw new WalletConfigError(error?.message, error);
+            }
```
jordaaash, line 95: There are a few issues here --

First, this code will never be reached unless `window.bitkeep.solana.isBitKeep` is true, because the `readyState` will not be "installed". Check out the Solflare adapter for a wallet that can both be injected and loaded with an SDK for how to handle this.

Second, you probably don't want to load the adapter SDK if the wallet is already installed, right? It's only used in the event that it isn't installed.

packages/wallets/bitkeep/src/adapter.ts
```diff
@@ -181,3 +360,4 @@ export class BitKeepWalletAdapter extends BaseMessageSignerWalletAdapter {
         }
     }
 }
+
```
jordaaash, line 363: All of this code should be simplified -- these shouldn't be literal copies of the code, they should just be exported variables and types that alias the names of the `Bitget...` ones so they don't break the exports.

jordaaash:
Thanks, gave it a review. 2 major issues --

1. Detection/loading isn't going to work as-is. It needs to conditionally async load the SDK _iff_ the extension isn't detected, and the readyState logic needs to be updated accordingly to reflect that it's installed or loadable. See Solflare.
2. Please minimally update the names and reexport the old names as needed, rather than copying the code.

PR #866, opened 2023-12-24
Prevents multiple bindings of the disconnect event handler

shiralwz:
**Issue:**
 if the wallet emits the disconnect event more than once, the `_disconnected` method will be added multiple times as an event listener. This might lead to unexpected issues.

**Fix:**
Added  `_onDisconnectBound` to store the bound `_disconnected` method. In the connect method, the `_disconnected` method is bound and set to `_onDisconnectBound` only if it hasn't been set already. This prevents multiple bindings of the same event handler.

### DIFF
--- a/packages/wallets/alpha/src/adapter.ts
+++ b/packages/wallets/alpha/src/adapter.ts
@@ -62,6 +62,7 @@ export class AlphaWalletAdapter extends BaseMessageSignerWalletAdapter {
         typeof window === 'undefined' || typeof document === 'undefined'
             ? WalletReadyState.Unsupported
             : WalletReadyState.NotDetected;
+    private _onDisconnectBound: (() => void) | null = null;
 
     constructor(config: AlphaWalletAdapterConfig = {}) {
         super();
@@ -106,6 +107,7 @@ export class AlphaWalletAdapter extends BaseMessageSignerWalletAdapter {
 
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             const wallet = window.alpha!;
+            if (!wallet) throw new WalletNotReadyError();
 
             if (!wallet.isConnected) {
                 try {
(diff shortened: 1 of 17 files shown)


jordaaash:
Thanks for the PR. I'm not sure I understand the sequence of events here though.

Generally, we add one listener on connect, and when the wallet emits the disconnect event, or we disconnect it ourselves, the first thing we do is remove the event listener for that event, and then proceed to tear down the adapter's internal state. Since this occurs synchronously, it shouldn't be possible for this event to be handled multiple times.

Can you explain in what conditions you're seeing this issue?

packages/wallets/alpha/src/adapter.ts
```diff
@@ -106,6 +107,7 @@ export class AlphaWalletAdapter extends BaseMessageSignerWalletAdapter {
 
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             const wallet = window.alpha!;
+            if (!wallet) throw new WalletNotReadyError();
```
jordaaash, line 110: The reason these property checks aren't _strictly_ necessary (and why we use the `!` non-null assertion) is because we already checked for `window.alpha` in the polling mechanism of the constructor to update the `readyState` to `installed`. Since we already checked this in `connect` function, we are relying on it not to have changed, but I'm not aware of a case where this assumption is faulty.

packages/wallets/avana/src/adapter.ts
```diff
@@ -126,7 +128,10 @@ export class AvanaWalletAdapter extends BaseMessageSignerWalletAdapter {
                 throw new WalletPublicKeyError(error?.message, error);
             }
 
-            wallet.on('disconnect', this._disconnected);
+            if (!this._onDisconnectBound) {
+                this._onDisconnectBound = this._disconnected.bind(this);
+                wallet.on('disconnect', this._onDisconnectBound);
+            }
```
jordaaash, line 134: The reason this listener should never get added twice is because we always check if the adapter is `connected` or `connecting` first in the `connect` function. Only by disconnecting (and thereby removing the listener) can it be added again.

packages/wallets/avana/src/adapter.ts
```diff
@@ -143,7 +148,10 @@ export class AvanaWalletAdapter extends BaseMessageSignerWalletAdapter {
     async disconnect(): Promise<void> {
         const wallet = this._wallet;
         if (wallet) {
-            wallet.off('disconnect', this._disconnected);
+            if (this._onDisconnectBound) {
+                wallet.off('disconnect', this._onDisconnectBound);
+                this._onDisconnectBound = null;
+            }
```
jordaaash, line 154: This merely complicates the existing check since we are relying on the checks in the `connect` function to guarantee only having one listener. The key thing is to prevent multiple listeners from being added. Is there a reason to believe we aren't doing that in all cases?

PR #873, opened 2023-12-30
Adding better and more intuitive UX to wallet adapter connecting

blockiosaurus:
This PR modifies the UX flow during wallet connection to provide a familiar and more intuitive feedback.


jordaaash:
Thanks for not waiting until April 1st to do this important work.

PR #906, opened 2024-02-27
Coinbase Wallet Adapter: Add supported transaction versions

willn-cb:
This PR updates the Coinbase Wallet adapter to properly list the supported transaction versions. Versioned transaction support was recently added to Coinbase Wallet.

I also cleaned up the TS a bit on the interface.

Tested locally that the correct tx versions are returned and that signing works as expected.

### DIFF
--- a/.changeset/olive-turtles-walk.md
+++ b/.changeset/olive-turtles-walk.md
@@ -1,0 +1,5 @@
+---
+'@solana/wallet-adapter-coinbase': patch
+---
+
+Add support for versioned transactions to Coinbase Wallet adapter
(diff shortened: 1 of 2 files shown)


jordaaash:
Thanks! Can I get you to consider [implementing the Wallet Standard](https://github.com/anza-xyz/wallet-adapter/blob/master/WALLET.md) in Coinbase Wallet? It's a pretty quick nonbreaking change that can be made with no dependencies and it will make Coinbase Wallet work on most Solana apps without requiring an adapter at all.

willn-cb:
Yeah, it's on the roadmap! Not sure of timeline but it shouldn't be too long :)

jordaaash:
Cool! I added a changeset and ran the tests, can you check on the test failure?

willn-cb:
@jordaaash – fixed the type issue, `build` ran fine locally. Don't think I have permission to run it again here

PR #914, opened 2024-03-02
Add root resolution for ngraveio/bc-ur

mcintyre94:
Second attempt at #910 

I thought the `overrides` field was the pnpm one, but turns out its top-level. I guess pnpm uses the top-level `resolutions` and we need the same overrides in both. This has added it to the `overrides` in pnpm-lock which I've also checked in. Hopefully this sorts it now!

### DIFF
--- a/.changeset/popular-fans-battle.md
+++ b/.changeset/popular-fans-battle.md
@@ -1,0 +1,5 @@
+---
+'@solana/wallet-adapter-keystone': patch
+---
+
+Add a resolution for ngraveio/bc-ur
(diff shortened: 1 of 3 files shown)


BankkRoll:
Following around the issues and PRs here as well as `bc-ur`.. Even using the latest version that included this PR for `@ngraveio/bc-ur": "1.1.6` this still doesn't fix the error and still requires a override as shown below not sure why but figured id drop this here for info


```
  "dependencies": {
    "@solana/wallet-adapter-base": "^0.9.23",
    "@solana/wallet-adapter-react": "^0.15.35",
    "@solana/wallet-adapter-react-ui": "^0.9.35",
    // omitted remaining for brevity
  },
  "overrides": {
    "@ngraveio/bc-ur": "1.1.6"
  },
```

PR #927, opened 2024-03-14
fix: fix Trezor adapter on Firefox browsers

gabrielKerekes:
We received a bug report that the Trezor wallet adapter was not available on Firefox. This PR fixes the Trezor adapter on Firefox (and Safari as well although Safari is not supported by Trezor Connect anyways).

Sorry, I didn't notice the `!navigator.hid` check before which I copied over from Ledger. It is not needed for the Trezor as Trezor doesn't use HID.

**How to reproduce**

Try the currently released Trezor adapter on Firefox browsers. See e.g. [here](https://jup.ag/) that Trezor is not available on Firefox while it is available on Chrome.

**How to test**

Try the updated Trezor adapter e.g. in the example in this repo. Trezor should be available on both Firefox and Chrome.

### DIFF
--- a/.changeset/silly-rats-accept.md
+++ b/.changeset/silly-rats-accept.md
@@ -1,0 +1,5 @@
+---
+'@solana/wallet-adapter-trezor': patch
+---
+
+Fix Trezor adapter on Firefox browsers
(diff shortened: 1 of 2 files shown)


mcintyre94:
Looks good! Could you create a changeset for this please? If you just run `npx changeset` at the root of the repo on this branch it should identify the trezor package as the only change, just make it a patch version and use your commit message as the summary. Then we'll automatically release a fixed version of the package when this is merged :)

gabrielKerekes:
Is https://github.com/anza-xyz/wallet-adapter/pull/927/commits/f159d224124de11fda7a7deb09cd1e948037dabb alright? I amended the previous commit and force-pushed with the changeset.

Pjrich1313 (commented):
Run debug and update nobes

PR #931, opened 2024-03-19
XDEFI Wallet Adapter: Update XDEFIWalletName

maxlm-devico:
Match name in adapter to XDEFI Wallet extension name. This fixes issue when XDEFI wallet listed twice in connect wallet modal
image

### DIFF
--- a/packages/wallets/xdefi/src/adapter.ts
+++ b/packages/wallets/xdefi/src/adapter.ts
@@ -46,7 +46,7 @@ declare const window: XDEFIWalletWindow;
 
 export interface XDEFIWalletAdapterConfig {}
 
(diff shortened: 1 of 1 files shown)


jordaaash:
We intentionally avoid superfluous "Wallet" suffixes from adapter names since they need to be short and this becomes repetitive when displayed in a list of wallets.

maxlm-devico:
> We intentionally avoid superfluous "Wallet" suffixes from adapter names since they need to be short and this becomes repetitive when displayed in a list of wallets.

It must match to the wallet name it detects or it will create a dupe which becomes really repetitive. 
Wallet name in adapter must come with "Wallet" suffix in order to match extension name and avoid dupe in list as provided on screenshot

PR #932, opened 2024-03-20
Bitget Wallet is the right brand name

EndlessJoke:
update brand name


jordaaash:
We intentionally avoid superfluous "Wallet" suffixes from adapter names since they need to be short and this becomes repetitive when displayed in a list of wallets.

PR #941, opened 2024-04-05
add type for connect function

sithu-go:
connect function type argument and it returns publicKey.


jordaaash:
This isn't a change we intend to make. Wallets don't generally support arguments to the `connect` function anyway.

PR #946, opened 2024-04-15
support gate wallet

vdgwerdfgCell:


jordaaash:
Thanks, but we are not accepting new adapters for wallets that can support the Wallet Standard. Please check [this guide](https://github.com/anza-xyz/wallet-standard/blob/master/WALLET.md) for integrating the Wallet Standard instead.

PR #960, opened 2024-05-09
Allow Disconnect of MWA 

alex-fung:
Currently, if other wallets are available on a dApp on Android, the only one that can be used is the MWA. This PR updates the `WalletProvider` so that it won't be auto-selected and can be disconnected if other wallets are available.

Comments posted at https://github.com/alex-fung/wallet-adapter/pull/1 are already addressed (opened against wrong branch)

**Before:**
https://github.com/anza-xyz/wallet-adapter/assets/2409008/19a522f9-d8f1-4ad9-8374-03ac8ca35de8

**After:**
With MWA and another wallet:
https://github.com/anza-xyz/wallet-adapter/assets/2409008/36dca9c4-b585-4017-a84a-0fadeb8c10eb
With just MWA:
https://github.com/anza-xyz/wallet-adapter/assets/2409008/da684467-2791-428d-8783-3a9c5485542c

### DIFF
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -159,8 +159,7 @@ export function WalletProvider({
     const handleConnectError = useCallback(() => {
         if (
             adapter &&
-            adapter.name !== SolanaMobileWalletAdapterWalletName &&
-            adaptersWithStandardAdapters.length === 0
+            (adapter.name !== SolanaMobileWalletAdapterWalletName || adaptersWithStandardAdapters.length !== 0)
         ) {
             // If any error happens while connecting, unset the adapter.
             changeWallet(null);
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -89,7 +89,14 @@ export function WalletProvider({
     const changeWallet = useCallback(
         (nextWalletName: WalletName<string> | null) => {
             if (walletName === nextWalletName) return;
-            if (adapter) {
+            if (
+                adapter &&
+                // Selecting a wallet other than the mobile wallet adapter is not
+                // sufficient reason to call `disconnect` on the mobile wallet adapter.
+                // Calling `disconnect` on the mobile wallet adapter causes the entire
+                // authorization store to be wiped.
+                adapter.name !== SolanaMobileWalletAdapterWalletName
+            ) {
                 adapter.disconnect();
             }
             setWalletName(nextWalletName);
@@ -100,6 +107,13 @@ export function WalletProvider({
         if (!adapter) return;
         function handleDisconnect() {
             if (isUnloadingRef.current) return;
+            // Leave the adapter selected in the event of a disconnection.
+            if (
+                walletName === SolanaMobileWalletAdapterWalletName &&
+                adaptersWithStandardAdapters.length === 0 &&
+                getIsMobile(adaptersWithStandardAdapters)
+            )
+                return;
             setWalletName(null);
         }
         adapter.on('disconnect', handleDisconnect);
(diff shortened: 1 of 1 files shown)


Funkatronics (commented):
Reviewed and approved by me, but I am not a maintainer here unfortunately. Will need input from @jordaaash or @steveluscher

alirezaayande92:
> Currently, if other wallets are available on a dApp on Android, the only one that can be used is the MWA. This PR updates the `WalletProvider` so that it won't be auto-selected and can be disconnected if other wallets are available.
> 
> Comments posted at https://github.com/alex-fung/wallet-adapter/pull/1 are already addressed (opened against wrong branch)

steveluscher (changes requested):
Understood; thanks for this!

I think the original presumption was that there are _no_ wallets that work (or yield a good experience) on mobile web, but I presume TipLink is out there to prove that wrong.

Can you add, to this PR, a quick screencast of how your loadable wallet works so that folks can understand the motivation for this change?

packages/core/react/src/WalletProvider.tsx
```diff
@@ -78,7 +78,9 @@ export function WalletProvider({
     }, [adaptersWithStandardAdapters, mobileWalletAdapter]);
     const [walletName, setWalletName] = useLocalStorage<WalletName | null>(
         localStorageKey,
-        getIsMobile(adaptersWithStandardAdapters) ? SolanaMobileWalletAdapterWalletName : null
+        adaptersWithStandardAdapters.length === 0 && getIsMobile(adaptersWithStandardAdapters)
```
jordaaash, line 81: Hmm. This has the effect of, if _any adapter_ is included, to not select MWA. The problem with this in practice is that almost all dapps will include adapters, even ones they don't need because they've been deprecated by Standard Wallets, or ones that simply don't work at all on the platform. This is especially true because most dapps I've seen don't use separate mobile builds with conditional compilation of included adapters.

This basically means that it's always going to be `false`, and we're never even going to do the `getIsMobile` check. Now, I'm probably fine with eliminating all preference for MWA, but I want to be clear that's what we're doing here, and if so, we might as well just do it outright and remove these checks.

jordaaash (changes requested):

packages/core/react/src/WalletProvider.tsx
```diff
@@ -76,10 +76,7 @@ export function WalletProvider({
         }
         return [mobileWalletAdapter, ...adaptersWithStandardAdapters];
     }, [adaptersWithStandardAdapters, mobileWalletAdapter]);
-    const [walletName, setWalletName] = useLocalStorage<WalletName | null>(
-        localStorageKey,
-        getIsMobile(adaptersWithStandardAdapters) ? SolanaMobileWalletAdapterWalletName : null
-    );
+    const [walletName, setWalletName] = useLocalStorage<WalletName | null>(localStorageKey, null);
```
jordaaash, line 79: Not sure there's anything good to be done about this, but if the user has opened the app before, I think it'll probably still have MWA in local storage and selected by default.

packages/core/react/src/WalletProvider.tsx
```diff
@@ -76,10 +76,7 @@ export function WalletProvider({
...
     const adapter = useMemo(
         () => adaptersWithMobileWalletAdapter.find((a) => a.name === walletName) ?? null,
```
jordaaash, line 81: One possibility for this: if the _only_ item in `adaptersWithMobileWalletAdapter` is MWA, could we select it by default?

This preserves the existing behavior if the user is on Android mobile (so no other wallets are being injected), the app is passing `wallets={[]}`, and no other Standard Wallets are registered. I don't know how common this is, but it might not hurt?

jordaaash (commented):
Thank you for the new video, this looks good. I left a comment in the code, but if MWA is the only adapter (or maybe if there's only one adapter in general) we could select it by default. We would have to consider if this has any other unintended effects on the state. If it does we can scrap the idea and a wallet will always need to be selected by name.

packages/core/react/src/WalletProvider.tsx
alex-fung: thats okay to us! That behavior exists for any adapter so I don't believe we mind

alex-fung:
Updated so that if only MWA is available, it basically locks it as the adapter

MWA only:
https://github.com/user-attachments/assets/3bd0afab-9e46-4397-9a44-97cea086b47c

Other adapters as well:
https://github.com/user-attachments/assets/7d8df7ea-fedb-4043-98e9-9118fde6c625

jordaaash:

jordaaash:
RIP, tests failing: https://github.com/anza-xyz/wallet-adapter/actions/runs/10532446524/job/29186623827

packages/core/react/src/WalletProvider.tsx
reneromeroverdugo1982: Fix

PR #967, opened 2024-05-18
Fix: BaseWalletConnection doesn't pass thru classname prop

asktree:
The BaseWalletConnection component is currently overwriting any className prop. I made it not do this. If this is intentional, then className should not be among its props.


ayushrungta8:
Is this planned to be fixed in a release soon? I see its almost a year old

PR #975, opened 2024-05-28
Adds labels as a prop to WalletConnectButton.

techiejd:
- Allows for i18n.
- Allows for customization.
- Both the labels object and the property are optional.
- Dependency inversion is good practice.
- It didn't make sense for labels to be a constant when they're the API one level below expects it as a prop.


techiejd:
I see now I should be using BaseWalletMultiButton. Better documentation would be nice.

PR #990, opened 2024-07-01
update trezor icon

vytick:
New trezor logo according to the brand.


jordaaash:

PR #997, opened 2024-07-05
add extra hooks

21e8:


packages/core/react/src/hooks.ts
```diff
@@ -0,0 +1,898 @@
...
+    const [error, setError] = useState<any>(null);
+    const { connection } = useConnection();
+
+    const fetchBalance = async () => {
+        if (!publicKey) return;
+        try {
+            setError(null);
+            setIsLoading(true);
+            const res = await connection.getBalance(publicKey);
+            setBalance(res);
```
glitch-txs, line 60: We should add a way to abort the call when the publicKey changes so the old calls get cleaned up.

Case:

- A wallet is connected
- The fetch function is triggered, this is called 1
- The wallet switches to another account
- The fetch function is triggered again, this is called 2
- Called 1 finishes and populates the states with old data

This is one of the things React Query prevents, it's a very common use case that may affect to any other hook as well

21e8: mhhh does this not mean we need some kind of abort controller inside the rpc calls?

glitch-txs: Maybe we can add something like this
https://react.dev/learn/removing-effect-dependencies#is-your-effect-doing-several-unrelated-things

jordaaash:
Thanks, the right place for these hooks is not in Wallet Adapter, but perhaps in the new web3.js. I've seen some preliminary work in that repo for these kind of hooks.

Issue #993, opened 2024-07-03
Add OKX wallet adapter


h0lme3:
Hi everybody,

We want to let users connect OKX wallet in our dApp and we are using `@solana/wallet-adapter-wallets` dep which is not listed yet

Can we add okx wallet like okx wallet adapter?

Thanks

glitch-txs:
Any plans on supporting the Wallet Standard?

puderty:
> Hi everybody,
> 
> We want to let users connect OKX wallet in our dApp and we are using `@solana/wallet-adapter-wallets` dep which is not listed yet
> 
> Can we add okx wallet like okx wallet adapter?
> 
> Thanks

which dApp and what platform is it? I think okx wallet already supported wallet adapter

h0lme3:
> which dApp and what platform is it? I think okx wallet already supported wallet adapter

Hmm I believe I am using the latest `@solana/wallet-adapter-wallets` version (`^0.19.32`) and it doesnt include okx wallet adapter I think at the moment.

h0lme3:
> Any plans on supporting the Wallet Standard?

Not now, just wrote for the feature request for now.

kAsky53:
> Any plans on supporting the Wallet Standard?

It already supports wallet standard.
I think it's still useful for installation though.

glitch-txs:
Then why do you need the adapter?

kAsky53:
> Then why do you need the adapter?

If I have not installed OKX wallet, I would still like it to be shown in available wallet list where I can click and install the wallet app.

glitch-txs:
what's your expected behavior when clicked by a user?

kAsky53:
> what's your expected behavior when clicked by a user?

It should redirect to `https://www.okx.com/web3` so that I can install the wallet app.

zwergdev:
Added `packages/wallets/okx/src/adapter.ts` in one of my projects. Everything is ok.

kAsky53:
> you could create a custom adapter and add name, icon and url as config args rather than one for each wallet. I guess it only needs to be filtered by WS and trigger the URL on `connect`

Would you mind sharing code snippet for it?

jordaaash:
Thanks, but we are not accepting new adapters for wallets that can support the Wallet Standard. Since OKX Wallet already does, this is unnecessary.

PR #1000, opened 2024-07-09
support OKX wallet adapter

kAsky53:
@jordaaash For issue #993


zwergdev:
Added [`packages/wallets/okx/src/adapter.ts`](https://github.com/anza-xyz/wallet-adapter/blob/d64d67fdc382ecbf6495c316eb199dd5bf91f232/packages/wallets/okx/src/adapter.ts) in one of my projects. Everything is ok.

jordaaash:
Thanks, but we are not accepting new adapters for wallets that can support the Wallet Standard. Since OKX Wallet already does, this is unnecessary.

PR #1001, opened 2024-07-11
add tonspack wallet adapter

Tonscriptions:
Add a telegram base solana web3 wallet Tonspack into wallet-adapter


jordaaash:
Thanks, but we aren't accepting new adapters. Please check [this guide](https://github.com/anza-xyz/wallet-standard/blob/master/WALLET.md) for integrating the Wallet Standard instead. You'll need to publish your own library that registers your Standard Wallet rather than distribute it in Wallet Adapter.

PR #1002, opened 2024-07-12
add versioned transaction support for keystone

soralit:
Hi team, I'm from Keystone team and our latest product Keystone 3 Pro has already supported Versioned transaction so I made this change to support it in `wallet-adapter`

CHANGES:
1. update @keystonehq/sol-keyring 's signTransaction method to accept a `txHex` as `Uint8Array` but not `Transaction` any longer to extend the flexibility.
2. update the `signTransaction` interface in keystone wallet instance to get the `txHex` as needed above.
3. update the `supportedTranscations` in keystone wallet instance to mark it as already supported `Versioned Transaction`.

### DIFF
--- a/packages/wallets/keystone/src/adapter.ts
+++ b/packages/wallets/keystone/src/adapter.ts
@@ -2,6 +2,7 @@ import type { DefaultKeyring } from '@keystonehq/sol-keyring';
 import type { WalletName } from '@solana/wallet-adapter-base';
 import {
     BaseMessageSignerWalletAdapter,
+    isVersionedTransaction,
     WalletAccountError,
     WalletLoadError,
     WalletNotConnectedError,
@@ -10,7 +11,7 @@ import {
     WalletReadyState,
     WalletSignTransactionError,
 } from '@solana/wallet-adapter-base';
-import type { Transaction } from '@solana/web3.js';
+import type { Transaction, TransactionVersion, VersionedTransaction } from '@solana/web3.js';
 import { PublicKey } from '@solana/web3.js';
 
 export interface KeystoneWalletAdapterConfig {}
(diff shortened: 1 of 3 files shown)


packages/wallets/keystone/src/adapter.ts
```diff
@@ -102,14 +103,23 @@ export class KeystoneWalletAdapter extends BaseMessageSignerWalletAdapter {
         this.emit('disconnect');
     }
 
-    async signTransaction<T extends Transaction>(transaction: T): Promise<T> {
+    async signTransaction<T extends Transaction | VersionedTransaction>(transaction: T): Promise<T> {
         try {
             const keyring = this._keyring;
             const publicKey = this._publicKey?.toString();
```
jordaaash, line 109:
```suggestion
            const publicKey = this._publicKey;
```
Can you change this line? We no longer need to stringify it.

packages/wallets/keystone/src/adapter.ts
```diff
@@ -102,14 +103,23 @@ export class KeystoneWalletAdapter extends BaseMessageSignerWalletAdapter {
...
             try {
-                return (await keyring.signTransaction(publicKey, transaction)) as T;
+                if (isVersionedTransaction(transaction)) {
+                    const txHex = transaction.serialize();
+                    const signature = await keyring.signTransaction(publicKey, txHex);
+                    transaction.addSignature(this._publicKey as PublicKey, signature);
+                } else {
+                    const txHex = transaction.serializeMessage();
+                    const signature = await keyring.signTransaction(publicKey, txHex);
+                    transaction.addSignature(this._publicKey as PublicKey, Buffer.from(signature));
```
jordaaash, line 120: With the above change made, we can just pass our local `publicKey` and not use `this._publicKey` or cast it.

jordaaash (changes requested):
Thanks!

jordaaash:
We're also going to need a changeset added for this to release it.

soralit:
Thanks for reviewing, will have a change later. About the changeset can you tell me how to add it? Or maybe you could raise a PR in our repo to this branch to add it.

packages/wallets/keystone/src/adapter.ts
soralit: the `publicKey` here is a parameter for keystone keyring and it's type is string so I converted it.

jordaaash:

PR #1004, opened 2024-07-22
update trust wallet icon

trustweb3team001:
update trust wallet icon


jordaaash:

PR #1006, opened 2024-08-04
Add Exodus Wallet

exodusteam001:


jordaaash:
The proposed adapter doesn't add any functionality that can't be found in the Standard Wallet implementation Exodus registers on the window. Exodus supports the Wallet Standard, so there's no need to use an adapter.

PR #1015, opened 2024-08-26
(typing) Allow readonly array of wallets for WalletProvider

dearlordylord:
In mine and some other codebases, immutability goes by default. 

The adapter doesn't like it, so the type like `const wallets: readonly [PhantomWalletAdapter, SolflareWalletAdapter]` won't work:

```ts
const useWallets = () => {
  const phantomWallet = new PhantomWalletAdapter();
  const solflareWallet = new SolflareWalletAdapter();
  return useMemo(() => [phantomWallet, solflareWallet] as const, [phantomWallet, solflareWallet]);
}

...

const wallets = useWallets();
<WalletProvider wallets={wallets}/>

```

I have to either cast or re-construct `<WalletProvider wallets={[...wallets]}/>`, which is undesirable.

This PR allows `readonly` arrays as `wallets` prop, as well as keeps mutable arrays backwards-compatible.


packages/core/react/src/WalletProvider.tsx
```diff
@@ -49,7 +49,7 @@ export function WalletProvider({
     onError,
 }: WalletProviderProps) {
     const { connection } = useConnection();
-    const adaptersWithStandardAdapters = useStandardWalletAdapters(adapters);
+    const adaptersWithStandardAdapters = useStandardWalletAdapters(adapters as Adapter[]);
```
dearlordylord, line 52: would require same improvement for useStandardWalletAdapters of @solana/wallet-standard-wallet-adapter-react

PR #1016, opened 2024-08-27
Coinbase Wallet universal links

ssamkkim:
This pr adds universal links to the Coinbase Wallet adapter, allowing IOS browser users to be redirected to their Coinbase Wallet app if it is installed. The code borrows semantics and logic from other wallet adapters that implement universal links such as Phantom and Solflare. 

https://github.com/user-attachments/assets/b33ea92c-ed89-432f-ba74-2e5526e959b9

### DIFF
--- a/packages/wallets/coinbase/src/adapter.ts
+++ b/packages/wallets/coinbase/src/adapter.ts
@@ -1,6 +1,7 @@
 import type { EventEmitter, SendTransactionOptions, WalletName } from '@solana/wallet-adapter-base';
 import {
     BaseMessageSignerWalletAdapter,
+    isIosAndRedirectable,
     isVersionedTransaction,
     scopePollingDetectionStrategy,
     WalletAccountError,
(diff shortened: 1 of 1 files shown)


mcintyre94 (commented):
Thanks for this! I haven't tested but the code/logic looks good, and makes sense to add this for Coinbase Wallet.

Could you point me to somewhere in their docs that shows this is the right way to redirect into their app? Just keen to make sure we're using something they properly support here.

ssamkkim:
> Thanks for this! I haven't tested but the code/logic looks good, and makes sense to add this for Coinbase Wallet.
> 
> 
> 
> Could you point me to somewhere in their docs that shows this is the right way to redirect into their app? Just keen to make sure we're using something they properly support here. 

Yep, here's the [link](https://docs.cdp.coinbase.com/wallet-sdk/docs/deep-link/) Coinbase Wallet's deep link docs.

It may also be helpful to use the IOS specific redirect URL instead. This video is what that would look like to use that instead of the universal link on the Arc Search IOS mobile browser. 

https://github.com/user-attachments/assets/17d65b99-ad86-4f1e-a27e-5868a729be52.

PR #1021, opened 2024-09-05
Fix Tests since MWA is no longer autoselected 

alex-fung:


packages/core/react/src/__tests__/WalletProviderMobile-test.tsx
```diff
@@ -470,8 +474,8 @@ describe('WalletProvider when the environment is `MOBILE_WEB`', () => {
                             mobileWalletAdapter.disconnect();
                         });
                     });
-                    it('should not clear the stored wallet name', () => {
-                        expect(localStorage.removeItem).not.toHaveBeenCalled();
+                    it('should clear the stored wallet name', () => {
+                        expect(localStorage.removeItem).toHaveBeenCalled();
```
jordaaash, line 478: Hmm. I think the original intent of the test is correct -- the wallet disconnecting should NOT deselect the wallet or remove it from LocalStorage, because then autoconnect will be broken on reload.

packages/core/react/src/__tests__/WalletProviderMobile-test.tsx
```diff
@@ -232,9 +232,9 @@ describe('WalletProvider when the environment is `MOBILE_WEB`', () => {
         beforeEach(() => {
             localStorage.removeItem(WALLET_NAME_CACHE_KEY);
         });
-        it('loads the mobile wallet adapter into state as the default', () => {
+        it('does not load the mobile wallet adapter into state as the default', () => {
             renderTest({});
-            expect(ref.current?.getWalletContextState().wallet?.adapter.name).toBe(SolanaMobileWalletAdapterWalletName);
+            expect(ref.current?.getWalletContextState().wallet?.adapter.name).toBe(undefined);
```
jordaaash, line 237: I think instead tests need to be changed/added such that

- If MWA (custom or default) is the only adapter, it should be selected by default.
- If it's not the only adapter, no adapter should be selected by default.

I think this satisfies the intent of the new code this is targeting.

packages/core/react/src/__tests__/WalletProviderMobile-test.tsx
```diff
@@ -326,10 +326,14 @@ describe('WalletProvider when the environment is `MOBILE_WEB`', () => {
...
         let errorThrown: WalletError;
-        beforeEach(() => {
+        beforeEach(async () => {
             errorThrown = new WalletError('o no');
             onError = jest.fn();
             renderTest({ onError });
+            await act(async () => {
+                ref.current?.getWalletContextState().select(SolanaMobileWalletAdapterWalletName);
+                await Promise.resolve(); // Flush all promises in effects after calling `select()`.
+            });
```
jordaaash, line 336: This change doesn't make sense to me. MWA shouldn't be deselected after each test, given the comment following this one)

jordaaash (changes requested):
I can see how these changes "fix" the tests, but they do so in a way that breaks the intent of what we wanted to test in the first place, and also don't test the new behavior that we want to have with the changes this targets.

packages/core/react/src/__tests__/WalletProviderMobile-test.tsx
alex-fung: is this true? AFAICT when using wallet adapter autoconnect is not supposed to reload the last used wallet if disconnect was called

packages/core/react/src/__tests__/WalletProviderMobile-test.tsx
alex-fung: hey sorry, not sure I totally follow – I don't think there's any selecting involved, the test is just to make sure that when the adapter emits an error, it expects `onError` to be called. or was the point of the test something different?

jordaaash:
This should be addressed in https://github.com/anza-xyz/wallet-adapter/pull/1049 and released in https://github.com/anza-xyz/wallet-adapter/pull/1011.

PR #1037, opened 2024-11-21
Add iframe wallet adapter

callmewhy:
Add wallet adapter for iframe

In the case of dapp nested in an iframe, this adapter can be used to access the parent web wallet within the iframe
without being restricted by browser security policies.


octavio12345300 (commented):
ok

PR #1049, opened 2025-02-06
Fix: MWA Select but No Connect Bug

Funkatronics:
This PR fixes an issue with MWA where the Solana Mobile wallet adapter gets stuck in a state that prevents the user from connecting to thier installed wallet. The issue is fully described with root cause and a video on our repo [here](https://github.com/solana-mobile/mobile-wallet-adapter/issues/1086). 

This fix is nearly identical to the changes in #1013. All tests have been fixed to accommodate the new behavior. The #1013 PR should be closed in favor of this PR. Alternatively I can rebase my changes on that branch if you prefer @jordaaash.

### DIFF
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -102,8 +102,6 @@ export function WalletProvider({
         if (!adapter) return;
         function handleDisconnect() {
             if (isUnloadingRef.current) return;
-            // Leave the adapter selected in the event of a disconnection.
-            if (walletName === SolanaMobileWalletAdapterWalletName && getIsMobile(adaptersWithStandardAdapters)) return;
             setWalletName(null);
         }
         adapter.on('disconnect', handleDisconnect);
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -76,10 +76,7 @@ export function WalletProvider({
         }
         return [mobileWalletAdapter, ...adaptersWithStandardAdapters];
     }, [adaptersWithStandardAdapters, mobileWalletAdapter]);
-    const [walletName, setWalletName] = useLocalStorage<WalletName | null>(
-        localStorageKey,
-        getIsMobile(adaptersWithStandardAdapters) ? SolanaMobileWalletAdapterWalletName : null
-    );
+    const [walletName, setWalletName] = useLocalStorage<WalletName | null>(localStorageKey, null);
     const adapter = useMemo(
         () => adaptersWithMobileWalletAdapter.find((a) => a.name === walletName) ?? null,
         [adaptersWithMobileWalletAdapter, walletName]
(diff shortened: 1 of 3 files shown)


packages/core/react/src/WalletProvider.tsx
```diff
@@ -76,10 +76,7 @@ export function WalletProvider({
...
         return [mobileWalletAdapter, ...adaptersWithStandardAdapters];
     }, [adaptersWithStandardAdapters, mobileWalletAdapter]);
-    const [walletName, setWalletName] = useLocalStorage<WalletName | null>(
-        localStorageKey,
-        getIsMobile(adaptersWithStandardAdapters) ? SolanaMobileWalletAdapterWalletName : null
-    );
+    const [walletName, setWalletName] = useLocalStorage<WalletName | null>(localStorageKey, null);
     const adapter = useMemo(
         () => adaptersWithMobileWalletAdapter.find((a) => a.name === walletName) ?? null,
         [adaptersWithMobileWalletAdapter, walletName]
```
mcintyre94, line 82: This is different to #960 and #1013 which change this to:

```ts
() =>
            adaptersWithMobileWalletAdapter.find((a) => a.name === walletName) ??
            (adaptersWithMobileWalletAdapter.length === 1 && adaptersWithMobileWalletAdapter[0] === mobileWalletAdapter
                ? mobileWalletAdapter
                : null),
        [adaptersWithMobileWalletAdapter, walletName, mobileWalletAdapter]
```

Do we not need that change any more?

It looks like that change was motivated by this comment: https://github.com/anza-xyz/wallet-adapter/pull/960#discussion_r1625162442

mcintyre94 (commented):
This looks good, thanks! Just one clarifying question about a change in the other PRs that isn't in this one.

packages/core/react/src/WalletProvider.tsx
Funkatronics: Correct, we do not want this change. If it was added I would propose a new PR to remove it as it would bring back the bug we are trying to fix here 😅

This change was added originally to try to preserve the existing behavior with MWA in the (rare) case that MWA is the one and only adapter available on Mobile. But we are aiming to remove the existing behavior, since it causes the bug with react state leading to MWA being non-functional for mobile users.

Funkatronics: Its an edge case that will likely never happen anyways, MWA being the one and only adapter available. So between that and the fact that it can reintroduce the bug we are trying to fix, we think its better to remove this special treatment for MWA. 

If at some point we decide to somehow change the react state handling in the UI layer to alleviate this issue caused by default selection of the adapter, we could then reintroduce special handling on MWA.

See [this comment](https://github.com/anza-xyz/wallet-adapter/pull/960#discussion_r1598992803) from, Jordan on the original pr.

mcintyre94: Got it, thanks! Glad we're finally getting this fixed!

PR #1073, opened 2025-04-08
feat: use WalletConnect provided package for `walletconnect` adapter

tomiir:
## Summary
- Replaces custom solution on `walletconnect` adapter with WC provided adapter.
- Includes `packageExports` directive to allow Parcel to parse path exports. [Source](https://parceljs.org/blog/v2-9-0/#new-resolver)


packages/wallets/walletconnect/src/index.ts
```diff
@@ -1 +1 @@
-export * from './adapter.js';
```
arein, line 1: is this breaking anybody?

tomiir: Our package is a replacement of that file, has the same exports.

packages/wallets/walletconnect/package.json
```diff
@@ -35,7 +35,7 @@
         "@solana/web3.js": "^1.77.3"
     },
     "dependencies": {
-        "@jnwng/walletconnect-solana": "^0.2.0",
+        "@walletconnect/solana-adapter": "0.0.7-b89b86f11cf15e97908341e7d27e61f1b358f7af.0",
```
arein, line 38:
```suggestion
        "@walletconnect/solana-adapter": "^0.0.7",
```

package.json
```diff
@@ -66,5 +67,8 @@
         "@types/web": "npm:typescript@~4.7.4",
         "eslint": "8.22.0",
         "@ngraveio/bc-ur": "1.1.12"
+    },
+    "@parcel/resolver-default": {
+        "packageExports": true
```
steveluscher, line 72: Can you explain what this does?

arein: CC @tomiir

tomiir: Hi @steveluscher ! My bad for not including it in the description (editing right now). 

Seems like Parcel v2.9+ requires this to be set on the root package json in order to properly parse path exports from other packages. 

Here's the related blog post from the parcel team: https://parceljs.org/blog/v2-9-0/#new-resolver

steveluscher (approved):
computer-office

We will release this, then mark it as deprecated with the following message:

```
Please use https://www.npmjs.com/package/@walletconnect/solana-adapter instead
```

PR #1047, opened 2025-01-13
feat: add ledger signMessage

ByteZhang1024:
Support signMessage for Ledger.

I have a question: would it be better to add a signOffchainMessage method in the Standard Wallet?
Or should we make signMessage compatible instead?

Issue #560, opened 2022-09-13
Remove generic `Name` parameter of WalletAdapterProps


jordansexton:
https://github.com/solana-labs/wallet-adapter/pull/558#discussion_r969033486

> Speaking of which, it's probably time for the generic `Name` to go! It doesn't actually do anything useful (because you don't have type safety for which wallets are available inside of your React context, which is how this is mostly used).

PR #1097, opened 2025-05-30
[WIP] 1.0.0

jordaaash:
We are preparing the release of Wallet Adapter v1.0.0

This will be the first major breaking change of Wallet Adapter in a long time, and may require small code changes in applications.

- It changes the API of `signMessage` to support the Solana Off-Chain Message Signing Format specification, allowing signing with Ledger devices (#1094 / #1047)
- It removes Mobile Wallet Adapter (#1107)

There are some things TBD (feel free to suggest others)

- We _may_ want to include `signAndSendAllTransactions` (#841 / #889)
- We probably need to include some utilities for creating and verifying OCMSF messages. We have some of this in the Solana extensions to the Wallet Standard now, but it makes sense to finish that code, and reuse and reexport it here.
- We should remove the `WalletName` generic (#560)

Release checklist (will add more steps/detail as we go)
- [ ] Identify and merge any other _non-breaking_ PRs to `master`
- [ ] Merge the (hopefully last) changeset Release PR for 0.x.x
- [ ] Identify and merge any other _breaking_ PRs into _this_ PR
- [ ] Update the demo and docs
- [ ] Add a changeset with major bump to all packages
- [ ] Merge this PR to `master`
- [ ] Merge the changeset Release PR for 1.0.0
- [ ] Update `@solana/wallet-standard-wallet-adapter-base` to be 1.0.0-compatible (will require a breaking change of that package)
- [ ] Update `@solana-mobile/wallet-adapter-mobile` to be 1.0.0-compatible (will require a breaking change of that package)

### DIFF
--- a/packages/core/base/src/signer.ts
+++ b/packages/core/base/src/signer.ts
@@ -1,4 +1,4 @@
-import type { SolanaSignInInput, SolanaSignInOutput } from '@solana/wallet-standard-features';
+import type { SolanaSignInInput, SolanaSignInOutput, SolanaSignMessageOutput } from '@solana/wallet-standard-features';
 import type { Connection, TransactionSignature } from '@solana/web3.js';
 import {
     BaseWalletAdapter,
@@ -116,7 +116,7 @@ export abstract class BaseSignerWalletAdapter<Name extends string = string>
 }
 
 export interface MessageSignerWalletAdapterProps<Name extends string = string> extends WalletAdapterProps<Name> {
-    signMessage(message: Uint8Array): Promise<Uint8Array>;
+    signMessage(message: Uint8Array): Promise<SolanaSignMessageOutput>;
 }
 
 export type MessageSignerWalletAdapter<Name extends string = string> = WalletAdapter<Name> &
@@ -126,7 +126,7 @@ export abstract class BaseMessageSignerWalletAdapter<Name extends string = strin
     extends BaseSignerWalletAdapter<Name>
     implements MessageSignerWalletAdapter<Name>
 {
-    abstract signMessage(message: Uint8Array): Promise<Uint8Array>;
+    abstract signMessage(message: Uint8Array): Promise<SolanaSignMessageOutput>;
 }
 
 export interface SignInMessageSignerWalletAdapterProps<Name extends string = string> extends WalletAdapterProps<Name> {
--- a/packages/core/base/src/signer.ts
+++ b/packages/core/base/src/signer.ts
@@ -115,8 +115,10 @@ export abstract class BaseSignerWalletAdapter<Name extends string = string>
     }
 }
 
+export type SignMessageOutput = SolanaSignMessageOutput;
+
 export interface MessageSignerWalletAdapterProps<Name extends string = string> extends WalletAdapterProps<Name> {
-    signMessage(message: Uint8Array): Promise<SolanaSignMessageOutput>;
+    signMessage(message: Uint8Array): Promise<SignMessageOutput>;
 }
 
 export type MessageSignerWalletAdapter<Name extends string = string> = WalletAdapter<Name> &
@@ -126,11 +128,14 @@ export abstract class BaseMessageSignerWalletAdapter<Name extends string = strin
     extends BaseSignerWalletAdapter<Name>
     implements MessageSignerWalletAdapter<Name>
 {
-    abstract signMessage(message: Uint8Array): Promise<SolanaSignMessageOutput>;
+    abstract signMessage(message: Uint8Array): Promise<SignMessageOutput>;
 }
 
+export type SignInInput = SolanaSignInInput;
+export type SignInOutput = SolanaSignInOutput;
+
 export interface SignInMessageSignerWalletAdapterProps<Name extends string = string> extends WalletAdapterProps<Name> {
-    signIn(input?: SolanaSignInInput): Promise<SolanaSignInOutput>;
+    signIn(input?: SignInInput): Promise<SignInOutput>;
 }
 
 export type SignInMessageSignerWalletAdapter<Name extends string = string> = WalletAdapter<Name> &
@@ -140,5 +145,5 @@ export abstract class BaseSignInMessageSignerWalletAdapter<Name extends string =
     extends BaseMessageSignerWalletAdapter<Name>
     implements SignInMessageSignerWalletAdapter<Name>
 {
-    abstract signIn(input?: SolanaSignInInput): Promise<SolanaSignInOutput>;
+    abstract signIn(input?: SignInInput): Promise<SignInOutput>;
 }
--- a/packages/core/react/package.json
+++ b/packages/core/react/package.json
@@ -37,7 +37,6 @@
         "react": "*"
     },
     "dependencies": {
-        "@solana-mobile/wallet-adapter-mobile": "^2.2.0",
         "@solana/wallet-adapter-base": "workspace:^",
         "@solana/wallet-standard-wallet-adapter-react": "^1.1.4"
     },
--- a/packages/core/react/src/WalletProvider.tsx
+++ b/packages/core/react/src/WalletProvider.tsx
@@ -1,16 +1,7 @@
-import {
-    createDefaultAddressSelector,
-    createDefaultAuthorizationResultCache,
-    createDefaultWalletNotFoundHandler,
-    SolanaMobileWalletAdapter,
-    SolanaMobileWalletAdapterWalletName,
-} from '@solana-mobile/wallet-adapter-mobile';
 import { type Adapter, type WalletError, type WalletName } from '@solana/wallet-adapter-base';
 import { useStandardWalletAdapters } from '@solana/wallet-standard-wallet-adapter-react';
 import React, { type ReactNode, useCallback, useEffect, useMemo, useRef } from 'react';
 import getEnvironment, { Environment } from './getEnvironment.js';
-import getInferredClusterFromEndpoint from './getInferredClusterFromEndpoint.js';
-import { useConnection } from './useConnection.js';
 import { useLocalStorage } from './useLocalStorage.js';
 import { WalletProviderBase } from './WalletProviderBase.js';
 
@@ -35,12 +26,6 @@ function getIsMobile(adapters: Adapter[]) {
     return getEnvironment({ adapters, userAgentString }) === Environment.MOBILE_WEB;
 }
 
-function getUriForAppIdentity() {
-    const location = globalThis.location;
-    if (!location) return;
-    return `${location.protocol}//${location.host}`;
-}
-
 export function WalletProvider({
     children,
     wallets: adapters,
@@ -48,38 +33,11 @@ export function WalletProvider({
     localStorageKey = 'walletName',
     onError,
 }: WalletProviderProps) {
-    const { connection } = useConnection();
     const adaptersWithStandardAdapters = useStandardWalletAdapters(adapters);
-    const mobileWalletAdapter = useMemo(() => {
-        if (!getIsMobile(adaptersWithStandardAdapters)) {
-            return null;
-        }
-        const existingMobileWalletAdapter = adaptersWithStandardAdapters.find(
-            (adapter) => adapter.name === SolanaMobileWalletAdapterWalletName
-        );
-        if (existingMobileWalletAdapter) {
-            return existingMobileWalletAdapter;
-        }
-        return new SolanaMobileWalletAdapter({
-            addressSelector: createDefaultAddressSelector(),
-            appIdentity: {
-                uri: getUriForAppIdentity(),
-            },
-            authorizationResultCache: createDefaultAuthorizationResultCache(),
-            cluster: getInferredClusterFromEndpoint(connection?.rpcEndpoint),
-            onWalletNotFound: createDefaultWalletNotFoundHandler(),
-        });
-    }, [adaptersWithStandardAdapters, connection?.rpcEndpoint]);
-    const adaptersWithMobileWalletAdapter = useMemo(() => {
-        if (mobileWalletAdapter == null || adaptersWithStandardAdapters.indexOf(mobileWalletAdapter) !== -1) {
-            return adaptersWithStandardAdapters;
-        }
-        return [mobileWalletAdapter, ...adaptersWithStandardAdapters];
-    }, [adaptersWithStandardAdapters, mobileWalletAdapter]);
     const [walletName, setWalletName] = useLocalStorage<WalletName | null>(localStorageKey, null);
     const adapter = useMemo(
-        () => adaptersWithMobileWalletAdapter.find((a) => a.name === walletName) ?? null,
-        [adaptersWithMobileWalletAdapter, walletName]
+        () => adaptersWithStandardAdapters.find((a) => a.name === walletName) ?? null,
+        [adaptersWithStandardAdapters, walletName]
     );
     const changeWallet = useCallback(
         (nextWalletName: WalletName<string> | null) => {
@@ -90,7 +48,7 @@ export function WalletProvider({
                 // sufficient reason to call `disconnect` on the mobile wallet adapter.
                 // Calling `disconnect` on the mobile wallet adapter causes the entire
                 // authorization store to be wiped.
-                adapter.name !== SolanaMobileWalletAdapterWalletName
+                adapter.name !== 'Solana Mobile Wallet'
             ) {
                 adapter.disconnect();
             }
@@ -125,7 +83,7 @@ export function WalletProvider({
     }, [autoConnect, adapter]);
     const isUnloadingRef = useRef(false);
     useEffect(() => {
-        if (walletName === SolanaMobileWalletAdapterWalletName && getIsMobile(adaptersWithStandardAdapters)) {
+        if (walletName === 'Solana Mobile Wallet' && getIsMobile(adaptersWithStandardAdapters)) {
             isUnloadingRef.current = false;
             return;
         }
@@ -159,7 +117,7 @@ export function WalletProvider({
     );
     return (
         <WalletProviderBase
-            wallets={adaptersWithMobileWalletAdapter}
+            wallets={adaptersWithStandardAdapters}
             adapter={adapter}
             isUnloadingRef={isUnloadingRef}
             onAutoConnectRequest={handleAutoConnectRequest}
--- a/packages/core/react/src/__tests__/WalletProviderBase-test.tsx
+++ b/packages/core/react/src/__tests__/WalletProviderBase-test.tsx
@@ -281,8 +281,8 @@ describe('WalletProviderBase', () => {
             expect(onError).not.toBeCalled();
         });
         describe('when a wallet is connected', () => {
-            beforeEach(async () => {
-                await act(() => {
+            beforeEach(() => {
+                act(() => {
                     ref.current?.getWalletContextState().connect();
                 });
                 expect(ref.current?.getWalletContextState()).toMatchObject({
@@ -330,7 +330,7 @@ describe('WalletProviderBase', () => {
                 fooWalletAdapter.connectionPromise = new Promise<void>((resolve) => {
                     commitConnection = resolve;
                 });
-                await act(() => {
+                act(() => {
                     ref.current?.getWalletContextState().connect();
                 });
             });
@@ -344,8 +344,8 @@ describe('WalletProviderBase', () => {
                 });
             });
             describe('once connected', () => {
-                beforeEach(async () => {
-                    await act(() => {
+                beforeEach(() => {
+                    act(() => {
                         commitConnection();
                     });
                 });
@@ -364,13 +364,13 @@ describe('WalletProviderBase', () => {
             beforeEach(async () => {
                 window.open = jest.fn();
                 renderTest({ adapter: fooWalletAdapter });
-                await act(() => {
+                act(() => {
                     ref.current?.getWalletContextState().connect();
                 });
                 fooWalletAdapter.disconnectionPromise = new Promise<void>((resolve) => {
                     commitDisconnection = resolve;
                 });
-                await act(() => {
+                act(() => {
                     ref.current?.getWalletContextState().disconnect();
                 });
             });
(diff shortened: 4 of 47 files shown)


cemozerr:
> I have a question—after making this change, it means that signMessage now has two different implementations (Phantom's implementation and Ledger's implementation). When a dApp verifies the result of signMessage, does it need to support both for compatibility?

Ideally every wallet should move to the Offchain Message Signing Spec @ByteZhang1024.

ByteZhang1024:
I originally envisioned two possible approaches:

All wallet implementations of signMessage should move to the Offchain Message Signing Spec going forward.

Introduce a new method to help wallets and dApps transition gradually, with the eventual deprecation of signMessage.

That's why I submitted another PR, but it looks like we've decided to go with option one. So that PR can probably be closed: https://github.com/anza-xyz/wallet-standard/issues/81

jordaaash:
Important to note that the Ledger implementation in this PR is using `0{32}` for the [application domain](https://docs.anza.xyz/proposals/off-chain-message-signing#application-domain) in the spec.

> **Application domain**
> 
> A 32-byte array identifying the application requesting off-chain message signing. This may be any arbitrary bytes. For instance the on-chain address of a program, DAO instance, Candy Machine, etc.
>
> This field SHOULD be displayed to users as a base58-encoded ASCII string rather than interpreted otherwise.

This is a poorly designed part of the spec, IMO, and I disagree with all of it. It's an off-chain message. Referencing an on-chain thing is unnecessary and antithetical. The spec also recommends that wallets display this value to users. Displaying a base58 pubkey of a Solana program that the user certainly will not recognize is useless and bad.

My recommendation to implementing wallets and apps:
- **DO** use `0{32}` for this value
- **DO NOT** display this value in base58 or otherwise, especially not `1111...1111`, to users
- **DO** use [Sign In With Solana](https://github.com/phantom/sign-in-with-solana) for the message body, which includes a human-readable domain and the user's pubkey, which will be recognizable to them, since both of these fields serve an actual purpose

vsakos:
I might be late to the discussion but IMO wallets will not and should not convert the dApp's message to an OCMSF message (at least for foreseeable future), because as soon as they do, any dApp that does not verify the `signedMessage` but only the original message and the received `signature`, will break instantly.

Until now, the only way to implement this was by implementing the Wallet Standard directly, and I don't think a lot of dApps do that. They might start doing it (and they should) with the release of the adapter v1.0.0 but it will take time until everybody upgrades since this is a breaking change and not just an `npm install`.

From the adapter's perspective, this is a breaking change that won't affect dApps until they upgrade the library, but as soon as wallets change how the `window.<wallet>.signMessage()` works, dApps will break. The same applies the Wallet Standard since the standard wallets replace the equivalent wallet adapter that's manually added by the dApp.

cemozerr:
@vsakos are you saying that apps should special case when the user tries to sign with a Ledger-imported account? (Otherwise, Ledger will reject the signature request due to incorrect formatting—it only accepts OCMSF messages now.)

If that's not the case, the ecosystem won't have Ledger support for off-chain signatures for quite some time.

Another question I'm wondering about—what major apps are currently using the message signing flow? My understanding is that not many teams have adopted it yet, since doing so currently excludes Ledger users, who represent a significant group.

jordaaash:
@vsakos Not too late, and it's good to highlight this problem.

Apps **AND** wallets, whether using Wallet Adapter or the Wallet Standard, are going to have to make changes to handle Ledger-signed messages, no matter what. To illustrate this, here are the set of possible outcomes I see:

0. Status quo: a given app and wallet both make no changes to support OCMSF. Setting aside the transaction signing tricks they are doing today, the app asks the wallet to sign a non-OCMSF message.
  a. It's a non-Ledger account. Signing in the wallet and verification in the app work.
  b. It's a Ledger account. Signing fails, with an opaque error in the wallet, app, or both, regardless of the version of the Solana Ledger installed.
1. A given app makes changes to supports OCMSF for all messages, and a given wallet makes no changes. The app asks to sign an OCMSF message. It uses the restricted ASCII format to get non-blind-signing display on Ledger, and it doesn't know if it's a Ledger account or not.
  a. If it's a non-Ledger account, the wallet decodes the message as utf-8, likely displays a lot of garbage from the OCMSF envelope. Signing and verification _maybe_ work, but it depends on the implementation of the wallet's parsing of the message.
  b. If it's a Ledger account, signing fails, potentially with an opaque error in the wallet, app or both, because the wallet made no changes to support OCMSF.
2. A given wallet makes changes to support OCMSF for all messages, and a given app makes no changes. The app asks to sign a non-OCMSF message. The wallet converts the message to OCMSF, signs it, and returns `signature`, and possibly the `signedMessage`. Regardless, the app is unaware of the conversion or `signedMessage`, so verification of `signature` fails.
3. A given wallet makes changes to support OCMSF for Ledger only, and a given app makes no changes. The app asks to sign a non-OCMSF message.
  a. It's a non-Ledger account, so the wallet does not convert the message to OCMSF. Signing and verification work.
  b. It's a Ledger account. The wallet converts the message to OCMSF, signs it, and returns `signature`, and possibly the `signedMessage`. Signing succeeds, but verification fails.
4. A given app and wallet both make changes to support OCMSF. The app asks to sign either an OCMSF message or non-OCMSF message.
  a. If it's a non-Ledger account, the wallet either converts the message to OCMSF or not, and signs it. The app either receives the `signedMessage` or not. It can still verify against the message it provided, and if that fails, it can convert its own message to OCMSF and verify against that.
  b. If it's a Ledger account, the wallet must either receive an OCMSF from the app or convert the message the OCMSF to sign it. The app either receives the `signedMessage` or not. It can still verify against the message it provided, and if that fails, it can convert its own message to OCMSF and verify against that.

My takeaways from this:

In the near term, wallets **SHOULD NOT** convert messages to OCMSF unless they are signing with a Ledger account. This does leak information about the nature of the account if it's a Ledger account, but it means that most calls to `window.<wallet>.signMessage()` will continue to work as-is (because they won't be signing an OCMSF message). Wallets **SHOULD** convert to OCMSF if they are signing with a Ledger account, because signing will fail otherwise.

This implies that apps also **SHOULD NOT** convert messages to OCMSF, because they don't know if they are signing with a Ledger account or not, and they don't know if a given wallet supports OCMSF. Apps **SHOULD** expect that wallets may have converted the message to OCMSF, so if they don't get a `signedMessage` back, and apps should verify both ways. If they do get a `signedMessage` back, they should use that to verify, but they will still also need to parse the format of the `signedMessage` to make sure it's valid and that it contains the message they asked to sign. In practice this is almost the same (but not quite) as just converting their own message to OCMSF (the main difference is that the application domain may not be `0{32}`, though it should be).

vsakos:
For me it still makes sense to keep the wallet side as-is (sign the whole payload and return the signature), basically this:

> A given app makes changes to supports OCMSF for all messages, and a given wallet makes no changes. The app asks to sign an OCMSF message. It uses the restricted ASCII format to get non-blind-signing display on Ledger, and it doesn't know if it's a Ledger account or not.
a. If it's a non-Ledger account, the wallet decodes the message as utf-8, likely displays a lot of garbage from the OCMSF envelope. Signing and verification maybe work, but it depends on the implementation of the wallet's parsing of the message.
b. If it's a Ledger account, signing fails, potentially with an opaque error in the wallet, app or both, because the wallet made no changes to support OCMSF.

I do not agree that it will fail since the signing is (and should always be) handled in Uint8Array and UTF-8 is strictly for the UI. If the app sends a correct OCMSF payload (just the message, not the whole `count + sig[] + msg` envelope), the wallet will just sign it as a whole. Some changes are definitely needed because as you said it will show garbage if we don't handle the data correctly on the UI but the core signing functionality should still work even without that.

Apps would have to change their logic anyway if they want to support Ledger users (either by formatting to OCMSF before signing or sending the base message but then verifying if the returned OCMSF is allright), but if for some reason they do not implement any of this, the base signing would still work without Ledger support.

Now, my recommendation would be to keep `signMessage()` as a "sign anything without touching the data" but with the ability to show both UTF-8 and OCMSF correctly on the UI, and add a new method that accepts only the UTF-8 (maybe even as a string) and returns `{ signature, signedMessage }` but I know this idea was discussed before and was rejected.

jordaaash:
There is no guarantee or mechanism to enforce that a wallet will not modify the message before signing, just as with `signTransaction`. The proposed API change makes this explicit by achieving compatibility with what the Wallet Standard already specifies and wallets who implement it support, but it is a breaking change. Apps are going to have to make changes eventually anticipating changes in wallet behavior, regardless of whether they use this release or 0.x releases.

We may want to take this approach now with some of the other methods as well for 1.0.0 if it makes sense to: a longer-term goal is to get everyone using the Standard and phase Wallet Adapter out (e.g. @solana/kit, fka web3js 2.x, supports the Wallet Standard and will not support Wallet Adapter which is coupled to web3.js 1.x). For example, it's been requested to add `signAndSendAllTransactions`, and when we do so, we will use the Wallet Standard API for this. We may want to make this part of 1.x even though we don't absolutely need to (it's non-breaking) in order to encourage updating.

For `VersionedTransaction` support, we added a `supportedTransactionVersions` field to the adapter. It could make sense for wallets on the window (or elsewhere) to do something like this to indicate OCMSF support, and for us to plumb it through the adapter, though I don't think that's sufficient reason to have two different methods for message signing since this is breaking. Individual wallets could implement two versions on their window objects, but we generally wouldn't call them from Wallet Adapter because the Wallet Standard API already covers this.

jordaaash:
A related data point -- the Mobile Wallet Adapter protocol returns a payload of `signedMessage || signature`, which is plumbed through their [Wallet Standard implementation](https://github.com/solana-mobile/mobile-wallet-adapter/blob/a9d7176b6b03917e41d7bfad368d7312bf7310ee/js/packages/wallet-standard-mobile/src/wallet.ts#L470) in expectation that the wallet may change it.

ByteZhang1024:
We at the OneKey team offer both hardware and software wallets, so I’d like to share our perspective as well.

## About `signMessage`

Initially, we only implemented `signMessage` in our software wallet. It wasn’t implemented in the hardware wallet because allowing arbitrary message signing is inherently dangerous. We believe Ledger shares the same concern, which is likely why they didn’t support it either.

However, websites like gmgn became extremely popular, and we received a large volume of user feedback complaining that our hardware wallet couldn’t connect to gmgn and similar dApps. As a result, we reluctantly allowed `signMessage` for Solana paths `m/44'/501'/x'/0'`, with strict checks to ensure the message is not a transaction and added clear warnings that this is unsafe.

## About OCMSF

From a wallet developer’s perspective, our priority is stability and reliability.
Therefore, we would like to see `signMessage` gradually deprecated — or at least augmented with a flag indicating that the dApp intends to use OCMSF.

For example:

```js
// Gradually deprecate signMessage
window.signMessage(Buffer.from("010203", 'utf8'))

// Add a second parameter to explicitly indicate OCMSF usage
window.signMessage(Buffer.from("010203", 'utf8'), "OCMSF")

// (Alternatively) Introduce a new API — though we’ve already discussed and abandoned this idea
window.signOffChainMessage(Buffer.from("010203", 'utf8'))  // wallet handles OCMSF and returns a signed result
```

The benefit of this approach is that both the dApp and the wallet clearly know what the other expects.

## The current approach in our PR.

If a wallet starts returning two different signature formats, it still heavily depends on dApp-side changes. We’re concerned that the intent of this change isn’t obvious enough — dApps may not realize that we’ve started supporting a new signing format. If a dApp doesn’t actively support OCMSF, then returning an OCMSF signature alone won’t help.

With the current direction, wallet developers have little incentive to be early adopters of OCMSF — because once we ship support, it could immediately break compatibility with many existing dApps.

dApps would need to make changes like this:

```js
const request = "010203"
const result = await signMessage(Buffer.from(request, 'utf8'))

const isValidSignature = nacl.sign.detached.verify(
  Buffer.from(request, 'utf8'),
  result,
  publicKey.toBytes(),
);

if (isValidSignature) {
  // signMessage: true
  return true
} else {
  const offchainMessage = new OffchainMessage({
    message: Buffer.from(request, 'utf8'),
  });
  const isValidSignature = nacl.sign.detached.verify(
    offchainMessage.serialize(),
    result,
    publicKey.toBytes(),
  );
  if (isValidSignature) {
    // OffchainMessage: true
    return true;
  }
}

return false;
```

Lastly, we’d be happy to help with external promotion and outreach related to OCMSF adoption.
Both our hardware products and Ledger’s share the same goal: to encourage the Solana ecosystem to migrate to OCMSF.

jordaaash:
> From a wallet developer’s perspective, our priority is stability and reliability. Therefore, we would like to see `signMessage` gradually deprecated — or at least augmented with a flag indicating that the dApp intends to use OCMSF.
> 
> For example:
> 
> ```js
> // Gradually deprecate signMessage
> window.signMessage(Buffer.from("010203", 'utf8'))
> 
> // Add a second parameter to explicitly indicate OCMSF usage
> window.signMessage(Buffer.from("010203", 'utf8'), "OCMSF")
> 
> // (Alternatively) Introduce a new API — though we’ve already discussed and abandoned this idea
> window.signOffChainMessage(Buffer.from("010203", 'utf8'))  // wallet handles OCMSF and returns a signed result
> ```

@ByteZhang1024 I want to address this first before coming back to the rest of the ideas here.

I'm not sure I understand why it would help to have an optional `"OCMSF"` flag. If the message is OCMSF formatted, a supporting wallet will parse `"\xsolana offchain"` at the start of the message, indicating it's OCMSF. A non-supporting wallet won't know about the `"OCMSF" flag and will just parse the message as UTF-8.

Wallets can add a separate `signOffChainMessage` method if they wish to, nothing in the proposed approach would prevent that. Their Wallet Standard wrappers can parse the message and call it if needed. But I don't think Wallet Adapter needs to have two methods, because it's supposed to use the Wallet Standard wherever possible. And if not possible, it needs to know about the underlying wallet implementation, and can call that wallet's `signOffChainMessage` method if it exists since the adapter will know about this.

As a result, I'm still not sure why this would be necessary, because the reality is that apps who want to do Ledger signing are going to have to be liberal in what they sign **AND** liberal in what they verify, because they simply cannot know for sure what a given wallet implementation will do, or if it even supports Ledger devices at all. The most flexible thing for them to do is to NOT convert messages to OCMSF, see what the wallet does with them, and verify accordingly.

As an example for why I think this is valuable --

Right now, we just have Ledger devices supporting OCMSF, displaying via restricted ASCII, and (maybe) blind signing for UTF-8. But nothing precludes Ledger or a different wallet from displaying longer messages in UTF-8 in the future for various reasons (e.g., perhaps for an EIP-712 type of format). The app has no way of knowing what type of device the account is on before encoding the message, but the wallet can and usually would know this. This leaves the wallet in _generally_ the best position to do the encoding, unless the app knows specifically what the wallet supports **AND** wants to be restrictive about the encoding it accepts. I think restrictiveness is an anti-goal for apps right now -- they just want signing of very basic messages, or SIWS messages, to work minimally in the broadest set of places.

vsakos:
Wallets don't have the luxury to do breaking changes in external facing APIs. They have to be backwards compatible, otherwise apps will break and none of us wants that.

> If the message is OCMSF formatted, a supporting wallet will parse "\xsolana offchain" at the start of the message, indicating it's OCMSF.

The `OCMSF` flag would indicate that the app knows how to handle this format and will not break if the wallet converts/formats the message. But then it's almost equivalent to having a separate `signOffChainMessage` function, except it's two functions (the legacy and the offchain compatible) merged into one.

As for the `VersionedTransaction` and `supportedTransactionVersions` example, the `signTransaction` interface and expected behavior were deterministic and fully backwards compatible - the app expects the same output format it used for the request (so it's always `Transaction -> Transaction` or `VersionedTransaction -> VersionedTransaction`). With the proposed approach for messages, `signMessage` would not be deterministic across wallets, sometimes it would return only the signature, other times both the signature and the modified message with no clear interface that would distinguish between the two. A new argument or a new method would solve this.

If `signMessage` returned a closed object that contained both the message and the signature and had it's own verification method (like the transaction objects) it would be different because the interface would support message modification from day one.

Also, once wallets release the new version that changes how the Wallet Standard's `solana:signMessage` works (talking about behavior, not the JS interface), dapps still using the v0.x version of the adapter will have no way to access the `signedMessage` if the wallet modified it.

jordaaash:
None of us wants apps to break, but apps are **going to have to change** to sign OCMSF messages with Ledger. On its own, returning an OCMSF-formatted message from `signMessage` _when used with a Ledger account_ is not going to break apps, because the status quo is that this already doesn't work. The thing that will break apps is returning a different type _in Wallet Adapter_ which is why it's a breaking change with a new major version here. The underlying API that wallets use for this can vary, but our primary focus is to support the Wallet Standard, which already anticipates this. My initial inclination is to not consider that in scope for this change because, as you point out:

> Also, once wallets release the new version that changes how the Wallet Standard's `solana:signMessage` works (talking about behavior, not the JS interface), dapps still using the v0.x version of the adapter will have no way to access the `signedMessage` if the wallet modified it.

However, we have some potential other options here:

```ts
type Signature = Uint8Array; // Added here for clarity
type SignMessageOutput = { signedMessage: Uint8Array; signature: Signature; signatureType?: 'ed25519' };
type OffChainMessage = Uint8Array & { __type: 'OffChainMessage }`; // Nominal type
```
Assumption is the app has to call a canonical method that converts and returns `OffChainMessage` (better) or explicitly cast a `Uint8Array` as `OffChainMessage` (worse). 

1. Breaking change to `signMessage` method:
```ts
signMessage(message: Uint8Array): SignMessageOutput;
```
Consistent with Wallet Standard but breaking, as discussed previously. "Easy" for Wallet Adapter, but hard for non-Standard Wallets to determine what to do.

2. New method:
```ts
signMessage(message: Uint8Array): Signature;
signOffChainMessage(message: OffChainMessage): Signature;
```
I think it'll create more confusion to apps to add another method, and there's nothing to indicate that a given wallet supports OCMSF. Meaning, we could add this method to Wallet Adapter, but we don't have a great way to detect that it's callable unless we add something to the adapter interface _and_ to the Wallet Standard to say so (this is the parallel I was drawing to `supportedTransactionVersions`).

3. Non-breaking dual method signature with optional arg:
```ts
signMessage(message: Uint8Array): Signature;
signMessage(message: UInt8Array, appSupportsOffChainMessage: true): SignMessageOutput;

// or

signMessage<T extends boolean>(message: UInt8Array, appSupportsOffChainMessage?: T): T extends true ? SignMessageOutput : Signature;
```
It's ambiguous whether the app should ever format the message as OCMSF, or just indicate it supports verifying OCMSF it gets from the wallet. The app doesn't know what the wallet supports or what kind of account it is, and wallets may simply ignore this and convert the message anyway when it's a Ledger account rather than throwing. Suffers from the same problem as 2.

4. Non-breaking dual method signature with nominal type:
```ts
signMessage(message: Uint8Array): Signature;
signMessage(message: OffChainMessage): Signature;
```

App is responsible for the conversion to OCMSF, but app doesn't know if the wallet, or a given account, supports OCMSF. This isn't substantially different from the app just converting to OCMSF and calling `signMessage(message: Uint8Array): Signature;` as the result is the same. The wallet is _presumably_ going to sign the message as-is, but still nothing precludes the wallet from modifying the message. App still doesn't know whether the wallet supports OCMSF.

---

Two issues I see with all of these approaches:

1. Standard Wallets are likely going to format the message as OCMSF and returned `signedMessage` where applicable, since that was the intent of that API, and nothing we do in the Wallet Adapter changes this.

2. Unless we add something to the Wallet Standard to indicate that the wallet supports OCMSF (i.e. a new version of the `solana:signMessage` feature with a flag) and plumb that through Wallet Adapter, it will be ambiguous to apps.

ByteZhang1024:
Sorry for not handling messages over the past two weeks as I was on vacation.

I think the first point we need to align on is the future support issue for signMessage. I believe this method will definitely be deprecated due to security concerns.
OCMSF message is the subsequent replacement method.

dApp continues to pass the message to the wallet, the wallet does formatting and then signs it, returning the signature and signedMessage.
dApp only needs to change the method name, and if there's validation, the validation method needs to be updated.
If dApp doesn't modify, it will continue to maintain the existing signMessage logic and keep working.

This compatibility issue has similar transitions in both ETH and SUI ecosystems. As long as the SOL specification makes changes, major wallets will quickly follow up and support it.

ETH personal_sign
or
https://github.com/MystenLabs/ts-sdks/blob/main/packages/wallet-standard/CHANGELOG.md#060

ByteZhang1024:
@jordaaash 
So let’s just follow what other ecosystems are doing and simply add a new API directly. This proposal has been delayed for too long. Releasing it sooner will allow everyone to support it earlier — just waiting like this isn’t getting us anywhere.
```
signOffChainMessage(message: OffChainMessage): Signature
```

etodanik:
Hey guys, jumping in here because there are multiple wallets (i.e Jupiter Wallet), Supabase (where I contributed and maintain the ledger auth process)  and other ecosystem players that are completely stuck due to what starts to feel like bike-shedding here. 

Can we either get players from `Phantom` and a few other places on a single call to resolve this, or ship `signOffChainMessage`? 

There are entire chunks of the ecosystem stuck because of lack of progress on this here.

etodanik:
@jordaaash, @ByteZhang1024, @cemozerr, @vsakos
Is there willingness to stop async-ing and get everyone (including representatives from Phantom) into an online call and decide? I'm willing to chase everyone down IRL, be the assistant and facilitate such a call.

It's a bit embarrassing and bad for the ecosystem that it's taking more than a year to solve such a fundamental issue.

I got good progress with Jupiter Wallet people and a few others and we're all literally just sitting and waiting for this PR to be decided and to drop. 

Both doing `signOffChainMessage` and the other solution are good, what's worse is bike shedding and never coming to any decision, and leaving the ecosystem with terrible UX and inability to use hardware wallets for sign in.

ByteZhang1024:
@mcintyre94 Could you please help push this issue forward?

mcintyre94:
Thanks to everyone for helping push this forward. We now have a plan, available in this document: https://www.notion.so/solanafoundation/Solana-Off-Chain-Message-Signing-OCMS-374d36dad52d80bf981ddbcebfeca5f0

Instead of modifying `signMessage` we are going to add a new wallet feature for offchain message signing.

We are not currently planning to ship a major version 1.0.0 or any breaking changes to wallet-adapter 

I'm going to close this PR now

PR #1124, opened 2025-09-19
Fixed cannot add custom class names to WalletMultiButton  component.

wrux:
The `WalletMultiButton` component allows a className prop. But `BaseWalletConnectionButton` overrides className using `className="wallet-adapter-button-trigger"` which is preventing from overriding default styles.

The suggested change allows to override the default CSS class name which allows custom styling for the button:

```tsx
function Test() {
  return (
    <WalletMultiButton
      className='button button--something'
    />
  );
}
```

An example applying Tailwind classes for a button using CVA:

```tsx
function Test() {
  return (
    <WalletMultiButton
      className={buttonVariants({
        size: 'lg',
      })}
    />
  );
}
```


beeman:
Seems like a duplicate of https://github.com/anza-xyz/wallet-adapter/pull/908 and https://github.com/anza-xyz/wallet-adapter/pull/1122.

PR #1129, opened 2025-11-01
Create SECURITY.md for security policy

Murshid69:
`worktree.guessRemote`::
	If no branch is specified and neither `-b` nor `-B` nor
	`--detach` is used, then `git worktree add` defaults to
	creating a new branch from HEAD.  If `worktree.guessRemote` is
	set to true, `worktree add` tries to find a remote-tracking
	branch whose name uniquely matches the new branch name.  If
	such a branch exists, it is checked out and set as "upstream"
	for the new branch.  If no such match can be found, it falls
	back to creating a new branch from the current `HEAD`.

`worktree.useRelativePaths`::
	Link worktrees using relative paths (when "`true`") or absolute
	paths (when "`false`"). This is particularly useful for setups
	where the repository and worktrees may be moved between
	different locations or environments. Defaults to "`false`".
+
Note that setting `worktree.useRelativePaths` to "`true`" implies enabling the
`extensions.relativeWorktrees` config (see linkgit:git-config[1]),
thus making it incompatible with older versions of Git.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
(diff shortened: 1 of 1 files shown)


Murshid69 (commented):
Release

PR #1140, opened 2026-02-02
Add Binance Wallet adapter

clinyong:
Hello everyone. I'm from Binance Wallet team.

This PR adds a wallet adapter for Binance Wallet. The adapter implementation follows the same pattern as other adapters in this repository.

Let me know if you have any questions.

### DIFF
--- a/packages/wallets/binance/LICENSE
+++ b/packages/wallets/binance/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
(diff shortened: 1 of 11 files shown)


mcintyre94:
Thanks for this @clinyong but you no longer need to create an adapter for this kind of wallet and we are no longer accepting PRs for new ones. Please see https://github.com/anza-xyz/wallet-adapter/blob/master/WALLET.md

You can instead use Wallet Standard, which will allow you to inject your wallet from your extension without needing dapps to add a new adapter. Please see [this guide for wallets](https://github.com/anza-xyz/wallet-standard/blob/master/WALLET.md) to implement the Wallet Standard. Note that we haven't been accepting new adapters of this type for a few years now, and dapps are very unlikely to add new adapters because of this. By using Wallet Standard, your extension will automatically be available to users of all dapps on Solana.

